PR #9, opened 2020-02-16
add go.mod for esbuild in go 1.13

awkj:


NaLLiFFuNT:
@evanw Hi, your project is great! Could you add go mod support so that it would be easier contributing your project?

evanw:
This is a personal side project and I'm not looking for contributions at this time. Currently esbuild doesn't have any dependencies and I'd like to keep it that way if I can. I'm also not intending for the Go source code in this repo to be imported as a library into other projects. See https://github.com/evanw/esbuild/issues/13#issuecomment-587111778 for my rationale.

tooolbox:
@awkj I'm not sure how this helps.  I guess with that file in place, you can do `cd src/esbuild && go build -o esbuild ./main/*` ?  You can't `go get` it AFAICT.

PR #54, opened 2020-04-04
Document how to install esbuild locally

maltsev:
In our project, we keep all build tools installed locally and listed in `package.json`. I've found a solution on how to install `esbuild` OS specific package automatically that way (using `optionalDependencies`). I think it's good to document that solution in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,18 @@ npm install -g esbuild-wasm       # for all other platforms
 
 This adds a command called `esbuild`.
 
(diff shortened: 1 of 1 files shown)


evanw:
I just landed a cross-platform package called `esbuild` that should automatically download the appropriate binary when installed, so these instructions are no longer necessary. I've updated the documentation to describe how to do a local install using the new package.

PR #57, opened 2020-04-12
Convert to go modules and simplify package structure

jschaf:
There's two commits here. The first uses go modules instead of GOPATH. The second commit simplifies the package structure to use `pkg` for library code and `cmd` for binaries.

### Working
- `make esbuild` succeeds and rebuilds for changes in `pkg/*` and `cmd/*`.
- `make test` succeeds.
- `make update-version-go` succeeds and updates the verison.
- `make platform-all` succeeds.
- `make clean` succeeds.
- `make demo-three` succeeds.
- `make bench-three` succeeds.

### Not working
- `make test262` hangs.


jschaf:
Oh, I missed your intent from the other PRs (https://github.com/evanw/esbuild/pull/9) I'll close.

PR #88, opened 2020-05-06
Fix Windows Path Issue

### LINKED ISSUE #87: Error using transform() API
Hi,

I run into an issue using transform() API.

When I run this example code  :
```js
(async () => {
  const ts = `
    export const value: number = 1;
  `;

  const esbuild = require("esbuild");
  const service = await esbuild.startService();

  try {
    const { js } = await service.transform(ts, { loader: "ts" });
    console.log(js);
  } catch (err) {
    console.error(err);
  }

  service.stop();
})();
```

I have the following error :
```
Error: Transform failed with 1 error
    at failureErrorWithLog (...\node_modules\esbuild\lib\main.js:35:17)
    at Object.transform (...\node_modules\esbuild\lib\main.js:250:38)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async ...\es.js:10:20 {
  errors: [
    { text: 'Could not read from file: E:\\input.ts', location: null }
  ],
  warnings: []
}
```

If I create a ts file and call the CLI command `npx esbuild`, it works properly.

I'm on Windows 10, using Node v12.16.3.

bekalshenoy:
fix #87

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "estraverse": "5.1.0",
-    "esutils": "2.0.3"
+    "esutils": "2.0.3",
+    "slash": "^3.0.0",
   }
 }
--- a/scripts/js-api-tests.js
+++ b/scripts/js-api-tests.js
@@ -1,6 +1,7 @@
 const { installForTests } = require('./esbuild')
 const childProcess = require('child_process')
 const assert = require('assert')
+const slash = require('slash')
 const path = require('path')
 const util = require('util')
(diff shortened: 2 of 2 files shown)


evanw:
This PR only modifies a test file, so it can't be the right fix for #87. Test files are not included in the final npm package so this PR doesn't change anything in the final npm package.

evanw:
I believe this should be fixed by 562a849d526c213549282b830a224e68b5327c0e.

PR #91, opened 2020-05-07
Skip install script if already installed

yyx990803:
The install script sometimes throws error on subsequent `yarn` installs since `yarn` seems to invoke install scripts when it is shuffling things around. It also causes issue for CI where the `node_modules` directory is cached across builds.

### DIFF
--- a/npm/esbuild/install.js
+++ b/npm/esbuild/install.js
@@ -17,6 +17,10 @@ function installPackage(package) {
       env[key] = process.env[key];
     }
   }
+  
+  if (fs.existsSync(installDir)) {
+    return;
+  }
 
   // Run "npm install" recursively to install this specific package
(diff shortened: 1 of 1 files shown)


evanw:
Ah that's annoying. I use npm, not yarn, so I've only been testing with that. You don't happen to have a repro, do you?

I'm wary that this change won't fix the problem. Right after the call to `installPackage()`, the install script then does a `fs.renameSync()` of a file inside the installation directory to move it into place. I'm guessing that call will fail if it's already been moved. So I suspect that you'll also have to tell the caller to return early too. But it'd be great to have a repro to double-check that this is correct.

evanw:
Actually, since it sounds like the `postinstall` script just needs to be idempotent, this should be something I can test. I'll merge this and fix it up so it works.

PR #93, opened 2020-05-08
fix(esbuild): Fix minimum version error on old versions of npm

getspooky:
This can happen if you have a dependency which lists an old version of npm in its own dependencies.

### DIFF
--- a/npm/esbuild/index.js
+++ b/npm/esbuild/index.js
@@ -1,0 +1,17 @@
+'use strict';
+
+const { exec } = require('child_process');
+
+exec('npm -v', (err, stdout) => {
+  if (err) throw err;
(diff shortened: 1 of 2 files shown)


evanw:
Thank you for trying to fix this issue. However, I would like to understand what the issue is first and make sure I can reproduce it myself. I tried these commands but I couldn't reproduce the issue:

```
npm install npm@4
npm install esbuild
npx esbuild
```

Can you provide steps that would help me reproduce the issue myself?

If the issue that the older version of `npm` from the local `node_modules` folder somehow overrides the newer one that's installed globally, I think a better fix would be to use the `npm_execpath` environment variable to derive the location of npm so that it's not subject to path resolution. That way the install script would still work even if you end up in this scenario.

Issue #94, opened 2020-05-08
CI/CD Setup


itshaydendev:
Just thinking how this could benefit from a CI/CD solution. Like, running all your tests and stuff as part of every push.

Deploying to NPM on every new tag.

That sorta thing. If you wanted, as long as I knew what to run, I could throw together a GitHub Actions config file for you! :)

evanw:
Wow, that'd be great! I've wanted to do this, mainly for running tests on Linux, macOS, and Windows. I already did the hard work of getting the tests running on Windows but I haven't done the GitHub Actions thing yet because I haven't used it before. The commands I'd want to run on all three platforms are:

```
go test ./internal/...
npm ci
node scripts/verify-source-map.js
node scripts/end-to-end-tests.js
node scripts/js-api-tests.js
```

This is essentially just `make test`. I was thinking of not running these core tests with `make` so people don't have to install the make toolchain on Windows to run tests locally. Later on I'll probably move this to a `npm test` script but I figure this would be a good start.

itshaydendev:
Due to npm/cli#558, I'm gonna have to use `npm i` over `npm ci` but the implementation is basically sorted. https://github.com/itshaydendev/esbuild/actions

itshaydendev:
I'll make a PR once this runs

PR #100, opened 2020-05-09
Add GitHub Actions CI

itshaydendev:
Related issue: #94 

So far, this runs the following on Ubuntu, macOS and Windows:

```shell
go test ./internal/...
npm i
node scripts/verify-source-map.js
node scripts/end-to-end-tests.js
node scripts/js-api-tests.js
```

Can't use `npm ci` until npm/cli#558 is resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,9 +8,13 @@ on:
 
 jobs:
 
-  build:
-    name: Build
-    runs-on: ubuntu-latest
+  esbuild:
+    name: esbuild CI
+    runs-on: ${{ matrix.os }}
+    strategy:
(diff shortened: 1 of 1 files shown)


evanw:
Cool! Thanks so much. It looks like there's some failure on Windows right now due to symbolic links, but that's something I'll look into.

Thanks for the pointer about `npm ci`. Most of the packages are just for benchmarks so I should move the packages needed for tests to a separate, smaller package.json. I bet then `npm ci` will work. At the very least tests will run faster.

itshaydendev:
@evanw If they're only for dev, why not make them dev dependencies?

Anyway, `npm ci` fails because it ignores platform

PR #132, opened 2020-05-22
add jsconfig support

viankakrisna:


evanw:
Thanks for taking a stab at this.

I haven't used `jsconfig.json` before and I'd like to understand it better if esbuild is going to support it. Right now I can't figure out how to test it. According to [the documentation](https://code.visualstudio.com/docs/languages/jsconfig) `jsconfig.json` is supposed to be just like `tsconfig.json` except with the `"allowJs"` attribute set to `true`, but I am unable to confirm this.

### tsconfig.json

I understand how `tsconfig.json` works. You can test whether or not `tsconfig.json` has an effect by configuring the `module` property, which affects the generated JavaScript. For example, if I make a `tsconfig.json` file like this:

```json
{ "compilerOptions": { "module": "amd" } }
```

And then compile this `foo.ts` file by running `tsc`:

```ts
import * as fs from 'fs'
console.log(fs)
```

I get the following JavaScript file:

```js
define(["require", "exports", "fs"], function (require, exports, fs) {
    "use strict";
    Object.defineProperty(exports, "__esModule", { value: true });
    console.log(fs);
});
```

It uses AMD so I can tell that the TypeScript compiler is actually reading `tsconfig.json`.

### jsconfig.json

I tried the same thing with `jsconfig.json`. I made this `jsconfig.json` file:

```json
{ "compilerOptions": { "module": "amd" } }
```

And this `foo.js` file:

```js
import * as fs from 'fs'
console.log(fs)
```

Then I tried to invoke the TypeScript compiler such that it picks up the settings in `jsconfig.json`. I wasn't able to do this though. Here's what I tried:

```sh
# This generates JavaScript in CommonJS format,
# not AMD format (so jsconfig.json was ignored)
tsc --outDir out --allowJs foo.js

# This doesn't generate any output files at all
tsc --outDir out --project jsconfig.json
```

Can you help me understand how `jsconfig.json` works? I'd like to understand how and when the TypeScript compiler parses that file.

viankakrisna:
from what i observe, jsconfig is usually only used to define paths and resolve mechanism for user space modules so vscode/supported compiler can know about the custom module resolution. This is the configuration that's honored by Create React App and several other compiler. AFAIK, no other compiler honor anything other than baseUrl and paths for compilerOptions for js files.

probably you still need --allowJs ? and i noticed that you missed the entry file on your testing
```
tsc --outDir out --project jsconfig.json --allowJs foo.js
```

this line  from docs
> jsconfig.json is tsconfig.json with "allowJs" attribute set to true

is probably a convention only used in vscode to run typecheck on js only project with --allowJs

as per now, esbuild only support baseUrl in a tsconfig files so we can setup module resolution even in a js only project, which is weird because that breaks expectation, new member on the project probably thought that the project is written in TS by the existence of tsconfig in the root folder, while in fact we only use it for setting up baseUrl

PR #135, opened 2020-05-23
add file loader

viankakrisna:


internal/bundler/bundler.go
```diff
@@ -93,15 +94,15 @@ func parseFile(
...
 		PrettyPath:   prettyPath,
 		Contents:     contents,
 	}
 
 	// Get the file extension
 	extension := ""
-	if lastDot := strings.LastIndexByte(path, '.'); lastDot >= 0 {
-		extension = path[lastDot:]
+	if lastDot := strings.LastIndexByte(absolutePath, '.'); lastDot >= 0 {
+		extension = absolutePath[lastDot:]
```
viankakrisna, line 105: this line is probably could be simplified with https://golang.org/pkg/path/#Ext

internal/bundler/bundler.go
```diff
@@ -93,15 +94,15 @@ func parseFile(
...
+		AbsolutePath: sourcePath,
 		PrettyPath:   prettyPath,
 		Contents:     contents,
 	}
 
 	// Get the file extension
 	extension := ""
-	if lastDot := strings.LastIndexByte(path, '.'); lastDot >= 0 {
-		extension = path[lastDot:]
+	if lastDot := strings.LastIndexByte(sourcePath, '.'); lastDot >= 0 {
```
viankakrisna, line 104: this line is probably could be simplified with https://golang.org/pkg/path/#Ext thinking @evanw any reason we don't just use it?

internal/bundler/bundler_test.go
```diff
@@ -9,11 +10,14 @@ import (
...
 	"github.com/evanw/esbuild/internal/resolver"
+	"github.com/kylelemons/godebug/diff"
 )
 
 func assertEqual(t *testing.T, a interface{}, b interface{}) {
 	if a != b {
-		t.Fatalf("%s != %s", a, b)
+		stringA := fmt.Sprintf("%v", a)
+		stringB := fmt.Sprintf("%v", b)
+		t.Fatalf(diff.Diff(stringA, stringB))
```
viankakrisna, line 20: i've added this change because it's hard to diff any whitespace differences in tests when it happens

cmd/esbuild/main.go
```diff
@@ -46,7 +46,7 @@ Options:
   --jsx-factory=...     What to use instead of React.createElement
   --jsx-fragment=...    What to use instead of React.Fragment
   --loader:X=L          Use loader L to load file extension X, where L is
-                        one of: js, jsx, ts, tsx, json, text, base64, dataurl
+                        one of: js, jsx, ts, tsx, json, text, base64, url, dataurl
```
evanw, line 49: This feature seems similar to the Webpack [file-loader](https://webpack.js.org/loaders/file-loader/) plugin, so I think people will understand the name `file` better than `url`. There is also a Rollup [url](https://github.com/rollup/plugins/tree/master/packages/url) plugin which does something different (it does what `dataurl` does in esbuild), so that may also be confusing for people coming from Rollup.

internal/bundler/bundler.go
evanw: Sounds good to me. I probably just wasn't aware of it existing.

internal/bundler/bundler.go
```diff
@@ -166,8 +167,19 @@ func parseFile(
 		ast := parser.ModuleExportsAST(log, source, parseOptions, expr)
 		results <- parseResult{source, ast, true}
 
+	case LoaderURL:
+		url := path.Base(sourcePath)
+		targetFolder := bundleOptions.AbsOutputDir
+		if targetFolder == "" {
+			targetFolder = path.Dir(bundleOptions.AbsOutputFile)
+		}
+		defer ioutil.WriteFile(path.Join(targetFolder, url), []byte(source.Contents), 0644)
```
evanw, line 176: The bundler isn't supposed to write files itself. The bundler returns all of the files to be written to the caller. This makes the bundler easier to unit test and also makes it easy for the API to return results without modifying the file system (see for example #139).

Doing things this way also defeats future tree shaking optimizations, which may end up removing this module entirely. In that case the file should not be emitted in the output directory since it's unreferenced.

Instead of writing to the file system, the contents of the file should be stashed somewhere associated with the file, probably just on the `parseResult` or `ast`. Then the bundler should return the file to the caller, probably as an extra `BundleResult` entry. But that requires generalizing `BundleResult` to handle non-JavaScript result files.

viankakrisna: thank you! yeah this is a part where i'm not really sure. Do you have any plans for generalizing `BundleResult`? I could try working on it but don't want to duplicate the efforts

cmd/esbuild/main.go
viankakrisna: that's true! url-loader seems to have more specific feature in webpack world https://webpack.js.org/loaders/url-loader/ we could use `url` to automatically switch between `dataurl` and `file` in the future i guess. i'll rename this to `file`

cmd/esbuild/main.go
```diff
@@ -635,6 +637,17 @@ func run(fs fs.FS, args argsObject) {
 		}
 		args.logInfo(fmt.Sprintf("Wrote to %s (%s)", path, toSize(len(item.JsContents))))
 
+		// Write out the additional files
+		for _, file := range item.AdditionalFiles {
```
evanw, line 641: The contents of `AdditionalFiles` end up being written out twice because they are both written out here and duplicated to become normal files in this code elsewhere:

```go
		for _, bundle := range group {
			for _, additionalFile := range bundle.AdditionalFiles {
				if additionalFile.Path != "" {
					results = append(results, BundleResult{additionalFile.Path, []byte(additionalFile.Contents), "", []byte(""), []AdditionalFile{}})
				}
			}
		}
```

That's ok, I can fix it after landing this.

internal/bundler/bundler_test.go
```diff
@@ -1,19 +1,28 @@
...
+	"github.com/kylelemons/godebug/diff"
 )
 
 func assertEqual(t *testing.T, a interface{}, b interface{}) {
 	if a != b {
-		t.Fatalf("%s != %s", a, b)
+		stringA := fmt.Sprintf("%v", a)
+		stringB := fmt.Sprintf("%v", b)
+		if strings.Contains(stringA, "\n") {
+			t.Fatal(diff.Diff(stringA, stringB))
```
evanw, line 22: I understand why you did this, but this is going to make it harder to update tests since I've been copying and pasting the output of the tests to accept the new output, and this interleaves the old and new output. Luckily it's easy to disable by just commenting out a few lines, so I don't think it's a big problem.

Really I should figure out some system of doing snapshot-style testing so that test output can be updated automatically...

PR #142, opened 2020-05-25
add resolve-extension-order

viankakrisna:


evanw:
This is an interesting feature. I'm not sure about the name `--resolve-extension-order` though. It's relatively long and I'm not sure I'd understand what it does given the name. What do you think about something like `--implicit-extensions` instead?

viankakrisna:
hmm that name is coming from the actual variable it's replacing in the resolver, webpack already has resolve.extensions which does the same thing https://webpack.js.org/configuration/resolve/

i think resolve-extensions is fine if we want to shorten it. this would help anyone migrating from webpack, but i dont really have a strong preferences either way

evanw:
Good to know about the precedent from Webpack. In that case `--resolve-extensions` sounds fine to me.

PR #143, opened 2020-05-25
Add: add Chinese docs link to README.md

92hackers:
@evanw , If the position of documentation section in README.md is not what you want or something else, comment it and i will modify.

And, i will continue maintain the Chinese docs of this project once it connects with repo.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,11 @@ Several reasons:
 * Everything is done in very few passes without expensive data transformations
 * Code is written with speed in mind, and tries to avoid unnecessary allocations
 
+## Documentation
+
+* [architecture](docs/architecture.md)
+* [中文文档](http://docs.breword.com/evanw-esbuild/)
+
 ## Status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,11 @@ Several reasons:
 * Everything is done in very few passes without expensive data transformations
 * Code is written with speed in mind, and tries to avoid unnecessary allocations
 
+## Documentation
+
+* [architecture](docs/architecture.md)
+* [中文文档](http://docs.breword.com/evanw-esbuild/)
```
ycjcl868, line 59: Why not adding into the current repo?

92hackers: Because translation docs my be outdated with the time going.

As follow suggestion from @evanw , add a link is better, and i will specific the version of the source code that translation docs represents.

evanw:
Awesome! Thanks so much for doing all of this translation work.

PR #144, opened 2020-05-26
add tsconfig.json compilerOptions.paths support

viankakrisna:


internal/resolver/resolver.go
```diff
@@ -578,13 +601,43 @@ func (r *resolver) loadAsFileOrDirectory(path string) (string, bool) {
...
-			basePath := r.fs.Join(*dirInfo.tsConfigJson.absPathBaseUrl, path)
-			if absolute, ok := r.loadAsFileOrDirectory(basePath); ok {
-				return absolute, true
+		if dirInfo.tsConfigJson != nil {
+
+			if dirInfo.tsConfigJson.absPathBaseUrl != nil {
+				if dirInfo.tsConfigJson.paths != nil {
+					for key, originalPaths := range dirInfo.tsConfigJson.paths {
+						for _, originalPath := range originalPaths {
+							if matched, err := regexp.MatchString("^"+key, path); matched && err == nil {
```
evanw, line 618: Using a regex to match this isn't correct. The pattern is an asterisk, which causes a regex to repeat the previous character zero or more times. That would mean a pattern `foo*` would also match a directory named `foooo` and `fo`. I took some time to look at what the TypeScript compiler itself does. The relevant code is in `tryLoadModuleUsingPaths()` [here](https://github.com/microsoft/TypeScript/blob/c194abb38c54c1607831b4ec678c0bdd15f7fe49/src/compiler/moduleNameResolver.ts#L1361-L1387).

Other relevant code is `tryParsePattern()` [here](https://github.com/microsoft/TypeScript/blob/c194abb38c54c1607831b4ec678c0bdd15f7fe49/src/compiler/utilities.ts#L6122-L6130):

```ts
    export function tryParsePattern(pattern: string): Pattern | undefined {
        // This should be verified outside of here and a proper error thrown.
        Debug.assert(hasZeroOrOneAsteriskCharacter(pattern));
        const indexOfStar = pattern.indexOf("*");
        return indexOfStar === -1 ? undefined : {
            prefix: pattern.substr(0, indexOfStar),
            suffix: pattern.substr(indexOfStar + 1)
        };
    }
```

and `isPatternMatch()` [here](https://github.com/microsoft/TypeScript/blob/c194abb38c54c1607831b4ec678c0bdd15f7fe49/src/compiler/core.ts#L2015-L2019):

```ts
    function isPatternMatch({ prefix, suffix }: Pattern, candidate: string) {
        return candidate.length >= prefix.length + suffix.length &&
            startsWith(candidate, prefix) &&
            endsWith(candidate, suffix);
    }
```

internal/resolver/resolver.go
```diff
@@ -578,13 +601,43 @@ func (r *resolver) loadAsFileOrDirectory(path string) (string, bool) {
...
+									if absolute, ok := r.loadAsFileOrDirectory(basePath); ok {
+										return absolute, true
+									}
+								}
+							}
+						}
+
+					}
+				} else {
+					basePath := r.fs.Join(*dirInfo.tsConfigJson.absPathBaseUrl, path)
```
evanw, line 635: It looks like the TypeScript compiler still respects `baseUrl` if no `paths` matched, so we should do that too.

internal/resolver/resolver.go
```diff
@@ -578,13 +601,43 @@ func (r *resolver) loadAsFileOrDirectory(path string) (string, bool) {
 	return "", false
 }
 
+func resolvePathWithoutStar(from, path string) (string, error) {
+	replaced := strings.Replace(path, "/*", "", -1)
+	return fp.Join(from, replaced), nil
```
evanw, line 606: The code in this file is careful to always use the `fs` interface in the `resolver` instead of `path/filepath`. All platform-dependent things are supposed to happen inside the `fs` interface so that the resolver can be platform-independent. That interface uses the real `path/filepath` module for the esbuild command and the `path` module for tests. That way tests don't do different things on Windows.

internal/resolver/resolver.go
viankakrisna: thanks! resolved on 2587e08d326018a478aa1b0ec1c8b0837ad846a2

internal/resolver/resolver.go
viankakrisna: thank you! resolved in 444bfae

internal/resolver/resolver.go
viankakrisna: thank you! i've added some commits to handle this more like ts compiler, but i haven't written the test case where `*` is at the start or middle of the path.

internal/bundler/bundler_test.go
```diff
@@ -1386,6 +1386,72 @@ console.log(main_esm_default());
 	})
 }
 
+func TestTsConfigPaths(t *testing.T) {
+	expectBundled(t, bundled{
+		files: map[string]string{
+			"/Users/user/project/src/entry.js": `
+				import fn from 'core/test'
+				import fn2 from 'testing/test'
```
evanw, line 1394: It appears this case isn't actually supposed to work. At least the TypeScript compiler throws an error here. I can do the testing work and make adjustments to the algorithm as appropriate.

internal/resolver/resolver.go
```diff
@@ -578,13 +598,47 @@ func (r *resolver) loadAsFileOrDirectory(path string) (string, bool) {
...
-		if dirInfo.tsConfigJson != nil && dirInfo.tsConfigJson.absPathBaseUrl != nil {
-			basePath := r.fs.Join(*dirInfo.tsConfigJson.absPathBaseUrl, path)
-			if absolute, ok := r.loadAsFileOrDirectory(basePath); ok {
-				return absolute, true
+		if dirInfo.tsConfigJson != nil {
+
+			if dirInfo.tsConfigJson.absPathBaseUrl != nil {
+				if dirInfo.tsConfigJson.paths != nil {
+					for key, originalPaths := range dirInfo.tsConfigJson.paths {
+						for _, originalPath := range originalPaths {
```
evanw, line 625: This doesn't follow the behavior of the TypeScript compiler which prefers exact matches over pattern matches and prefers matches with longer prefixes over matches with shorter prefixes. That's ok though, I'll fix it.

PR #146, opened 2020-05-26
Support PowerPC 64 LE architecture

runlevel5:
This PR adds support for PowerPC64 LE Linux.


evanw:
Woah, I didn't expect someone to want to use esbuild on a platform like this. What's your use case, if you don't mind me asking? I'm curious to know where esbuild is being used.

Keep in mind that even if this PR lands, this platform won't be officially supported because it won't be covered by automated tests.

runlevel5:
> Woah, I didn't expect someone to want to use esbuild on a platform like this. What's your use case, if you don't mind me asking? I'm curious to know where esbuild is being used.

It is common that people assume PowerPC is only used for HPC or some 'enterprise' application.

My usecase is for web frontend development just like how it is done under x84_64 platform. FYI the PowerPC that we know today is that of OpenPOWER. IBM/RedHat has put much resources to make PowerPC system as easy and intuitive as that of x86_64.

Regarding automated test, please try Travis CI, they do support ppc64le.

darkbasic:
Thanks for providing ppc64le support, I use my OpenPOWER workstation for fullstack web development as well.

Issue #152, opened 2020-05-26
Any plan for a Go API?


bep:
I'm leading the Hugo project, and esbuild delivers something I have been wanting badly to integrate for a long, long time.

Unfortunately, for good reasons I'm sure, there does not seem to be a public Go API. The `main` function is easy to understand and seem to cover what we need, but all the API it calls lives in the `internal` package.

Any plans making (at least part of) this public? Note that I don't mind API-breakage in < 1 versions and would love to get to test this out.

https://github.com/gohugoio/hugo/issues/7321

carlmjohnson:
IMO, a good compromise if the project wants to keep its APIs internal is to create an `app.CLI(args []string) error` public function that main can call. This lets other Go projects embed esbuild without having to install a binary somewhere or committing to any public interface beyond the CLI. I use this pattern quite often, e.g. [here](https://github.com/spotlightpa/poor-richard/commit/fd30159d60f37d4d86e698717045fa7d7c7fa3e6).

bep:
>app.CLI(args []string) error

Not to get ahead of myself here, but for this to be practical/usable for Hugo, we would need access to the filesystem/resolver API, i.e. something more powerful than the above.

carlmjohnson:
I can see why the esbuild team might not want to commit to an external API now, but I think accepting #153 is pretty much a no-brainer, since it doesn't add any new API surface but makes the existing CLI interface easier to use.

evanw:
> Not to get ahead of myself here, but for this to be practical/usable for Hugo, we would need access to the filesystem/resolver API, i.e. something more powerful than the above.

Can you say what your requirements would be explicitly? That could help inform the discussion.

I could see exposing a Go-based transform API to mirror the [existing JavaScript-based API](https://github.com/evanw/esbuild/blob/45dc7e7237638e7429b359c494b1279fad46f5ad/README.md#transforming-a-file). That would let you use esbuild as a minifier or as a way to convert JSX/TypeScript to JavaScript. It's essentially the same as the stable stdin/stdout command-line interface but exposed such that you don't need to touch the file system.

bep:
>Can you say what your requirements would be explicitly?

[...]

PR #153, opened 2020-05-26
Break out esbuild.CLI public Go CLI API

carlmjohnson:
Refs #152


internal/exitcode/exitcode.go
```diff
@@ -0,0 +1,64 @@
+package exitcode
```
carlmjohnson, line 1: Copypasta from my own github.com/carlmjohnson/exitcode

carlmjohnson:
Make cmd/esbuild a one liner call into pkg/esbuild, which has a single public function: `esbuild.CLI(osargs []string) error`. This allows esbuild to be embedded into other Go application without interprocess calls _or_ introducing a new public API beyond the CLI.

evanw:
I see why you would like to do this. However, I don't think this is a good way to integrate esbuild into other Go apps. There would be no way to programmatically list out build errors, for example. This approach also has other issues such as causing GC to be disabled because esbuild disables GC to improve performance when invoked in batch mode.

Instead of doing this, I'd rather have an API more like the [current JavaScript API](https://github.com/evanw/esbuild/blob/45dc7e7237638e7429b359c494b1279fad46f5ad/README.md#javascript-api-usage). That API is future proof like this approach but also more usable. It'd be a much bigger change though, so perhaps not appropriate for a PR.

evanw:
After a major refactor, the Go CLI is now available in version 0.5.0. I have also added build and transform APIs which are more flexible than the CLI API. The new Go APIs are documented [here](https://github.com/evanw/esbuild/blob/master/docs/go-api.md).

carlmjohnson:
Great!

PR #164, opened 2020-05-31
Create nested directory if needed

quasor:


evanw:
There is already a call to `os.MkdirAll` above. Can you provide a description of the problem you're running into and a way to reproduce the problem?

quasor:
Looks like the issue is the file suffix can contain a `/` 
e.g. `Wrote to out/preview_twitter.yDerV/Jy.svg (828 bytes)`

PR #177, opened 2020-06-11
Add Godoc badge

tooolbox:
Godoc is a common way for Go devs to look at a package.  It only shows exported methods and types so you get a clear view of the API.  Having a Godoc badge provides a handy shortcut.

I made the badge point at `pkg/api` since that is the main integration point and I'm guessing will see more use than `pkg/cli`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # esbuild
 
+[![Godoc](https://godoc.org/github.com/evanw/esbuild?status.svg)](https://godoc.org/github.com/evanw/esbuild/pkg/api)
+
 This is a JavaScript bundler and minifier. It packages up JavaScript and TypeScript code for distribution on the web.
 
 ## Documentation


evanw:
Sure, sounds good to me. I just moved the Go documentation over to godoc so the information and examples will be there too.

shilangyu:
Godoc has already been deprecated, the badge should point to [pkg.go.dev](https://pkg.go.dev/github.com/evanw/esbuild/pkg/api?tab=doc) instead.

tooolbox:
> deprecated
> 
> (chiefly of a software feature) be usable but regarded as obsolete and best avoided, typically due to having been superseded.

Not really obsolete, not particularly best to avoid it, not yet superseded.

1. New documentation site doesn't have all of the features of the old one yet.
2. New documentation site has bugs not present in the old one.
3. New badge isn't available yet, as you mention.

At some point it will be superseded, at which point this and many other projects will need to switch.  Although, godoc will redirect anyway, so it's minor bookkeeping.

References:
- https://github.com/golang/go/issues/37819
- https://github.com/golang/go/issues/39144

evanw:
When I moved the documentation over I took a look at both and chose godoc.org because I think the presentation looks better. I'm going to leave it linking to godoc.org. It will redirect to pkg.go.dev anyway when pkg.go.dev is ready.

Issue #96, opened 2020-05-09
Inject script tag into html after bundling (using with React)


oze4:
I love this project and was looking for the ability to inject built files, as script tags, into my `index.html` file after bundling.  I was not able to find anything online about injecting script tags with `esbuild`, so forgive me if this type of functionality is out of scope.

The functionality I am looking for is comparable to the [`HtmlWebpackPlugin`](https://webpack.js.org/plugins/html-webpack-plugin/), in a `webpack` world...

Regardless, thank you for your amazing work! :)

evanw:
If esbuild does support ways of working with HTML, I've been thinking that the solution would look more like #31 where you give esbuild a HTML file with `<script>` tags already in it and then esbuild would add the linked JavaScript files as entry points, build them, and generate a new HTML file with the `<script>` tag rewritten to point to the generated JavaScript bundle. This is similar to how Parcel works.

The reason is because people using things like `HtmlWebpackPlugin` potentially have a lot of custom requirements, and it wouldn't make sense to add all of that to esbuild as built-in options. There's a whole template system and many configuration options. Using a HTML file as an entry point naturally gives you complete control over exactly what the HTML file looks like and lets esbuild just handle the bundling part.

However, if that doesn't work and something like `HtmlWebpackPlugin` is still desired, it should be very easy for someone else to build it as a wrapper for esbuild or as a post-processing step to call after esbuild is done. That way people can address those custom requirements without that complexity being added to esbuild itself.

oze4:
@evanw thank you very much for your response!  I suppose I was using the webpack plugin as an example, not exactly looking for a 1:1. Using an HTML file as an entry point sounds like a solid idea to me :smile: 

Regardless of this feature, esbuild has made it extremely simple, quick, and easy to bundle/transpile code. Truly an amazing project and it's exactly what I have been searching for. So much easier and straight forward to use than Rollup or Webpack, as far as bundling/transpiling is concerned. 

It also looks like some folks have turned esbuild into an [webpack plugin](https://github.com/sorrycc/esbuild-webpack-plugin) so at the very least I could use esbuild inside webpack along with the HtmlWebpackPlugin for React projects. 

Thanks again for your response!

ChildishGiant:
I'm going to be using ejs to insert the script tags with appropriate integrities etc

vikasgarghb:
@ChildishGiant How did you integrate ejs with esbuild?

ChildishGiant:
[...]

Issue #20, opened 2020-02-18
[MVP] Bundle CSS modules


evanw:
I want esbuild to demonstrate that it's possible for a normal web development workflow to have high-performance tools. I consider bundling CSS modules a part of my initial MVP feature set for esbuild since some form of CSS modularity is essential for a large-scale web app and I find CSS modules an elegant solution. This issue tracks the CSS module bundling part of my MVP.

At the bare minimum, importing a CSS file should ensure it's included in the bundle. I plan to also implement [CSS modules](https://github.com/css-modules/css-modules), which makes CSS selector names local to the file and allows JavaScript to import the selector names as strings.

There are interesting extensions to think about once I get everything working. Tree shaking of unused CSS is an optimization that seems worth exploring, for example. That should be possible with CSS modules since unused imports correspond to unused CSS.

eberkund:
Would this include font files, or is that a separate task?

evanw:
Binary resources that don't need to be parsed such as PNG, JPEG, and font files are all the same to a bundler. Fonts don't need to be special cased as far as I'm aware. I think the existing loaders such as `dataurl` and `file` should cover fonts.

evanw:
I'm starting to work on CSS support. I just landed a basic parser and AST, and will be working on integrating it into the build system next.

Note that I'm not planning to support the full CSS ecosystem with this feature. Today's CSS is a very diverse ecosystem with many non-standard language variants. This is possible in part due to the fault-tolerant nature of CSS. I'm imagining this feature as mainly a "CSS linker" that is responsible for the critical parts of joining CSS files together and removing unused CSS code. The CSS parser will be expecting real CSS for use with browsers. Non-standard extensions may still be parsed without syntax errors but may be passed through unmodified (e.g. not properly minified or renamed) since they aren't fully understood by the parser. It should be possible to use syntax extensions (e.g. SASS) with a plugin that transforms the source code before esbuild reads it.

LarsDenBakker:
@evanw that's great to hear! There are different kinds of CSS modules out there, and there is also ongoing for a web standard for CSS modules (see https://github.com/whatwg/html/pull/4898 and https://github.com/tc39/proposal-import-assertions). WIll esbuild allow specifying which variant to use?

nitsky:
[...]

PR #180, opened 2020-06-14
Add empty loader

viankakrisna:
this is useful in case we want to have an empty string as the result of importing a module, for example: while running tests we don't want to emit files or load the full contents of the imported file to the memory.

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -216,6 +216,10 @@ func parseFile(args parseArgs) {
 			jsonMetadataChunk: jsonMetadataChunk,
 		}
 
+	case LoaderEmpty:
+		expr := ast.Expr{ast.Loc{0}, &ast.EString{lexer.StringToUTF16("")}}
+		result.file.ast = parser.ModuleExportsAST(args.log, source, args.parseOptions, expr)
+
 	default:
 		result.ok = false
 		args.log.AddRangeError(args.importSource, args.pathRange,
@@ -424,6 +428,7 @@ const (
 	LoaderBase64
 	LoaderDataURL
 	LoaderFile
+	LoaderEmpty
 )
 
 func DefaultExtensionToLoaderMap() map[string]Loader {
--- a/internal/bundler/bundler_test.go
+++ b/internal/bundler/bundler_test.go
@@ -2803,6 +2803,41 @@ func testAutoDetectMimeTypeFromExtension(t *testing.T) {
 	})
 }
 
+func TestLoaderEmpty(t *testing.T) {
+	expectBundled(t, bundled{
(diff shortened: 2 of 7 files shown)


floydspace:
Hey @viankakrisna, maybe the result of the loader should be `null` or `undefined`? not all file types return a string

viankakrisna:
@floydspace i think that could be a separate loader. the use case for this is to replace the loaders that's usually output string to js like `file`, `text`, `base64`, and `dataurl`, but we don't want to use them in some cases. probably `emptystring` is a better name for this loader?

floydspace:
@viankakrisna as long as there is already `text` loader we could name it `emptytext` or `mocktext`

evanw:
I've been thinking about this for a bit. I agree that something like `emptytext` would be a better name for this since I'd expect something named `empty` to cause the file to be empty.

However, I think this is starting to get at more complicated loader scenarios, and I want to think more carefully about loaders like this to avoid adding something that doesn't fit well into the overall design.

I've been thinking about loader chaining recently. Really this loader is two concepts: forcing the input file to be empty and packaging up a file as a raw string. It would be nice if you could say `--loader=empty,text` instead to avoid needing to special-case the combination of these two concepts.

This has come up for other loaders in the past too. For example, at one point #107 included code for a SVG loader, except it combined the two concepts of loading an SVG file and packaging up the file as a data URL. The SVG loader would be a natural place to add SVG-specific optimizations over time (e.g. minification). But if you wanted the SVG file as a raw string instead of a data URL, you'd be out of luck. Or at least you'd have to add a lot of extra options to the SVG loader for different output formats, or you'd have to make lots of SVG loaders for different combinations of concepts. It seems like a cleaner design to me to be able to do something like `--loader=svg,dataurl` and keep the concepts separate by chaining loaders. So I'm thinking this is the direction I'd like to take loaders.

However, I want to think carefully about how this works in the general case. I also need loaders to support HTML (#96) and CSS (#20). I'm not sure how I feel about requiring all loaders to ultimately produce a JavaScript file. It might be nice to be able to just bundle CSS to CSS or just minify some HTML, for example. Some loaders also need to produce multiple output files (e.g. the `file` loader, also CSS modules in the future). So I haven't yet decided what the right interface between loaders in a loader chain is.

I also haven't yet done research into how other bundlers handle their loader APIs, and how they solve these problems. Ideally that research would be done as part of this too.

evanw:
This specific scenario and other similar scenarios are already addressable with the [on-load plugin API](https://esbuild.github.io/plugins/#load-callbacks) which was added since this PR was opened. I'm closing this because I don't think this empty file feature should be built-in given that it's pretty specific and is already covered by plugins.

PR #186, opened 2020-06-18
fix(externals): exclude external child paths from the bundle

floydspace:
As an example, if `aws-sdk` is an external, imports from `aws-sdk/clients/dynamodb` must be excluded from the bundle as well.

Does it make sense @evanw ?

### DIFF
--- a/internal/bundler/bundler_test.go
+++ b/internal/bundler/bundler_test.go
@@ -4846,6 +4846,44 @@ render(h(App, null), document.getElementById("app"));
 	})
 }
 
+func TestExternalModuleExclusion(t *testing.T) {
+	expectBundled(t, bundled{
+		files: map[string]string{
+			"/index.js": `
+				import { S3 } from 'aws-sdk';
+				import { DocumentClient } from 'aws-sdk/clients/dynamodb';
+				export const s3 = new S3();
+				export const dynamodb = new DocumentClient();
+			`,
+		},
+		entryPaths: []string{"/index.js"},
+		parseOptions: parser.ParseOptions{
+			IsBundling: true,
+		},
+		bundleOptions: BundleOptions{
+			IsBundling:    true,
+			AbsOutputFile: "/out.js",
+		},
+		resolveOptions: resolver.ResolveOptions{
+			ExternalModules: map[string]bool{
+				"aws-sdk": true,
+			},
+		},
+		expected: map[string]string{
+			"/out.js": `// /index.js
+import {S3} from "aws-sdk";
+import {DocumentClient} from "aws-sdk/clients/dynamodb";
+const s3 = new S3();
+const dynamodb2 = new DocumentClient();
+export {
+  dynamodb2 as dynamodb,
+  s3
+};
+`,
+		},
+	})
+}
+
 // This test case makes sure many entry points don't cause a crash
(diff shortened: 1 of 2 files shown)


internal/bundler/bundler_test.go
```diff
@@ -4846,6 +4846,44 @@ render(h(App, null), document.getElementById("app"));
...
+			},
+		},
+		expected: map[string]string{
+			"/out.js": `// /index.js
+import {S3} from "aws-sdk";
+import {DocumentClient} from "aws-sdk/clients/dynamodb";
+const s3 = new S3();
+const dynamodb2 = new DocumentClient();
+export {
+  dynamodb2 as dynamodb,
```
floydspace, line 4879: @evanw as a side question, could you tell why this is bundled with suffix, I expected that it would be direct variable, should it be considered as a bug?

evanw: This happens because of a complex interaction between the parser and the linker. The parser pre-allocates a namespace symbol for each import statement in case it's needed by the linker later. The hidden namespace is used in various cases such as the variable to store the result of `require()` for the CommonJS format. So for example this code might be turned into something like this for CommonJS:

```js
const ns1 = __toModule(require("aws-sdk"));
const ns2 = __toModule(require("aws-sdk/clients/dynamodb"));
const s3 = new ns1.S3();
const dynamodb = new ns2.DocumentClient();
```

This is done at parse time instead of link time because esbuild is trying to do as few passes as possible. Instead of doing an extra pass after parsing and before linking to rewrite external imports such as `S3` into property accesses such as `ns1.S3`, that rewriting pass is merged with the parser pass. The parser doesn't yet know if the imported module is external or not because that's computed in parallel, so it pessimistically assumes that all modules could be external and always does this transformation, which involves generating a namespace symbol for each import.

Except to make code more readable when not minifying, the name of the imported namespace is derived from the last part of the file path. This doesn't affect correctness at all but is more readable than `ns1, ns2, ...` especially when there are a lot of namespaces. So the two imports really turn into something more like this:

```js
const aws_sdk = __toModule(require("aws-sdk"));
const dynamodb = __toModule(require("aws-sdk/clients/dynamodb"));
```

But the name `dynamodb` collides with the existing symbol with the same name, which would be invalid JavaScript. The linker has a general-purpose solution that automatically renames all name collisions. That way earlier passes in the compiler can ignore the complexity of generating unique names for automatically-generated symbols and renaming can happen all in one place where it's easy to understand and audit.

What happens here is that the import statement is processed first and grabs the `dynamodb` name for the namespace, then the variable statement is processed next and must be renamed to `dynamodb2`. You'll see that the variable statement will grab the `dynamodb` name first if you reorder the conflicting import statement to come after the variable statement.

I do not consider this a bug because it does not affect correctness or minified build size. The minifier will use one-letter names for all of these identifiers, so the original names don't matter. The pretty-printing of the bundled code when not minifying just is a best-effort thing that tries to ensure the code is readable but doesn't guarantee that the symbols keep their original names. In the general case, keeping the original names is impossible anyway because ES6 the same export may be imported and re-exported using different aliases and many different symbols may be represented by a single symbol after scope flattening happens during linking.

floydspace: Thank you for the detailed explanation, clear like the sky

PR #200, opened 2020-06-27
Make `npm install`-able on platforms supported by Go

kikuchan:
Although my platform is supported by Go, `npm install` fails because of a prebuilt binary is missing for the platform.
I think adding every platform supported by Go to the installation script and npm registry every time would be cost.

This adds install time compilation of `esbuild` binary.

I'm not sure this patch is the right way to build binary on-the-fly in npm ecosystem, but it seems to work on my platform (FreeBSD in this case) at least.

### DIFF
--- a/lib/api-node.ts
+++ b/lib/api-node.ts
@@ -13,13 +13,6 @@ let esbuildCommandAndArgs = (): [string, string[]] => {
   let platform = process.platform;
   let arch = os.arch();
 
-  if (
-    (platform === 'darwin' && arch === 'x64') ||
-    (platform === 'linux' && (arch === 'x64' || arch === 'arm64' || (arch === 'ppc64' && os.endianness() === 'LE')))
-  ) {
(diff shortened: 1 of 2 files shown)


evanw:
Thank you for wanting to fix this. However, this is not the way I'd like to solve this issue.

I want to make esbuild a good citizen in the frontend ecosystem. The majority of people in today's frontend ecosystem do not have the Go toolchain installed, and I don't want installing that to be a requirement for using esbuild.

Instead I would like to have precompiled binaries available for the platforms people need. Go's excellent cross-platform compilation support makes this trivial on my end so I don't mind adding more platform targets.

I could see potentially having a fallback to the WebAssembly version if there's no native binary available, but it's really pretty terrible compared to the native version (one measurement was ~10x slower) so I'd only want to do that for a platform that Go's compiler doesn't have support for.

kikuchan:
Thank you for the reply, and thank you for the great tool.

OK, I understand.
I'll try to create a new patch to support the platform in another way.

PR #217, opened 2020-07-03
Add support for FreeBSD

kikuchan:
...and, make it easy to support additional UNIX-like systems for the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,5 @@
 ESBUILD_VERSION = $(shell cat version.txt)
 
-# Supported UNIX-like platforms. notation: GOOS/GOARCH
-PLATFORMS_UNIXLIKE += darwin/amd64
-PLATFORMS_UNIXLIKE += freebsd/amd64
-PLATFORMS_UNIXLIKE += freebsd/arm64
-PLATFORMS_UNIXLIKE += linux/amd64
-PLATFORMS_UNIXLIKE += linux/arm64
-PLATFORMS_UNIXLIKE += linux/ppc64le
-
 esbuild: cmd/esbuild/*.go pkg/*/*.go internal/*/*.go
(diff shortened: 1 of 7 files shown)


evanw:
Thanks for putting this together. I understand why you refactored it this way, but I'd like to avoid metaprogramming in the makefile because I find it hard to read. I'll gladly accept the rest of the PR though. Feel free to remove the metaprogramming yourself, or I'm also happy to land it as-is and remove the metaprogramming myself afterwards if you prefer. Just let me know.

Also, can you confirm that you have tested esbuild's bundling functionality on both `freebsd/amd64` and `freebsd/arm64` and that it works correctly? I am unable to test this myself so this will be an unsupported build configuration, but I want to make sure someone else at least verifies that it works correctly before I start publishing builds.

kikuchan:
Thank you for the review.
I've removed the metaprogramming in the Makefile, but still try to put them together by using `platform-unixlike`. How about this?

I've also tested by running `make test-all` on `freebsd/amd64` natively and `freebsd/arm64` by using qemu, and they said all passed. Of course a generated executable works fine so far.

evanw:
That's great, thanks! And I appreciate you taking the time to test it.

evanw:
I tried to publish it but npm's publish command failed because of their aggressive package name spam detector. I have reached out to npm support and will try to publish again once the name has been unblocked.

kikuchan:
Oh, ok. I'm waiting for it.
Anyway, thank you for merging!

evanw:
The [esbuild-freebsd-64](https://www.npmjs.com/package/esbuild-freebsd-64) package has now been published (version 0.5.24).

kikuchan:
Thank you! I confirmed it works perfectly via `npm install`. I really appreciate it.

PR #223, opened 2020-07-05
Reuse the unicode isSpace function

valaymerick:
Reuse of the `unicode.IsSpace()` function of the standard Go library to check Unicode spaces, which includes spaces as defined by the ECMAScript spec.

### DIFF
--- a/internal/lexer/lexer.go
+++ b/internal/lexer/lexer.go
@@ -543,32 +543,10 @@ func IsIdentifierContinue(codePoint rune) bool {
 
 // See the "White Space Code Points" table in the ECMAScript standard
 func IsWhitespace(codePoint rune) bool {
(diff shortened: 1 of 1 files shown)


evanw:
Why do you want to make this change? Is there some problem with the existing code? If so, please file a bug report with a failing test case.

I don't believe this is a valid substitution because it looks like this change introduces regressions. For example, `unicode.IsSpace(0x85)` is true but `eval('var\x85test')` is a syntax error. There's a reason why I have listed out these cases explicitly here. You can also refer to the [White Space section in the specification](http://www.ecma-international.org/ecma-262/11.0/#sec-white-space) for more information.

valaymerick:
Thank you for your feedback. I will test eventual subsequent proposals better.

Issue #111, opened 2020-05-14
Support the esbuild plug-in system？


ycjcl868:
`esbuild` is small but complete in every detail, do you have any plans to support the plugin-in system to extend the web development workflows?

evanw:
Not right now. I may figure out an extensibility story within esbuild after the project matures, but I may also keep esbuild as a relatively lean bundler with only a certain set of built-in features depending on how the project evolves.

The use case of using esbuild as a library was one I hadn't originally considered. It's interesting to see it start to take off and I want to see where it goes. It could be that most users end up using other bundlers and esbuild is just an implementation detail that brings better performance to those bundlers.

I'm still thinking about how I might add extensibility to esbuild in the back of my mind. Obviously it's made more complicated by the fact that it's written in Go. It would be possible to "shell out" to other processes to delegate the task of transforming input files, but that would almost surely be a huge slowdown because JavaScript process startup overhead costs are really high.

One idea I've been thinking about is to have esbuild start up a set number of JavaScript processes (possibly just one) and then stream commands to it over stdin/stdout. That could potentially amortize some of the JavaScript startup cost (loading and JITing packages from disk). It would be more of a pain to debug and might still be surprisingly slow due to all of the serialization overhead and the single-threaded nature of the JavaScript event loop.

Another idea is to turn the esbuild repository into a Go-based "build your own bundler" kit. Then you could write plugins in Go to keep your extensions high-performance. The drawback is that you'd have to build your own bundler, but luckily Go compiles quickly and makes cross-platform builds trivial. That would likely require me to freeze a lot of the Go APIs that are now internal-only, which would prevent me from making major improvements. So this definitely isn't going to happen in the short term since esbuild is still early and under heavy development.

rsms:
Finding this interesting!

TL;DR: The subprocess approach sounds like a solid idea.

### Some thoughts in regards to the conversation above

> It would be possible to "shell out" to other processes to delegate the task of transforming input files, but that would almost surely be a huge slowdown because JavaScript process startup overhead costs are really high.

This is a really interesting approach I think, although not the most portable (i.e. running systems like iOS that do not support subprocess creation.) I'd think about the quality-performance problem here in the same way as with Figma plugins:
- Making it clear to the user that they are using "extra things"
- Warnings when things seems to be slow or hung (like how a web browser running a locked-up javascript might ask the user "Foo is taking a long time to run. Do you want to wait for Foo or kill it?".
[...]

PR #234, opened 2020-07-06
Proof-of-concept for transform plugins

matthewmueller:
Hey Evan,

I wanted to share a proof-of-concept for how esbuild could support simple 3rd-party plugins using the virtual filesystem. I don't expect this to be merged – this is throwaway code. If there's interest, I'd be more than happy to make this PR legit.

It's less ambitious than some of the ideas in #111, but still very flexible and the changes required are minimal. I also think it's quite fair to ask people to write their own `main.go` file to customize `esbuild`. This would be similar to how `rollup.config.js` or `webpack.config.js` works.
 
For a public `FileSytem` interface, it should be possible re-use the interface and implementation from [godoc](https://godoc.org/golang.org/x/tools/godoc/vfs/mapfs). That virtual filesystem is used quite a bit in the Go ecosystem.

```go
package main

import "golang.org/x/tools/godoc/vfs/mapfs"

func main() {
  fs := mapfs.New(map[string]string{...})
}
```


evanw:
I'm glad you're excited about esbuild plugins! I am too. This is very timely because I'm planning to start experimenting with a plugin API for esbuild soon.

> I also think it's quite fair to ask people to write their own main.go file to customize esbuild. This would be similar to how rollup.config.js or webpack.config.js works.

This is a reasonable first approach, but I personally think that's too much friction. I have kept the Go and JavaScript APIs similar in shape and feature set so far and I plan to continue to do this with the plugin API. It should be possible to extend esbuild easily from JavaScript as well as Go. The JavaScript bridge will be written using IPC just like the existing JavaScript API.

And having a plugin written in JavaScript will likely lead to higher performance for certain use cases. For example, it will avoid having to repeatedly pay node's startup cost for every transform like in this example.

matthewmueller:
Excellent! I'm thrilled to hear that this is top of mind. 

> This is a reasonable first approach, but I personally think that's too much friction

I absolutely agree that the IPC approach would be more accessible to more folks. I wrote something similar for doing a Go CLI where you can install third-party subcommands. If nothing else, it could be a useful point of reference: https://github.com/matthewmueller/plugin

Excited to see this come together!

PR #246, opened 2020-07-10
Fix broken links (docs/js-api.md)

shcca:


evanw:
Sorry about that. Thanks for the fix.

PR #252, opened 2020-07-12
Attempt wasmURL construction

guybedford:
Most browser bundlers include support for `__dirname` as a Node.js convention as originally implemented by Browserify.

This attempts to automatically construct `wasmURL` using this in the browser if it is defined only.

If shipping an ES module browser build we would be able to rely on `import.meta.url` but because there is no feature detection for this we can't check for it if we cannot assume the module type of the output file.

An alternative could be to provide two browser builds - one CommonJS and one ES module, and then only do the `import.meta.url` case in the ES module build.

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -21,8 +21,14 @@ let transformSync: typeof types.transformSync = (input, options) => {
 
 let startService: typeof types.startService = options => {
   if (!options) throw new Error('Must provide an options object to "startService"');
-  if (!options.wasmURL) throw new Error('Must provide the "wasmURL" option');
-  return fetch(options.wasmURL).then(r => r.arrayBuffer()).then(wasm => {
+	let wasmURL = options.wasmURL;
+  if (!wasmURL) {
+		if (typeof __dirname !== 'undefined')
+			wasmURL = __dirname + '/../esbuild.wasm';
+		else
+			throw new Error('Must provide the "wasmURL" option');
+	}
+  return fetch(wasmURL).then(r => r.arrayBuffer()).then(wasm => {
     let code = `{` +
       `let global={};` +
       `for(let o=self;o;o=Object.getPrototypeOf(o))` +


evanw:
Sorry for not responding to this promptly. I meant to think more about this but never circled back.

Can you say more about how this would work? Specifically, which browser bundlers are you referring to? I just tried both Webpack and Parcel. Webpack appears to always set `__dirname` to `"/"` and Parcel appears to set `__dirname` to the absolute path of the file (so `"/Users/evan/" + ... + "/node_modules/esbuild-wasm/lib"`. Neither of these seem like they would do the right thing in the browser with this PR.

I'm also concerned that this is somewhat magical. Specifically, this could mean things automatically work sometimes but then stop working for non-obvious reasons (e.g. you moved some code into a different folder). With the way it is right now, the user has to explicitly think about configuring `wasmURL` and any URL that fails to load will originate from their code.

I suppose another alternative could be falling back to a globally-available URL from a CDN but printing a warning to the console. You likely wouldn't want to ship code that way (e.g. the CDN may happen to be blocked in certain countries) but it would make this library easier to get started with in development, which seems to be the goal of this PR?

guybedford:
So this actually comes from Browserify originally where it would by default output `/` but it used to be a customizable public directory path at least.

It seems that in Webpack it only ever outputs `/` now but note that would at least send a request for `/esbuild.wasm` which isn't completely stupid I suppose. Parcel seems to have dropped this entirely as a vestige yes.

For what it's worth it would at least get jspm working out of the box - as there is no other way to handle relative asset references as we do with `new URL('./file.wasm', import.meta.url)` in modules for CommonJS. Personally it just seems the natural analog to me. Perhaps if we rebuild on top of this pattern Webpack and Parcel will reappreciate the merits in turn.

guybedford:
To be very specific here, `__dirname` is a feature of the CommonJS module format, and treating `__dirname` as a `/` starting URL is the browserify precedent that Webpack maintained.

Issue #192, opened 2020-06-21
Support SystemJS as a format


chowey:
Any plans for [SystemJS](https://github.com/systemjs/systemjs) as an output format?

I am currently using this for browser support. It seems to be supported by other bundlers, such as Rollup and Webpack. It would be nice to see it in ESBuild.

evanw:
I don't currently have any plans to support SystemJS. The format seems pretty complicated compared to other browser-based formats, especially the way live bindings need to be implemented. All of esbuild's live exports currently just use getters, which is much simpler for me to implement and maintain and which often results in smaller code size. It's currently already possible to target the browser with esbuild using the `iife` and `esm` output formats. Is there a reason those formats don't work for you?

chowey:
The use-case is when there is a library I want to `import` without bundling.

For me in particular, I am using this to load data that the server is generating dynamically. For example, I can do:

```js
import data from "/api/endpoint";

doSomethingWith(data);
```

With SystemJS, it emulates all aspects of ES6, so all I have to do is wrap my `/api/endpoint` as a SystemJS module and the example above acts just like importing a blob of JSON. Dynamically! Before my module even runs! Very useful for me.

I can imagine other reasons someone might want to import an external module that gets resolved at run-time.

I don't know how `iife` would handle importing an external module like this. Would it work the same way? I've solved it in other bundlers using SystemJS.

fazouane-marouane:
My understanding is that you could import Systemjs globally from your html and use `const data = System.import('/api/endpoint');`. So maybe there’s no need to have the bundler do it ?

chowey:
The difference is that `System.import` returns a promise. Using a static `import` does not return a promise. It returns the data right now. The fetching is done ahead-of-time by the module loader, at the same time that the JavaScript is being fetched.

I could also get my data using `const data = fetch('/api/endpoint')` with the same penalty. It returns a promise.

Compare something like this:

```js
import data from "/api/endpoint";

doSomethingWith(data);
doSomethingElseWith(data);
```

to something like this:

```js
fetch('/api/endpoint').then(res => {
  if (!res.ok) {
    return res.text().then(txt => {
      throw new Error(txt);
    });
  }
  return res.json();
}).then(data => {
  doSomethingWith(data);
  doSomethingElseWith(data);
});
```

It kind of gets more convoluted if there are multiple data endpoints involved.

[...]

PR #255, opened 2020-07-13
Very simple SystemJS format support

guybedford:
Implements https://github.com/evanw/esbuild/issues/192.

The benefit of outputting the SystemJS module format is being able to use code splitting workflows in non modules browsers.

This does not include support for live bindings, and will likely have edge cases still to be worked out.

There is repetition between chunk import/export rendering and entry point export rendering. Ideally these separate codepaths could be combined. See for example in RollupJS here how finalizers use their own import/export data structures to render to any format using this interface - https://github.com/rollup/rollup/blob/master/src/finalisers/index.ts#L11. Fully separating the finalizers also avoids cluttering the rendering code with format gating which isn't ideal here.

Very much open to feedback on this PR if you have suggestions for improvements. And let me know if you have any other questions about the format as well.

### DIFF
--- a/internal/printer/printer.go
+++ b/internal/printer/printer.go
@@ -1093,10 +1093,17 @@ func (p *printer) printRequireOrImportExpr(importRecordIndex uint32) {
 	p.printSpaceBeforeIdentifier()
 
 	// Preserve "import()" expressions that don't point inside the bundle
-	if record.SourceIndex == nil && record.Kind == ast.ImportDynamic && p.options.OutputFormat.KeepES6ImportExportSyntax() {
-		p.print("import(")
-		p.print(Quote(record.Path.Text))
-		p.print(")")
+	if record.SourceIndex == nil && record.Kind == ast.ImportDynamic {
+		if p.options.OutputFormat.KeepES6ImportExportSyntax() {
+			p.print("import(")
+			p.print(Quote(record.Path.Text))
+			p.print(")")
+		}
+		if p.options.OutputFormat == config.FormatSystemJS {
+			p.print("__context.import(")
+			p.print(Quote(record.Path.Text))
+			p.print(")")
+		}
 		return
 	}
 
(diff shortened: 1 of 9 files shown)


internal/printer/printer.go
```diff
@@ -1093,10 +1093,17 @@ func (p *printer) printRequireOrImportExpr(importRecordIndex uint32) {
 	p.printSpaceBeforeIdentifier()
 
 	// Preserve "import()" expressions that don't point inside the bundle
-	if record.SourceIndex == nil && record.Kind == ast.ImportDynamic && p.options.OutputFormat.KeepES6ImportExportSyntax() {
-		p.print("import(")
-		p.print(Quote(record.Path.Text))
-		p.print(")")
+	if record.SourceIndex == nil && record.Kind == ast.ImportDynamic {
```
markmav09, line 1096: 👍

shrinktofit:
Any progress on this?

joeljeske:
@evanw if this were updated, would it be accepted? (Asking based on this comment https://github.com/evanw/esbuild/issues/192#issuecomment-650604222) 

SystemJS support is the last requirement I have from being able to adopt esbuild, which I very much would like to do! What are the blockers to supporting SystemJS?

evanw:
> @evanw if this were updated, would it be accepted? (Asking based on this comment https://github.com/evanw/esbuild/issues/192#issuecomment-650604222) 

Not right now. That comment is still accurate. And in addition to that comment, I'm in the middle of a rewrite of most of the linker to address code splitting limitations, support top-level await, and add other features. Supporting yet another output format would also make that rewrite harder.

lifeart:
Up

pmjhonwang:
> > @evanw if this were updated, would it be accepted? (Asking based on this comment [#192 (comment)](https://github.com/evanw/esbuild/issues/192#issuecomment-650604222))
> 
> Not right now. That comment is still accurate. And in addition to that comment, I'm in the middle of a rewrite of most of the linker to address code splitting limitations, support top-level await, and add other features. Supporting yet another output format would also make that rewrite harder.

I’m use single-spa for microfrontend with systemjs, If this format is supported, it would be really appreciated!
Any progress

PR #271, opened 2020-07-18
Update README's «Used By» section to include Kretes

zaiste:
Kretes bundles TypeScript code in development using `esbuild` similarly to Vite.


evanw:
Closing. There is no longer a "used by" section in the readme.

Issue #339, opened 2020-08-20
Incorrectly compiles require() inside constructor call


twiss:
```bash
$ echo "new (require('node-localstorage').LocalStorage)('')" | node
$ echo "new (require('node-localstorage').LocalStorage)('')" | npx esbuild --bundle --platform=node | node
[stdin]:1244
new require_LocalStorage().LocalStorage("");
^

TypeError: require_LocalStorage is not a constructor
    at [stdin]:1244:1
    at Script.runInThisContext (vm.js:116:20)
    at Object.runInThisContext (vm.js:306:38)
    at Object.<anonymous> ([stdin]-wrapper:9:26)
    at Module._compile (internal/modules/cjs/loader.js:959:30)
    at evalScript (internal/process/execution.js:80:25)
    at internal/main/eval_stdin.js:29:5
    at Socket.<anonymous> (internal/process/execution.js:192:5)
    at Socket.emit (events.js:215:7)
    at endReadableNT (_stream_readable.js:1184:12)
```
This should instead output:
```js
new (require_LocalStorage()).LocalStorage("");
```

PR #340, opened 2020-08-21
Fix broken NewExpressions containing require calls

rtsao:
Fixes https://github.com/evanw/esbuild/issues/339 and adds a regression test.

This flag could probably be unset when not strictly necessary, but the extra wrapping parens will at worst be merely redundant.


evanw:
Thank you for sending a fix for this issue!

Issue #187, opened 2020-06-19
Support `exports` syntax in package.json


matthiasg:
More modern ESM node modules are supposed to use `exports` syntax to control visibility of exports. 

https://nodejs.org/api/esm.html#esm_main_entry_point_export

In case this is something that could be done an issue should track any progress

evanw:
I've been following along with discussions about this on the issue trackers for other bundlers. As far as I know the `exports` syntax doesn't interact well with other bundler features that are already widely in use, namely the `module` and `browser` fields in `package.json`. Here is an example thread: https://github.com/webpack/webpack/issues/11014. I think it's premature to add support for this before other bundlers figure this out and the community settles on a way to work around these issues.

matthiasg:
I think that is a prudent approach, but should at least be documented in a section about the module resolution strategy. Node 14 is advocating its use [export takes precendence over main](https://nodejs.org/docs/latest/api/esm.html#esm_package_entry_points) which will lead to some surprises.

developit:
Just wanted to re-up this issue to get it back on your radar. The `"exports"` field is now supported in every major bundler except Parcel and ESbuild. Some examples to look at:

[Webpack](https://webpack.js.org/guides/package-exports)
[Rollup](https://github.com/rollup/plugins/pull/693)
[Vite](https://github.com/vitejs/vite/issues/1418#issuecomment-756349726)
[Snowpack](https://github.com/snowpackjs/snowpack/issues/2203)
[WMR](https://github.com/snowpackjs/snowpack/issues/2203)

evanw:
Thanks for the links. However, it looks like decisions on how this stuff is supposed to work are still settling. See for example [this long Twitter thread](https://twitter.com/FredKSchott/status/1347352979154767873). Big +1 from me to what Devon is advocating for. I'd like to wait until y'all figure it out first so I don't end up implementing something that doesn't follow the community conventions.

developit:
@evanw Shouldn't it be safe to implement Node's algorithm at this point though? It has been shipped as stable for 3 major versions, which makes it pretty locked in at this point. The additional community-proposed fields can certainly be subject to change, but that may always be the case, and they're only ever going to be a layer atop the base 3 fields.

Tobbe:
How about using this [https://github.com/lukeed/resolve.exports](https://github.com/lukeed/resolve.exports)?

lukeed:
[...]

PR #342, opened 2020-08-23
Add ESM build for browsers

calebeby:
Hi @evanw!

I added a ESM build for use in browsers, so that the CJS shim code is not needed for browsers which support ESM natively. I added it to `pkg.module` which is enough for my use case.

As a bonus, I added `pkg.exports`, which is not necessary for my use case, but I thought I might as well. I saw the discussion in [this issue](https://github.com/evanw/esbuild/issues/187) and since then, the discussions between webpack and rollup authors (and others) seem to have stabilized. [Here's a good summary of those discussions](https://github.com/webpack/webpack/issues/11014#issuecomment-641550630) (keep in mind they went with `module` instead of `moduleOnly`). If you decide that you aren't quite ready to have `pkg.exports` here, I am totally fine with removing that from this PR :slightly_smiling_face: 

Thank you for your work with esbuild! I am very excited to see where this project goes in the future, and I am looking forward to new competition in the tooling landscape!


calebeby:
Hey @evanw Could you take a look at this?

evanw:
Thanks for the ping. I can definitely ship something like this.

The export logic isn't quite right though since you don't want bundlers that support modules picking up `browser.mjs` for non-browser targets, so I'm going to change how exports are configured before shipping it.

evanw:
Changes I made:

* I tested both builds in the browser and fixed everything that caused it to be broken (e.g. no `ESBUILD_VERSION`).

* I changed the target `es2020` to `es2017` because otherwise the use of certain newer features such as optional catch bindings would unnecessarily break in older browsers that still support esm syntax.

* I removed `exports` in `package.json` because as far as I know, adding it is a backwards-incompatible change, and I don't want this to break backwards compatibility.

* I removed `module` because the `esbuild-wasm` package can also be used in node, and bundling it with a module-aware bundler shouldn't cause the browser build to be bundled.

    Technically the way to do this is to keep the `module` field but make it a node esm build, then add a `browser` mapping that overrides both `main` and `module`. However, the node code relies on `__dirname` for which the replacement is `import.meta.url` which is new and not supported everywhere, so I backed out of that change.

* I moved the `lib/browser.mjs` file to the `esm` folder instead to avoid potential issues due to adjacent files with the same name but different extensions.

* I renamed the file to `esm/browser.js` because some web servers serve `.mjs` files with `application/octet-stream` which causes the browser to refuse to parse the JavaScript code.

* I removed the unnecessary `default` export.

So basically the only change in the next release will be a new `esm/browser.js` file, without any changes to `package.json` at the moment.

Issue #109, opened 2020-05-13
Support `--format=` when without bundle


CompuIves:
First of all, this is an incredible project. I'm really impressed with the speed!

I'm trying to run `esbuild --format=cjs add.js` on 

```
import createMathOperation from './_createMathOperation.js';

var add = createMathOperation(function(augend, addend) {
  return augend + addend;
}, 0);

export default add;
```

But the output is 

```
import createMathOperation2 from "./_createMathOperation.js";
var add = createMathOperation2(function(augend, addend) {
  return augend + addend;
}, 0);
export default add;
```

Is this intended? Or should it only work with `--bundle`?

My use case is that I'm looking for a quick transformer that can transform an esmodule to commonjs, I'm wondering if esbuild could do something like this.

evanw:
It's a good idea. I've been planning to do this but haven't gotten to it yet. It should be fairly straightforward. However, the whole bundling transform is currently being rewritten to support tree shaking and code splitting. So I'm going to wait to do this until after that is done.

WenheLI:
Any updates on the `cjs` output? 
I also want to output a project in `cjs` format.

evanw:
I just landed a change that implements this. There are some details in [the release notes](https://github.com/evanw/esbuild/blob/8b76046a9c9aeb5d5d0da75935fac380f978ff1d/CHANGELOG.md#unreleased).

I decided to not implement CommonJS to ECMAScript module conversion for now. There are a few different options and it's not immediately clear to me how to implement it. I'm not sure if there's a single way to implement it that would satisfy all use cases, or even what the majority use case is. Some of my open questions:

* What environments and situations is this useful in? I think perhaps the browser is one of them, but this seems to me like something that would be pretty custom per app and wouldn't really be appropriate to build into esbuild's core.

* Naive conversion of `require()` to `import` is imprecise because it potentially changes the execution order, which could introduce bugs. How much do people care about these bugs? What do people usually do to fix them? For example, this could be worked around by translating CommonJS to an ECMAScript module that exports a "require function" to invoke to call it. But that likely wouldn't work well with existing tooling.

[...]

PR #350, opened 2020-08-26
feat: support format option for transform

sastan:
This allows to use `transform` to output a different format without
going through `build` which would resolve all imports.

Import handling:

- `cjs`: imports are not resolved and re-written as `require` calls
- `iife`: any import will throw

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -259,7 +259,7 @@ func parseFile(args parseArgs) {
 	if args.options.IsBundling {
 		result.resolveResults = make([]*resolver.ResolveResult, len(result.file.ast.ImportRecords))
 
-		if len(result.file.ast.ImportRecords) > 0 {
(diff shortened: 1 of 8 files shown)


nathanforce:
Love to see this! Is there any way I can test this out against my project locally?

sastan:
@nathanforce I have no published package. You need to check out the branch and use `make platform-all` to have a bundle.

sastan:
@evanw  Can i help with anything to get this PR merged? Very much open to feedback on this PR if you have suggestions for improvements.

evanw:
Thank you for putting this PR together. However, this is sufficiently complicated and enough of a core part of esbuild that I don't think implementing this as an external contribution makes sense. Some examples of things that aren't handled here:

* Making this work the same way with transforming and with building without bundling
* Making sure transforming doesn't touch the file system
* Handling lazy exports from various non-JavaScript loaders

I have also already started work on implementing #109 and am most of the way there. I will make sure this works as a part of the work for #109.

evanw:
Closing this as #109 has just been implemented.

Issue #345, opened 2020-08-24
Tree shaking when concatenating strings?


eigilsagafos:
Given the following structure:

src/lib/fragment.js
```js
export const fragment = "fragment"
```
src/lib/stringA.js
```js
import { fragment } from "./fragment"
export const StringA = "String A with fragment: " + fragment
```
src/lib/stringB.js
```js
export const StringB = "String B without fragment"
```
src/lib/stringC.js
```js
export const StringC = "String C without fragment"
```
src/lib/stringC.js
```js
export const StringC = "String C without fragment"
```
src/lib/index.js
```js
export { StringA } from "./stringA"
export { StringB } from "./stringB"
export { StringC } from './stringC'
```

src/index.js
```js
import { StringB } from "./lib"
console.log(StringB)
```

Running `esbuild src/index.js --bundle --platform=node` produces the following:
```js
// src/lib/fragment.js
const fragment = "fragment";

// src/lib/stringA.js
const StringA = "String A with fragment " + fragment;

// src/lib/stringB.js
const StringB = "String B without fragment";

// src/lib/stringC.js

// src/lib/index.js

// src/index.js
console.log(StringB);
```

I would expect `fagment` and `stringA` to be tree-shaken. If I make StringA a function it is. Is this expected? Any way around this?

eigilsagafos:
Thanks @rtsao for taking a go at this. Any thoughts about this @evanw?

evanw:
This is tricky because esbuild doesn't yet have support for this kind of complex cross-file analysis. This would indeed be something to aspire to but as #351 shows, doing this safely isn't easy. This would also need to check that the transformation doesn't change behavior due to the TDZ (temporal dead zone) of any reference involved, which is also analysis that esbuild doesn't do yet. Since this seems like a big step up in complexity, I'm not sure yet if it makes sense to include this kind of thing in esbuild. It might turn out to be too complex and hard to maintain. This requires more investigation.

I tried to come up with a way around this by using a function with a `/* @__PURE__ */` comment. I have adopted the convention used by other tools (e.g. Rollup, UglifyJS) that a function call preceded by a `/* @__PURE__ */` comment is considered to be removable by tree shaking if the result is unused. Unfortunately the convention only applies to call expressions and there's no way to annotate any other kind of expression. Still, you could in theory do this so that tree shaking could remove `StringA` if it's not used:

```js
function concat(a, b) {
  return a + b
}

export const fragment = "fragment"
[...]

PR #351, opened 2020-08-27
Add tree shaking of unused pure binary expressions

rtsao:
Fixes https://github.com/evanw/esbuild/issues/345

- Adds logic to recursively check if the left and right sub-expressions of a binary expression would be able to pruned. If they both are, then the binary expression itself can also be pruned.
- Adds a regression test.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -8910,6 +8910,9 @@ func (p *parser) exprCanBeRemovedIfUnused(expr ast.Expr) bool {
 		*ast.EString, *ast.EThis, *ast.ERegExp, *ast.EFunction, *ast.EArrow, *ast.EImportMeta:
 		return true
 
+	case *ast.EBinary:
+		return p.exprCanBeRemovedIfUnused(e.Left) && p.exprCanBeRemovedIfUnused(e.Right)
+
 	case *ast.EDot:
 		return e.CanBeRemovedIfUnused
 
--- a/internal/bundler/bundler_dce_test.go
+++ b/internal/bundler/bundler_dce_test.go
@@ -662,3 +662,23 @@ func TestTreeShakingReactElements(t *testing.T) {
 		},
 	})
 }
+
+func TestTreeShakingBinaryExpressions(t *testing.T) {
+	dce_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.js": `
+				let a = "a";
+				let b = a + "b";
+
+				let c = "c";
+				let d = c + "d";
+				console.log(d);
+			`,
+		},
+		entryPaths: []string{"/entry.js"},
+		options: config.Options{
+			IsBundling:    true,
+			AbsOutputFile: "/out.js",
+		},
+	})
+}
--- a/internal/bundler/bundler_dce_test.go
+++ b/internal/bundler/bundler_dce_test.go
@@ -673,6 +673,14 @@ func TestTreeShakingBinaryExpressions(t *testing.T) {
 				let c = "c";
 				let d = c + "d";
 				console.log(d);
+
+				let e = {
+					toString() {
+						console.log("effect");
+						return "e";
+					}
+				}
+				let noprune = "str" + e;
 			`,
 		},
 		entryPaths: []string{"/entry.js"},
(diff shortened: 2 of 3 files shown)


internal/parser/parser.go
```diff
@@ -8910,6 +8910,9 @@ func (p *parser) exprCanBeRemovedIfUnused(expr ast.Expr) bool {
 		*ast.EString, *ast.EThis, *ast.ERegExp, *ast.EFunction, *ast.EArrow, *ast.EImportMeta:
 		return true
 
+	case *ast.EBinary:
```
rtsao, line 8913: I'm not very familiar with Go, but I would assume order might have an effect on performance in terms of relative frequency of these nodes in an average AST.

I just added this to the top of the non-trivial special cases based on alphabetical order, but this might be something to consider.

rtsao: Speaking of performance, obviously any extra minification optimization doesn't come for free. I have not yet done any profiling, but I noticed there is already a similar optimizations for array literals, so I figured this was probably just an oversight rather than an intentional omission.

rtsao:
Oops, I forgot `+` can cause type coercion that would potentially cause side effects.

For example:
```js
let a = {
  toString() {
    console.log("hi");
    return "a";
  },
};

let b = a + "b";
```

So properly handling this will require a lot more work to identify cases that this will be possible. I am now noticing `isPrimitiveToReorder`, `toNumberWithoutSideEffects`, and `toBooleanWithoutSideEffects`, but I think scope analysis would have to be involved for this to be useful.

rtsao:
It looks like [Rollup fails to account for this](https://rollupjs.org/repl/?version=2.26.6&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmxldCUyMGElMjAlM0QlMjAlNUMlMjJhJTVDJTIyJTNCJTVDbmxldCUyMGIlMjAlM0QlMjBhJTIwJTJCJTIwJTVDJTIyYiU1QyUyMiUzQiU1Q24lNUNubGV0JTIwYyUyMCUzRCUyMCU1QyUyMmMlNUMlMjIlM0IlNUNubGV0JTIwZCUyMCUzRCUyMGMlMjAlMkIlMjAlNUMlMjJkJTVDJTIyJTNCJTVDbmNvbnNvbGUubG9nKGQpJTNCJTVDbiU1Q25sZXQlMjBlJTIwJTNEJTIwJTdCJTVDbiUyMCUyMHRvU3RyaW5nKCklMjAlN0IlNUNuJTIwJTIwJTIwJTIwY29uc29sZS5sb2coJTVDJTIyZWZmZWN0JTVDJTIyKSUzQiU1Q24lMjAlMjAlMjAlMjByZXR1cm4lMjAlNUMlMjJlJTVDJTIyJTNCJTVDbiUyMCUyMCU3RCUyQyU1Q24lN0QlM0IlNUNubGV0JTIwbm9wcnVuZSUyMCUzRCUyMCU1QyUyMnN0ciU1QyUyMiUyMCUyQiUyMGUlM0IlNUNuJTVDbiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=) and *incorrectly* prunes binary expressions with side effects:

```js
let a = "a";
let b = a + "b";

let c = "c";
let d = c + "d";
console.log(d);

let e = {
  toString() {
    console.log("effect");
    return "e";
  },
};
let noprune = "str" + e;
```
```js
let c = "c";
let d = c + "d";
console.log(d);
```

rtsao:
Closing for now. Will open another PR if there's a good way of handling this

Issue #372, opened 2020-09-08
Breaks async template literal tags


frandiox:
Hi! First of all, thanks for this amazing tool!
There might be an issue with async template literals tags (such as https://github.com/porsager/postgres):

```js
const result = await sql`select * from table`
```

is built as:

```js
const result = (await sql)`select * from table`;
```

The parenthesis basically break the code and the whole thing just returns a `Promise`. instead of the actual result.

I've tried v0.6.30 and v0.6.32. I'm using the Rollup plugin https://github.com/egoist/rollup-plugin-esbuild with TS target `es2018` -- same happens for `es2016` but with `yield` instead of `await`.

**Edit**

This works:
```js
const result = await (sql`select * from table`)
```
builds as:

```js
const result = await sql`select * from table`;
```

However, formatters such as Prettier will likely remove those parenthesis before esbuild.

evanw:
Thanks for reporting this! It looks like the precedence of template literal tags is incorrect. I just fixed the bug and I will do a release later today.

frandiox:
@evanw Thank you so much!

evanw:
The fix was just released in version 0.6.33.

PR #376, opened 2020-09-08
Fix parsing of awaited tagged template expressions

rtsao:
Fixes https://github.com/evanw/esbuild/issues/372

Before this, the parser would incorrectly parse `` await foo`bar` `` as `Template(Await(...))` instead of `Await(Template(...))`

I assume `LPrefix` was used before because of the location of `await` in the [precedence table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table) (17), but it seems this doesn't match up with the observed behavior of other parsers.

In addition to tagged templates (with and without substitutions), I added test cases for `await` with both the lowest (comma) and highest (grouping and property access) operators.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -2093,7 +2093,7 @@ func (p *parser) parsePrefix(level ast.L, errors *deferredErrors, flags exprFlag
 					if p.fnOptsParse.arrowArgErrors != nil {
 						p.fnOptsParse.arrowArgErrors.invalidExprAwait = nameRange
 					}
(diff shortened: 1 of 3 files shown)


evanw:
Unfortunately I don't believe this is a valid fix for this issue. This change means `await 1, 2` would be incorrectly parsed as `await (1, 2)` while it should be parsed as `(await 1), 2` instead. This happens because changing `LPrefix` to `LLowest` here means "continue parsing all operators of any precedence after the await token".

PR #377, opened 2020-09-09
Fix parsing of type declaration edge case

rtsao:
```ts
type
Foo = {}

// should be transformed to:

type;
Foo = {};
```
- https://www.typescriptlang.org/play?ts=4.0.2#code/C4TwDgpgUAYg9nABAXkQbwL5SA
- https://github.com/microsoft/TypeScript/blob/ee5f51bc0fc06964f2ba27217b5555a82a398673/tests/baselines/reference/asiPreventsParsingAsTypeAlias01.js
- https://github.com/microsoft/TypeScript/blob/ee5f51bc0fc06964f2ba27217b5555a82a398673/tests/baselines/reference/asiPreventsParsingAsTypeAlias02.js

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -5325,7 +5325,7 @@ func (p *parser) parseStmt(opts parseStmtOpts) ast.Stmt {
 				if p.TS.Parse {
 					switch name {
 					case "type":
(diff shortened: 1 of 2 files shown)


evanw:
Good catch! Thanks for the fix, and for adding test coverage.

PR #383, opened 2020-09-12
Mark test helpers

rtsao:
Adds [`T.Helper`](https://golang.org/pkg/testing/#T.Helper) calls to test helper functions.

By doing so, these test helpers will be skipped when printing the file and line number on test errors, thereby producing a more useful error message:

```diff
 ~/code/esbuild/internal/parser
 ❯ go test ../parser/...
 --- FAIL: TestSomeBrokenTest (0.00s)
     --- FAIL: TestSomeBrokenTest/foo (0.00s)
-        util.go:12: foo;
+        parser_test.go:692: foo;
              != bar
 FAIL
 FAIL   github.com/evanw/esbuild/internal/parser        1.470s
 FAIL
```

After this change, it is possible to simply click the filename and line number from the terminal output and be taken to the broken assertion (assuming you are in the correct directory). Otherwise, it would merely show the file and line of the test helper itself which isn't particularly useful.

I've found this change to be very convenient when doing TDD.

### DIFF
--- a/internal/bundler/bundler_test.go
+++ b/internal/bundler/bundler_test.go
@@ -31,6 +31,7 @@ func es(version int) compat.Feature {
 }
 
 func assertEqual(t *testing.T, a interface{}, b interface{}) {
+	t.Helper()
 	if a != b {
 		stringA := fmt.Sprintf("%v", a)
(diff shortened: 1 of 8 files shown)


evanw:
Looks good. Thanks!

Issue #79, opened 2020-04-29
Flow support


jankuca:
Hello, are you also considering Flow support or will this only support TypeScript?

evanw:
I'm not currently considering Flow support. I've never used Flow myself and I had the (perhaps incorrect) impression that the web development community was gradually moving away from Flow to TypeScript. My goal for esbuild is to build a bundler that solves for a sweet spot of common use cases, not to build a bundler that works for any codebase.

That said, I could consider it. To get an idea of the scope of work that this would require (since I'm completely unfamiliar with Flow):

* Does Flow ever use types to change how code is emitted? Or is it possible to convert Flow to JavaScript just by "deleting the types"?
* Does flow add any extra syntax to JavaScript like TypeScript does (`enum` and `namespace`) that would need to be converted to JavaScript?
* Is there a good comprehensive public suite of test files that, if they are parsed correctly, could give reasonable confidence that the parser will work on real-world code?

rtsao:
1. Flow is purely type annotations for JavaScript, so conversion to JS merely involves removing the annotations. This is what the Flow Babel plugin does.
2. Flow adds special syntax for type annotations, but does not introduce any additional language features that need to be converted to JS. As stated before, you just need to remove the annotations.
3. There might be a test suite somewhere, but at the very least, there are number of large projects using Flow (Parcel for example) that might be good candidates.

jankuca:
As @rtsao says, Flow does not add any new language constructs. All Flow syntax can simply be stripped to get working JS code.

The Flow project contains multiple packages, one of which is [the `flow-remove-types` package](https://github.com/facebook/flow/tree/master/packages/flow-remove-types) with [a test file](https://github.com/facebook/flow/blob/master/packages/flow-remove-types/test/source.js) featuring all the added syntax and [multiple possible outputs](https://github.com/facebook/flow/tree/master/packages/flow-remove-types/test) which are valid JS.

Please look at this material to see if it can help. The package itself (published on NPM as `flow-remove-types`) could be used directly. (It is not written in Go though.)

evanw:
Thanks for the pointers! That's very helpful. I'm currently focused on building some other features first (code splitting and tree shaking) but I would like to do this at some point.

murrayju:
I'd like to see this as well. The flow syntax is extremely similar to TypeScript, really just a handful of differences. So hopefully this wouldn't be very hard to add.

rtsao:
[...]

PR #384, opened 2020-09-12
WIP: Add Flow syntax parsing

rtsao:
Work-in-progress implementation of Flow syntax support https://github.com/evanw/esbuild/issues/79


internal/parser/parser.go
```diff
@@ -5083,6 +5083,54 @@ func (p *parser) parseStmt(opts parseStmtOpts) ast.Stmt {
...
+
+				default:
+					p.lexer.Unexpected()
+				}
+			}
+
+			p.lexer.ExpectContextualKeyword("from")
+			p.parsePath()
+			p.lexer.ExpectOrInsertSemicolon()
+			return ast.Stmt{Loc: loc, Data: &ast.STypeScript{}}
```
rtsao, line 5132: This is copied from the TypeScript-specific section of the import parser. Should there be a separate AST type for Flow types? Or should should this just emit a blank statement?

evanw: I think `STypeScript` should be changed to something more generic such as `SIgnoredType`. The only purpose is to tell the caller to drop the statement. These statements never make it out of the parser.

evanw:
I'm closing this since #79 is now closed. The context is that the Flow team recently announced that they are going to diverge from the JavaScript language and no longer be just JavaScript with types: https://medium.com/flow-type/clarity-on-flows-direction-and-open-source-engagement-e721a4eb4d8b. So Flow support now seems better-suited for an esbuild plugin than being built into esbuild itself.

Issue #388, opened 2020-09-16
Incorrect async transform


breeswish:
```js
// esbuild --bundle index.ts --outfile=out.js --target=es2015

const myFunc = function (c) {
  return {
    foo: async (message, options = {}) => {
      if (message.a !== "1") {
        throw new Error("foo");
      }
      console.log("Pass!");
    },
  };
};

myFunc({}).foo({ a: "1" });
```

The snippet above will throw an error instead of printing "Pass" after esbuild transformation.

The transformed result is:

```js
(() => {
  var __async = (__this, __arguments, generator) => {
    return new Promise((resolve, reject) => {
      var fulfilled = (value) => {
        try {
          step(generator.next(value));
        } catch (e) {
          reject(e);
        }
      };
      var rejected = (value) => {
        try {
          step(generator.throw(value));
        } catch (e) {
          reject(e);
        }
      };
      var step = (result) => {
        return result.done ? resolve(result.value) : Promise.resolve(result.value).then(fulfilled, rejected);
      };
      step((generator = generator.apply(__this, __arguments)).next());
    });
  };

  // src/index.ts
  const myFunc = function(c) {
    return {
      foo: (_0) => __async(this, arguments, function* (message, options = {}) {
        //                       ^^^^^^^^^
        // Notice that `arguments` here is incorrect, since arrow function does
        // not provide arguments.
        if (message.a !== "1") {
          throw new Error("foo");
        }
        console.log("Pass!");
      })
    };
  };
  myFunc({}).foo({a: "1"});
})();
```

rtsao:
It looks like this regression was introduced in `v0.6.32` via https://github.com/evanw/esbuild/commit/d121b389256b50b0ebc472a21c0679e3ca9a41d7

Which makes sense because removing `options = {}` from your example fixes the issue.

PR #390, opened 2020-09-16
Fix broken edge case of async function lowering

rtsao:
Fixes https://github.com/evanw/esbuild/issues/388 and adds regression test.

Also adds an optimization to skip argument forwarding altogether when it is trivial to determine that all default parameter expressions cannot throw.

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -8987,7 +8987,7 @@ func (p *parser) visitExprInOut(expr js_ast.Expr, in exprIn) (js_ast.Expr, exprO
 		p.pushScopeForVisitPass(js_ast.ScopeFunctionBody, e.Body.Loc)
 		e.Body.Stmts = p.visitStmtsAndPrependTempRefs(e.Body.Stmts)
 		p.popScope()
-		p.lowerFunction(&e.IsAsync, &e.Args, e.Body.Loc, &e.Body.Stmts, &e.PreferExpr, &e.HasRestArg)
+		p.lowerFunction(&e.IsAsync, &e.Args, e.Body.Loc, &e.Body.Stmts, &e.PreferExpr, &e.HasRestArg, true)
 		p.popScope()
 
 		if p.MangleSyntax && len(e.Body.Stmts) == 1 {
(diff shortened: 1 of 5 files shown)


internal/js_parser/js_parser_lower.go
```diff
@@ -268,35 +268,26 @@ func (p *parser) lowerFunction(
 					*args = append(*args, js_ast.Arg{Binding: js_ast.Binding{Loc: arg.Binding.Loc, Data: &js_ast.BIdentifier{Ref: argRef}}})
 				}
 
-				// Forward all arguments from the outer function to the inner function
-				if p.fnOnlyDataVisit.argumentsRef != nil {
```
rtsao, line 272: The problem seems to be `argumentsRef` is non-nil if the async arrow function is nested inside a regular function. This code path only triggers when the parameters have a default value (and hence `couldThrowErrors = true`), so both conditions are required for this bug.

internal/js_parser/js_parser.go
```diff
@@ -8987,7 +8987,7 @@ func (p *parser) visitExprInOut(expr js_ast.Expr, in exprIn) (js_ast.Expr, exprO
 		p.pushScopeForVisitPass(js_ast.ScopeFunctionBody, e.Body.Loc)
 		e.Body.Stmts = p.visitStmtsAndPrependTempRefs(e.Body.Stmts)
 		p.popScope()
-		p.lowerFunction(&e.IsAsync, &e.Args, e.Body.Loc, &e.Body.Stmts, &e.PreferExpr, &e.HasRestArg)
+		p.lowerFunction(&e.IsAsync, &e.Args, e.Body.Loc, &e.Body.Stmts, &e.PreferExpr, &e.HasRestArg, true)
```
rtsao, line 8990: I suppose adding something like `&e.isArrow` might be more readable. But I figured omitting such a property on `Fn` was intentional?

scripts/end-to-end-tests.js
```diff
@@ -1032,6 +1032,20 @@ in.js:24:30: warning: Writing to getter-only property "#getter" will throw
         )
       `,
     }, { async: true }),
+    test(['in.js', '--outfile=node.js', '--target=es6'],
+    {
+      'in.js': `
+        function nonArrowWrapper() {
+          return async (x, paramWithDefault = {}) => {
```
rtsao, line 1039: Some expressions can be statically determined to never throw. Of these, it might be worth handling the subset that are both fairly common and trivial to identify (e.g. `false`, `true`, etc.)

internal/js_parser/js_parser.go
evanw: Yes, that's correct. This change sounds good.

internal/js_parser/js_parser_lower.go
evanw: Thanks for the summary.

evanw (approved):
This looks good. Thanks for fixing this.

PR #392, opened 2020-09-16
Introduce module dependency analysis without compiling the output bundle

### LINKED ISSUE #386: Using esbuild for dependency analysis
I maintain a big project consisting of more than a thousand modules and using `RequireJS` for modularisation. Analysing module dependencies for documentation and refactoring purposes becomes painfully slow. I am thinking if `esbuild` could not help me just like it helps to speed up minification and concatenation or module bundles.

`esbuild` "almost" does what I would need by producing a metafile (in particular the `imports` property) together with the output bundle. But I do not need the bundle :-)

For example, the public `api` and `cli` packages could support _Analyse_ in addition to _Build_ and _Transform_. Analyse would traverse module dependencies like Build, but produce only the module metadata (`Metafile`). It would call only `bundler.ScanBundle`, but not `bundler.Compile`.

What do you think?

Being able to build a module dependency analyser using the `esbuild` API would be a half of the task. The second half would be making `esbuild` understand `AMD.JS` module syntax. It is an off-topic in this issue, but did you consider recognising the AMD.JS module syntax?

Thank you!

prantlf:
Adds an `analyse` execution mode, modelled after building a bundle, but ending after `bundler.ScanBundle` and writing the metadata out. Attempts to fix #386.

* Add the command-line flag `--analyse` to compute metadata of a bundle.
* Add the `Analyse` method to the API, which produces metadata after `bundler.ScanBundle`.
* Add the `analyse` command to the service, available only for node.
* Include only "inputs" in metadata from the module analysis.

Remarks:

* The method `parseOptionsImpl` becomes less legible with the conditions checking `buildOpts`, `transformOpts` and now `analyseOpts`.
* Many options between `esbuild --bundle` and `esbuild --analyse` are shared, which becomes visible in the copied code between methods `flagsForBuildOptions` and `flagsForAnalyseOptions` and structures `BuildOptions` and `AnalyseOptions`.
* Some flags from `AnalyseOptions` are probably not necessary. I was not sure, which flags can affect parsing and module resolving. I added all, that I suspected at least a little.

### DIFF
--- a/cmd/esbuild/main.go
+++ b/cmd/esbuild/main.go
@@ -18,6 +18,7 @@ Usage:
   esbuild [options] [entry points]
 
 Options:
+  --analyse             Collect metadata about the source modules
   --bundle              Bundle all dependencies into the output files
   --define:K=V          Substitute K with V while parsing
   --external:M          Exclude module M from the bundle
--- a/cmd/esbuild/service.go
+++ b/cmd/esbuild/service.go
@@ -242,6 +242,11 @@ func (service *serviceType) handleIncomingPacket(bytes []byte) (result outgoingP
 				refCount: refCount,
 			}
 
+		case "analyse":
+			return outgoingPacket{
+				bytes: service.handleAnalyseRequest(p.id, request),
+			}
+
 		case "error":
 			// This just exists so that errors during JavaScript API setup get printed
 			// nicely to the console. This matters if the JavaScript API setup code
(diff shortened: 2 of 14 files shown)


divy-work:
This would be really helpful!

ka808 (approved):

evanw:
I'm closing this for the reasons state in #386. I consider this to be out of scope and the increase in API surface isn't justified IMO given how similar this is to the regular build API.

PR #396, opened 2020-09-19
fix build failed after upgrade golang.org/x/sys

akayj:
After upgrading `golang.org/x/sys` to latest version,  `v0.0.0-20200918174421-af09f7315aff` for example,
build will failed, because of [this](https://github.com/golang/sys/commit/6fcdbc0bbc04dcd9e7dc145879ceaf9bf1c6ff03)
__unix.SYS_IOCTL__ was removed,  using `syscall.SYS_IOCTL` instead.

### DIFF
--- a/internal/logger/logger_darwin.go
+++ b/internal/logger/logger_darwin.go
@@ -5,6 +5,7 @@ package logger
 import (
 	"os"
 	"unsafe"
+	"syscall"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


evanw (approved):
Thanks for fixing this.

evanw:
I ended up switching to `unix.IoctlGetWinsize` instead since it seemed more appropriate and higher-level.

PR #432, opened 2020-10-05
Update Readme.md

MjCode01:
corrected grammar.


evanw:
The original grammar is correct.

PR #450, opened 2020-10-12
Fix sourcemap path problem

### LINKED ISSUE #444: sourcemap path problem
Thanks to the author for amazing work.

I will try to use `esbuild` in [Vetur](https://github.com/vuejs/vetur) project https://github.com/vuejs/vetur/pull/2301.
This project have a lot of dependencies that make it hard to bundle and minify.
The `esbuild` is 128x faster than `rollup`. 
The cost time is from `6m 24.9s` to `3s`.

The VSCode debugger is broken with `esbuild` sourcemap.
I find two problems with sourcemap.
- The sourcemap `sources` path is base on esbuild execution path.
To be precise, it should be based on outFile path.
I need to hack `process.cwd()` for pin it.
https://github.com/vuejs/vetur/blob/2dd9fa83678a1f40cf78d4f1a13756eef78bd1e2/build/rollup-plugins.js#L30.
- The sourcemap don't have `sourceRoot`.
vscode debugger can't resolve path.
so I need to overrides to absoulte path.
https://github.com/vuejs/vetur/blob/2dd9fa83678a1f40cf78d4f1a13756eef78bd1e2/.vscode/launch.json#L32.

If we're fixing the first problem, it seems the second one isn't necessary.

yoyo930021:
Fixed #444 

This is my first time writing about golang.
I'm sorry if you have any questions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,7 @@
 [*]
 indent_style = tab
 indent_size = 2
+
+[*.js]
+indent_style = space
+indent_size = 2
--- a/internal/bundler/linker.go
+++ b/internal/bundler/linker.go
@@ -3442,7 +3442,7 @@ func (repr *chunkReprJS) generate(c *linkerContext, chunk *chunkInfo) func([]ast
 		}
 
 		if c.options.SourceMap != config.SourceMapNone {
-			sourceMap := c.generateSourceMapForChunk(compileResultsForSourceMap)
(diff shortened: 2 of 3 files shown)


evanw:
Thanks for the pull request. I agree that this should be fixed, although I don't think this is the best way to do it. Or at least the real situation is more complicated.

With the introduction of plugins there are also going to be modules that don't correspond to anything on disk at all. Attempting to interpret those module names as file system paths might lead to pretty confusing and incorrect results.

So I'm going to land this now but I may change this and/or rewrite it in the future to better support additional use cases.

PR #457, opened 2020-10-13
Pass config.Options always by reference

prantlf:
* Make the options usage consistent throughout the source code.
* Allow hosting objects in the options, which must not be cloned, like mutexes.

When inspecting all usages of `config.Options`, I was surprised seeing some functins accepting and some strcutures nesting `*config.Options`, while the others `config.Options.`. Was there a reason behind that? Did you expect that some functions should be able to modify the options and those should not affect the caller's options?

If not, I am offering this change to pass around and nest the `config.Options` object always using a pointer to prevent the cloning. What do you think?

(I stumbled over this, when I was prototyping support for AMD.JS. I added a map to `config.Options.AMD` and included a mutex there too, to be able to synchronize accesses to the map. Locking did not work and running `go vet` discovered that `config.Options` were passed around by value, which must not happen if a mutex is there.)

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -121,7 +121,7 @@ type parseArgs struct {
 	importSource    *logger.Source
 	flags           parseFlags
 	importPathRange logger.Range
-	options         config.Options
+	options         *config.Options
 	results         chan parseResult
 
 	// If non-empty, this provides a fallback directory to resolve imports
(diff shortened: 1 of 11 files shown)


prantlf:
Oops, I noticed too late that some tests failed. Does it make sense to look at them, or is the passing by value as-designed and I should withdraw this change?

evanw:
Yes, passing by value is intentional. The options may be mutated at various points to override values. Some examples:

https://github.com/evanw/esbuild/blob/eb43977e73410fc4a01a52034fd49744557a918d/internal/bundler/bundler.go#L506-L508

https://github.com/evanw/esbuild/blob/eb43977e73410fc4a01a52034fd49744557a918d/internal/bundler/bundler.go#L551-L555

prantlf:
Oh, you're right. I should've looked for such examples earlier. Thank you!

PR #470, opened 2020-10-18
Prebuild package for armv6l and above

0x11-dev:
Hi @evanw!

I add prebuild scripts for armv6l and above architectures.  

Node.js use `arm` to specify ARM 32-bit hardware. So map `arm` to`GOARCH=arm GOARM=6` in this PR to support as much device as possible.  

Since Raspberry Pis have ARMv6Z CPU, this PR would help users on these platform.

docs:
- https://github.com/golang/go/wiki/GoArm
- https://en.wikipedia.org/wiki/Raspberry_Pi#Specifications

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest, self-hosted]
 
     steps:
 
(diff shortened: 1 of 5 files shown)


evanw:
Caveat: I'm not at all familiar with the ARM ecosystem.

First, thank you for removing the `armv7l` target. I assume it's not necessary given the `armv6l` target. Some questions:

* Am I correct in assuming that this PR is about 32-bit ARM support?
* Why not target `GOARM=5` instead?
* Have you tried the generated `esbuild` binary and confirmed that it actually works on the target architecture?
* I'm thinking of changing `esbuild-linux-armv6l` to just `esbuild-linux-arm` so there's a single package for 32-bit ARM. Any concerns about that?
* Also, could you remove the binary itself from your commit? Build artifacts should not be checked in to version control.

evanw:
I believe this PR is no longer relevant now that #528 has been resolved.

PR #493, opened 2020-10-29
docs: adding webpack@5 with filesystem cache

alexander-akait:
webpack@5 supports cache and I think it's worth show this, I have not found a script to generate image.

Time around 2-3 seconds on my computer.

Maybe other bundlers have cache too. But I don't found something for parcel, for rollup - https://github.com/rollup/rollup/issues/2182

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,7 @@ require/webpack/node_modules:
 require/webpack5/node_modules:
 	mkdir -p require/webpack5
 	echo '{}' > require/webpack5/package.json
-	cd require/webpack5 && npm install webpack@5.0.0-rc.4 webpack-cli@4.0.0-rc.1 ts-loader@8.0.4 typescript@4.0.3
+	cd require/webpack5 && npm install webpack@5.3.1 webpack-cli@4.1.0 ts-loader@8.0.7 typescript@4.0.3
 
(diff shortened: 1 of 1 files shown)


evanw:
Yes, it's true that this benchmark shows worst-case bundling time. However, I'd like to focus this benchmark on worst-case bundling time because it's the hardest problem to solve and because I don't think it's getting enough attention from the community.

Most people in the community focus on incremental build time instead. I think the advocacy for that is already covered elsewhere. It's pretty easy to add a cache and have good incremental build time, so I don't think it's a good metric to use for a benchmark that's meant to encourage the community to improve performance. If anything, the focus on incremental build time could be part of the reason why our tools are so slow. It's somewhat of a crutch that hides the real performance cost.

As for why there isn't a deeper breakdown: There are many different dimensions to look at bundling time. For example, this benchmark involves generating source maps, and of course not generating them can speed things up. However, that's a slippery slope and I don't want this benchmark to have 30 different bars for various combinations of parameters because it makes it harder to read and dilutes the message. (This benchmark contains source maps because I believe they are an important part of running code in production, for error traces)

evanw:
I'm closing this PR for the reasons stated above.

Issue #467, opened 2020-10-17
esbuild error on WSL (Ubuntu 20.04.1 LTS - Windows 10) / via yarn 2


rizkyramadhan:
After creating fresh yarn 2 project.  install esbuild as usual, and then run it:

```
/home/riz/Projects/test-esbuild/.yarn/unplugged/esbuild-npm-0.7.16-564ecde0e0/node_modules/esbuild/bin/esbuild:1
ELF
^

SyntaxError: Invalid or unexpected token
    at wrapSafe (internal/modules/cjs/loader.js:979:16)
    at Module._compile (internal/modules/cjs/loader.js:1027:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.external_module_.Module._load (/home/riz/Projects/test-esbuild/.pnp.js:4749:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

rizkyramadhan:
Strangely it works when I install esbuild-windows-64, WSL 2 is supposed to run native linux binary...

evanw:
I don't use yarn myself, but it appears to me that however you are running it is causing the binary executable to be interpreted as JavaScript code.

I once came across a strange page on the yarn repo that seemed to indicate that yarn 2 is deliberately incompatible with binary executables: https://github.com/yarnpkg/berry/issues/882. This seems kind of unbelievable to me so I'm not sure if it's true. But could that be the case here?

If so, it could be that you could get it to work by changing how you are invoking it perhaps (no ideas for what to try, sorry), or by using esbuild's [JavaScript API](https://github.com/evanw/esbuild/blob/eb0190fb97ede7de2476d646899272601d4bbdcc/docs/js-api.md#running-a-build) instead of the command-line interface. You could also consider using a different package manager if that still doesn't work.

If not, you could try providing more details here about how exactly you are running esbuild in case that makes something stand out.

arcanis:
> I once came across a strange page on the yarn repo that seemed to indicate that yarn 2 is deliberately incompatible with binary executables: [yarnpkg/berry#882](https://github.com/yarnpkg/berry/issues/882).

[...]

PR #494, opened 2020-10-30
Use JS bin wrapper if yarn berry detected during install

rtsao:
Fixes https://github.com/evanw/esbuild/issues/467

- Adds preservation of status code from spawned child process
- Installs using JS wrapper if Yarn 2+ is detected via `process.env.npm_config_user_agent`

### DIFF
--- a/lib/install.ts
+++ b/lib/install.ts
@@ -187,7 +187,18 @@ function removeRecursive(dir: string): void {
   fs.rmdirSync(dir);
 }
 
-function installOnUnix(name: string): void {
+function isYarnBerryOrNewer(): boolean {
+  const { npm_config_user_agent } = process.env;
+  if (npm_config_user_agent) {
+    const match = npm_config_user_agent.match(/yarn\/(\d+)/);
+    if (match && match[1]) {
+      return parseInt(match[1], 10) >= 2;
+    }
+  }
+  return false;
+}
+
+function installDirectly(name: string) {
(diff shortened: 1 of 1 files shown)


lib/install.ts
```diff
@@ -197,24 +208,37 @@ function installOnUnix(name: string): void {
...
   fs.writeFileSync(
     binPath,
     `#!/usr/bin/env node
 const path = require('path');
-const esbuild_exe = path.join(__dirname, '..', 'esbuild.exe');
+const esbuild_exe = path.join(__dirname, '..', ${JSON.stringify(wrapper_filename)});
 const child_process = require('child_process');
-child_process.spawnSync(esbuild_exe, process.argv.slice(2), { stdio: 'inherit' });
+const { status } = child_process.spawnSync(esbuild_exe, process.argv.slice(2), { stdio: 'inherit' });
+process.exitCode = status || 1;
```
evanw, line 219: Why do `|| 1` here? Doesn't this mean the exit code is always non-zero?

evanw:
Thanks for fixing this. This patch is going to require more hacks to avoid a slowdown with Yarn 2 because now the JavaScript API will instantiate two copies of node recursively before running the binary itself. I'll take it from here.

Note to self: Yarn 2 is really annoying to debug because cache invalidation is broken, at least for local `npm pack` packages. On macOS I had to remove and re-create the directory containing `package.json` and completely remove `~/.yarn` and `~/Library/Caches/Yarn` to get it to stop using the previous package contents.

PR #498, opened 2020-10-31
Initial work on CommonJS code splitting

jacob-ebey:
Here is some initial work on CommonJS code splitting. Looks like I was on an outdated copy but I'd love some feedback on the approach and help resolving the few failing tests.

I know there is a better way to do the `module.exports` generation using the "__exports" runtime utility but I'm not sure on the AST API for it.


jacob-ebey:
Looks like using `require` instead of import breaks things. Need to add some test cases to start tracking things down.

jacob-ebey:
Opening a new PR

PR #504, opened 2020-11-04
Fix broken metafile with css import from js

nitsky:
When a css file is imported from js, esbuild adds a js stub for it. The stub does not have `jsonMetadataChunk` set on it, because it isn't a real file. The code that generates the metafile does not check for this, which causes it to produce invalid json. This PR adds a check to confirm a file is not a stub before attempting to include it in the "inputs" section of the metafile.


evanw:
Oh no, sorry about that. Thanks for the fix.

PR #513, opened 2020-11-09
Introduce UMD as a new output format

prantlf:
--format=...  Output format (iife | cjs | umd | esm, no default
                  when not bundling, otherwise default is iife when
                  platform is browser and cjs when platform is node)

[UMD] format is similar to the IIFE format. Just the module wrapper and setting
the global variable differs. For example:

    // IIFE
    let moduleName = (() => {
      ... bundled code ...
      return exports;
    })();

    // UMD
    (function(root, factory) {
      if (typeof define === 'function' && define.amd) {
        define(factory);
      } else if (typeof module === 'object' && module.exports) {
        module.exports = factory();
      } else {
        root.returnExports = factory();
      }
    }(typeof self !== 'undefined' ? self : this, () => {
      ... bundled code ...
      return exports;
    }));

The module contents is generated with the help of the CJS wrapper, like IIFE.

This change supports only building of [standalone applications or libraries]. All
dependencies have to be included inside the output bundle or loaded by global
variables. Referring to external dependencies using AMD or CJS is not supported.

[UMD]: https://github.com/umdjs/umd
[standalone applications or libraries]: https://github.com/umdjs/umd/blob/master/templates/returnExports.js


internal/bundler/linker.go
```diff
@@ -4044,6 +4088,44 @@ func (c *linkerContext) generateGlobalNamePrefix() string {
 	return text
 }
 
+func generateModuleNameAssignment(options *config.Options) string {
```
jridgewell, line 4091: Should we expand this to the `iife` format, too?

jridgewell (approved):
Overall this looks good to me, though I'd love to see external module support. @prantlf, are you interested in resolving the merge conflicts? If not, I could open a new PR with your changes and the resolution commit.

imbhargav5:
Would be super cool to see this merged. Thanks for your work on this PR @prantlf ! 🙇🏽

cloverhearts:
oh my god, i want this

langhuihui:
i want this too

WNemencha:
Would be very useful to me too, what is blocking this to get merged?

bnussman (approved):

drewloomer:
For anyone trying to do this before this PR gets merged, the following worked for me:
```js
esbuild.build({
      // ...
      format: 'iife',
      globalName: 'Foo.Bar',
      target: 'es6',
      banner: {
        js: `(function(root, factory) {
  if (typeof define === "function" && define.amd) {
    define(factory);
  } else if (typeof module === "object" && module.exports) {
    module.exports = factory();
  } else {
    factory();
  }
}(typeof self !== "undefined" ? self : this, () => {
`
      },
      footer: {
        js: `
  return Foo.Bar;
}));
`
      },
      // ...
    })
```

Vanilagy:
Go for it bois

mlshv:
image

dy:
Hello there? 
Turns out that'd be of great help https://github.com/dy/sprae/issues/24

Issue #128, opened 2020-05-21
[Feature] Inline const enum values in TypeScript


evanw:
Right now `const enum` statements in TypeScript are treated like regular `enum` statements. This is consistent with how the TypeScript compiler itself behaves when the `isolatedModules` setting is enabled.

But it'd be nice if the values were inlined, since that would be better for run-time performance and code size. It should be possible for esbuild to do this since it bundles everything end-to-end and has the original TypeScript source code.

I'm not going to do this right now but I'm adding this issue now so I don't forget.

gregveres:
+1 in a big, big way. I have const enums all over my automatically generated typescript definitions for my back end. Our backend uses enums in most APIs since the back end is in C#. All of these enums are in namespaces that contain interface definitions and const enum definitions. In our current implementation, all of these things are compile time only entities because the const enums get compiled to const numbers.

zhusjfaker:
When I import files according to enumeration, esbuild will report an error

![image](https://user-images.githubusercontent.com/31839470/89507568-41df3180-d7ff-11ea-8c99-e5be9aae6947.png)

When enumerating in the same file, esbuild has no problem

![image](https://user-images.githubusercontent.com/31839470/89507715-78b54780-d7ff-11ea-92b0-029fb1f3a5e4.png)

evanw:
@zhusjfaker The problem you mentioned is not related to const enums, so it's irrelevant for this issue. Please file a new issue if you have a problem instead of adding on to an existing one. Also for what it's worth this looks to me like a configuration issue with [esbuild-loader](https://github.com/egoist/esbuild-loader) and Webpack, not with esbuild itself.

MarvinRudolph:
+1 This would be really cool 😊

Gelio:
I put some time into this and managed to come up with https://github.com/evanw/esbuild/pull/534

DanielRosenwasser:
Would it be better to to always inline enums regardless of const-ness? This might disincentivize people from using const enums in the first place.

gregveres:
@DanielRosenwasser why would always inlining enums dis-incentivize people from using const enums? 
Coming from a structured, typed programming language background, enums are a fantastic tool. I find it shocking that javascript developers use strings instead of enums as a widely accepted practice. Enums, compiled down to integers are a great way to increase the expressiveness in my program and to get good type safety at compile time. dissuading people from using const enums is, in my opinion, the wrong way to go.

leeoniya:
[...]

PR #534, opened 2020-11-15
Inline and erase local const enums

Gelio:
The aim of this PR is to partially address https://github.com/evanw/esbuild/issues/128

This PR changes the behavior of handling const enums:
1. For unexported const enums:
    1. Inline and erase if they are used only as known property access
    2. Inline values, but do not erase the enum if it is used in some other way than known property access (e.g. using the enum object in a function call)
2. For exported const enums and const enums inside namespaces:
    1. Preserve them, i.e. keep the existing logic

Changes in this PR:
1. Detect `const enum`s (`IsConst` field on `SEnum`)
1. Display errors when non-literals are used in a `const enum`
1. Add warnings when an unknown member of an enum is referenced (this is for both `const` and regular enums)
1. Mark local const enums as `CanBeRemovedIfUnused` and `DoesNotAffectTreeShaking`. Combined with calling not calling `recordUsage` on the enum, this lets the existing mangler (I think) erase the declaration if it is not used in any other way.

    I thought this way is probably the most elegant.

Keep in mind this is one of my first contributions to esbuild, and one of my first lines of go code, so please point out any mistakes/suggestions 😄

### DIFF
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -154,19 +154,19 @@ func TestTSInvalidConstEnum(t *testing.T) {
 			AbsOutputFile: "/out.js",
 		},
 		expectedScanLog: `entry.ts: error: Enum member must have initializer
-entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
-entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
-entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
+entry.ts: error: Const enum member initializers can only contain literal values and other computed enum values
+entry.ts: error: Const enum member initializers can only contain literal values and other computed enum values
+entry.ts: error: Const enum member initializers can only contain literal values and other computed enum values
 entry.ts: warning: Unknown member H on enum Foo
 entry.ts: warning: Unknown member X on enum Foo
 `,
 	})
 }
 
-func TestTSValidConstEnum(t *testing.T) {
+func TestTSLocalConstEnumWithUnknownPropertyAccess(t *testing.T) {
 	ts_suite.expectBundled(t, bundled{
 		files: map[string]string{
-			"/entry.ts": `
+			"/common.ts": `
 				const enum Foo {
 					A = 1,
 					B = A * 2,
@@ -174,7 +174,12 @@ func TestTSValidConstEnum(t *testing.T) {
 					D = 'stringsWorkToo',
 				}
 				
-				let usage = bar(Foo.A, Foo.B, Foo.C, Foo.D, Foo.E)
+				export let usage = bar(Foo.A, Foo.B, Foo.C, Foo.D, Foo.E)
+			`,
+			"/entry.ts": `
+				import { usage } from './common.ts';
+
+				usage();
 			`,
 		},
 		entryPaths: []string{"/entry.ts"},
@@ -182,8 +187,66 @@ func TestTSValidConstEnum(t *testing.T) {
 			Mode:          config.ModeBundle,
 			AbsOutputFile: "/out.js",
 		},
-		expectedScanLog: `entry.ts: warning: Unknown member E on enum Foo
+		expectedScanLog: `common.ts: warning: Unknown member E on enum Foo
 `,
+	})
+}
+
+func TestTSInlinableLocalConstEnum(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/common.ts": `
+				const enum Foo {
+					A = 1,
+					B = A * 2,
+					C,
+					D = 'stringsWorkToo',
+				}
+
+				export let usage = bar(Foo.A, Foo.B, Foo.C, Foo.D)
+			`,
+			"/entry.ts": `
+				import { usage } from './common.ts';
+
+				usage();
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+			MangleSyntax:  true,
+		},
+	})
+}
+
+func TestTSInlinableLocalConstEnumInsideNamespace(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/common.ts": `
+				export namespace NS {
+					const enum Foo {
+						A = 1,
+						B = A * 2,
+						C,
+						D = 'stringsWorkToo',
+					}
+
+					export let usage = bar(Foo.A, Foo.B, Foo.C, Foo.D)
+				}
+			`,
+			"/entry.ts": `
+				import { NS } from './common.ts';
+
+				NS.usage();
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+			MangleSyntax:  true,
+		},
 	})
 }
 
@@ -199,6 +262,7 @@ func TestTSImportConstEnum(t *testing.T) {
 			"/entry.ts": `
 				import { Foo } from './common';
 				
+				// Imported const enums are not inlined
 				let usage = bar(Foo.A, Foo.B)
 			`,
 		},
@@ -216,7 +280,7 @@ func TestTSImportConstEnumInsideNamespace(t *testing.T) {
 			"/common.ts": `
 				export namespace NS {
 					const enum Base {
-						A = 1,
+						A = 'from base enum',
 					}
 
 					export const enum Foo {
@@ -233,7 +297,7 @@ func TestTSImportConstEnumInsideNamespace(t *testing.T) {
 			"/entry.ts": `
 				import { NS } from './common';
 				
-				let usage = bar(NS.Foo.A, NS.Foo.B, Foo.FromBase)
+				let usageWithoutInliningValues = bar(NS.Foo.A, NS.Foo.B, NS.Foo.FromBase)
 				let noErrorsShownFor = bar(NS.invalidMember, NS.Foo.invalid)
 			`,
 		},
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -219,8 +219,12 @@ func TestTSLocalMinifiedConstEnumMerging(t *testing.T) {
 			`,
 			"/premature-use.ts": `
 				const enum Foo {BAR=1, BAZ=2};
+				function falsePositiveWarningInsideFunction() {
+					console.log(Foo.BAR, Foo.BAZ, Foo.ZOO);
+				}
 				console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
-				const enum Foo {BOO=3};
+				const enum Foo {BOO=3, ZOO=4};
+				falsePositiveWarningInsideFunction();
 			`,
 		},
 		entryPaths: []string{"/valid-use.ts", "/premature-use.ts"},
@@ -230,7 +234,8 @@ func TestTSLocalMinifiedConstEnumMerging(t *testing.T) {
 			MangleSyntax: true,
 			AbsOutputDir: "/",
 		},
-		expectedScanLog: `premature-use.ts: warning: Unknown member BOO on enum Foo
+		expectedScanLog: `premature-use.ts: warning: Unknown member ZOO on enum Foo
+premature-use.ts: warning: Unknown member BOO on enum Foo
 `,
 	})
 }
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -187,6 +187,29 @@ func TestTSValidConstEnum(t *testing.T) {
 	})
 }
 
+func TestTSImportConstEnum(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/common.ts": `
+				export const enum Foo {
+					A = 1,
+					B = 'a',
+				}
+			`,
+			"/entry.ts": `
+				import { Foo } from './common';
+				
+				let usage = bar(Foo.A, Foo.B)
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+		},
+	})
+}
+
 func TestTSDeclareEnum(t *testing.T) {
 	ts_suite.expectBundled(t, bundled{
 		files: map[string]string{
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -210,6 +210,43 @@ func TestTSImportConstEnum(t *testing.T) {
 	})
 }
 
+func TestTSImportConstEnumInsideNamespace(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/common.ts": `
+				export namespace NS {
+					const enum Base {
+						A = 1,
+					}
+
+					export const enum Foo {
+						A = 1,
+						B = 'a',
+						FromBase = Base.A
+					}
+
+					let usageInsideNamespace = bar(Base.A, Foo.A, Foo.invalid)
+				}
+
+				let usageInBaseModule = bar(NS.Foo.A, NS.Foo.FromBase, NS.Foo.invalid)
+			`,
+			"/entry.ts": `
+				import { NS } from './common';
+				
+				let usage = bar(NS.Foo.A, NS.Foo.B, Foo.FromBase)
+				let noErrorsShownFor = bar(NS.invalidMember, NS.Foo.invalid)
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+		},
+		expectedScanLog: `common.ts: warning: Unknown member invalid on enum Foo
+`,
+	})
+}
+
 func TestTSDeclareEnum(t *testing.T) {
 	ts_suite.expectBundled(t, bundled{
 		files: map[string]string{
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -229,10 +229,10 @@ func TestTSLocalMinifiedConstEnumMerging(t *testing.T) {
 		},
 		entryPaths: []string{"/valid-use.ts", "/premature-use.ts"},
 		options: config.Options{
-			RemoveWhitespace: true,
+			RemoveWhitespace:  true,
 			MinifyIdentifiers: true,
-			MangleSyntax: true,
-			AbsOutputDir: "/",
+			MangleSyntax:      true,
+			AbsOutputDir:      "/",
 		},
 		expectedScanLog: `premature-use.ts: warning: Unknown member ZOO on enum Foo
 premature-use.ts: warning: Unknown member BOO on enum Foo
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -192,6 +192,42 @@ func TestTSLocalConstEnumWithUnknownPropertyAccess(t *testing.T) {
 	})
 }
 
+func TestTSLocalConstEnumMergingWithoutWarnings(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.ts": `
+				const enum Foo {BAR=1, BAZ=2};
+				const enum Foo {BOO=3};
+				console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+		},
+	})
+}
+
+func TestTSLocalMinifiedConstEnumMerging(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.ts": `
+				const enum Foo {BAR=1, BAZ=2};
+				const enum Foo {BOO=3};
+				console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			RemoveWhitespace: true,
+			MinifyIdentifiers: true,
+			MangleSyntax: true,
+			AbsOutputDir: "/",
+		},
+	})
+}
+
 func TestTSInlinableLocalConstEnum(t *testing.T) {
 	ts_suite.expectBundled(t, bundled{
 		files: map[string]string{
--- a/internal/bundler/bundler_ts_test.go
+++ b/internal/bundler/bundler_ts_test.go
@@ -130,6 +130,63 @@ func TestTSDeclareNamespace(t *testing.T) {
 	})
 }
 
+func TestTSInvalidConstEnum(t *testing.T) {
+	ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.ts": `
+				const enum Foo {
+					A = 1,
+					B = A * 2,
+					C,
+					D = 'stringsWorkToo',
+					E,
+					F = {},
+					G = someIdentifier,
+					H = function () { }
+				}
+
+				let usage = bar(Foo.A, Foo.B, Foo.H, Foo.X)
+			`,
+		},
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+		},
+		expectedScanLog: `entry.ts: error: Enum member must have initializer
+entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
+entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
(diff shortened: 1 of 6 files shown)


internal/bundler/snapshots/snapshots_default.txt
```diff
@@ -2006,9 +2006,9 @@ var require_es6_ns_export_async_function = __commonJS((exports) => {
 var require_es6_ns_export_enum = __commonJS((exports) => {
   var ns2;
   (function(ns3) {
-    let Foo3;
-    (function(Foo4) {
-    })(Foo3 = ns3.Foo || (ns3.Foo = {}));
+    let Foo2;
```
Gelio, line 2009: I presume that the numbering in this snapshot is changed because the `const enum` was not emitted

internal/bundler/bundler_ts_test.go
```diff
@@ -130,6 +130,63 @@ func TestTSDeclareNamespace(t *testing.T) {
...
+		entryPaths: []string{"/entry.ts"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			AbsOutputFile: "/out.js",
+		},
+		expectedScanLog: `/entry.ts: error: Enum member must have initializer
+/entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
+/entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
+/entry.ts: error: const enum member initializers can only contain literal values and other computed enum values
+/entry.ts: warning: Unknown member H on enum Foo
```
Gelio, line 160: `H` is a known member, but I did not manage to determine that 😞 This is because the value is not saved in `knownEnumValues`.

Any suggestions on how to best handle that would be appreciated

internal/bundler/snapshots/snapshots_ts.txt
```diff
@@ -305,6 +305,12 @@ var Foo;(function(e){let a;(function(p){foo(e,p)})(a=e.Bar||(e.Bar={}))})(Foo||(
 ---------- /b.js ----------
 export var Foo;(function(e){let a;(function(p){foo(e,p)})(a=e.Bar||(e.Bar={}))})(Foo||(Foo={}));
 
+================================================================================
+TestTSValidConstEnum
+---------- /out.js ----------
+// /entry.ts
+let usage = bar(1, 2, 3, "stringsWorkToo", Foo.E);
```
Gelio, line 312: The unknown member of `Foo` is still emitted. I am not sure how to proceed here - whether to leave the current behavior (emit only a warning) or emit an error, as this most likely will never work

Suggestions are welcome

internal/js_parser/js_parser.go
```diff
@@ -73,7 +73,7 @@ type parser struct {
 	shouldFoldNumericConstants bool
 	emittedNamespaceVars       map[js_ast.Ref]bool
 	isExportedInsideNamespace  map[js_ast.Ref]js_ast.Ref
-	knownEnumValues            map[js_ast.Ref]map[string]float64
+	knownEnumValues            map[js_ast.Ref]map[string]js_ast.E
```
Gelio, line 76: This is to handle both `EString` and `ENumber`, as with `const enum`s both can be inlined.

Any better approaches for how to handle it better would be great 🙂

internal/js_parser/js_parser.go
```diff
@@ -7364,23 +7377,26 @@ func (p *parser) visitAndAppendStmt(stmts []js_ast.Stmt, stmt js_ast.Stmt) []js_
...
-		valueStmts := []js_ast.Stmt{}
-		if len(valueExprs) > 0 {
-			if p.MangleSyntax {
-				// "a; b; c;" => "a, b, c;"
-				joined := js_ast.JoinAllWithComma(valueExprs)
-				valueStmts = append(valueStmts, js_ast.Stmt{Loc: joined.Loc, Data: &js_ast.SExpr{Value: joined}})
-			} else {
-				for _, expr := range valueExprs {
-					valueStmts = append(valueStmts, js_ast.Stmt{Loc: expr.Loc, Data: &js_ast.SExpr{Value: expr}})
+		if !s.IsConst {
```
Gelio, line 7380: Right now `valueStmts` are still computed when `IsConst`, despite the fact they will never be used.

We could optimize it further to avoid computing `valueStmts` in this case. Let me know if you strongly want it done in this PR

Gelio:
Thanks for the `esprima` test, I noticed there is a problem with importing the const enum. I added a test case to capture the behavior. I do not have a clue how to fix it, though. Any pointers to places I would need to change would be awesome

evanw:
Unfortunately this isn't simple to accomplish, which is the reason why I haven't attempted this yet. Here are some of the trickier cases:

```ts
export namespace N {
  export const enum E {
    X = 123
  }
}
console.log(N.E.X)
```

```ts
import {N} from './file' // where "file" is the code written above
console.log(N.E.X)
```

Doing this transformation correctly in all cases requires a type system, and it's a non-goal for esbuild to replicate TypeScript's type system due to the complexity and maintenance overhead. That's why I currently fall back to having `const enum` behave like `enum`.

There is an approach where you try to do this when you can and fall back to regular enums when you can't. Doing that means detecting whether or not the identifier for the enum is ever used for something other than a known property access. This is doable for a non-exported enum within a single file, and is hard to do (but not impossible) across multiple files.

Gelio:
Thanks for the detailed description of the problem. I'll try to experiment a bit more over the next few days

internal/js_parser/js_parser.go
```diff
@@ -7999,17 +7999,30 @@ func (p *parser) visitAndAppendStmt(stmts []js_ast.Stmt, stmt js_ast.Stmt) []js_
...
 					nextNumericValue = e.Value + 1
 				case *js_ast.EString:
 					hasStringValue = true
+					if s.IsConst {
+						valuesSoFar[name] = e
+					}
+				default:
+					if s.IsConst {
+						r := js_lexer.RangeOfIdentifier(p.source, value.Value.Loc)
+						p.log.AddRangeError(&p.source, r, "const enum member initializers can only contain literal values and other computed enum values")
```
Gelio, line 8013: If the const enum is exported, then the enum will still be emitted, given current logic. Thus, despite the error, the code will work. Do you think it is worth it to leave the error in, and not expose internals (that const enums are still emitted when they are exported)?

The same goes for the _Enum member must have initializer_ error.

Gelio:
Hey! Sorry for the lack of activity 😟

I rebased the MR on the latest `master` and adjusted the code a bit. 

Right now:
1. `export const enum` are preserved
2. Some values (numbers and strings) in `export const enum`s are inlined in the same module/namespace they are defined
3. Values of `export const enum` are not inlined outside of the enum's namespace
4. Not exported `const enum` are skipped, no matter how they are used. Their values are inlined

The `TestTSImportConstEnumInsideNamespace` test should show most of the results when combining namespaces and const enums.

The doubts that I have:
1. When a `const enum` is preserved (e.g. because it is exported), should we inline string values (as if the `const enum` would be skipped)? I believe that those string values should remain enum property access, so then those `const enum`s behave like regular enums
2. When an unexported `const enum` is used in a way other than known property access, there will possibly be a warning, and the enum would still be skipped. This is what you mentioned in an earlier comment. For example:

    ```ts
	const enum Foo {
	  A = 1,
	}

	(Foo as any).B = 2;
	let usage = bar(Foo.A, Foo.B)
    ```

    We could try to detect such an unexpected use and decide to keep the enum. Given this input, TypeScript skips the enum [TS playground](https://www.typescriptlang.org/play?ssl=7&ssc=31&pln=1&pc=1#code/MYewdgzgLgBApmArgWxgMRCGBvAUDGAQRgF4YBGAGlwF9dcAKDLAQwhhbAE8BKAOgBCpGACYA3PQA2cWIggsA5nGEAjFgCcmmPoUrptAnmKA). As far as I see, esbuild detects that `(Foo as any).B` is still property access on an enum, so I reckon there would be more changes necessary to detect this scenario.

    One approach I was thinking about was to count the number of known property accesses for an enum (similarly to `parser.recordUsage`), and compare it with the number of times the enum was used. If they are not equal, then the enum should be preserved.

    If I'm not mistaken, this would require to know all the uses of the enum (also those below the enum declaration) when the enum is scanned. Is this possible without scanning twice?

    Do you think this is the right approach?

jods4:
Thanks for working on that! We have tons of const enums and that's a lot of dead weight if they aren't "const".

Are there many cases other than namespaces that are troublesome?
If esbuild supported compiling away const enums outside namespaces, that would already be a big improvement for many if not most real-world use-cases.

Keeping const enums in namespaces as plain enums could remain a documented limitation.

Full support of namespaced enums is out of question anyway because as you noted, namespaces can be aliased and then require full typing to know what is what.

On the other hand I believe you can't alias a const enum itself, so that's probably doable.
> - You can alias the enum as a type but we don't care for codegen: `const enum E {}; type F = E`
> - You can alias an import but that's something you can track without too much trouble `import { E as F } from "./enums"`

Gelio:
@jods4 I appreciate the support :slightly_smiling_face: As far as I know, the 2 cases in which `const enum`s are hard to erase are
1. Namespaces
2. `export const enum` (I believe this is an error with the `--isolatedModules` flag, which is recommended when using esbuild)

In the solution I prepared, in both cases the `const enum` will be left intact as if it was a regular `enum`. 

Only unexported `const enum`s (used only inside of a single module) will be erased :slightly_smiling_face:

jods4:
I think 2. is doable and is an important use case as enums are often used across files.

For example in our codebase we have _lots_ of enums auto-generated in a single file that is _just_ const enums exports and nothing else.

I believe that:
- `export const enum` is actually treated as `export enum` with the `--isolatedModules` flag. As I said I use const enums a lot and they still work in snowpack, vite or webpack w/ fork-ts-checker-plugin. It's just that the codegen is much bigger than it should.
- You can't reference a `const enum` value any other way than directly on the declared enum name. So it's actually trackable without full typing. You just need to know which imports are consts and substitute the values. This still requires inter-module analysis but that's the point of #128.

This last bullet is not true once a const enum is exported from a namespace, because the namespace itself can be aliased into new variables, parameters, destructured, etc. etc. This means 1. is not doable without full typing, which esbuild is not going to do.

Gelio:
What you are saying about point 2 makes sense. I got fooled by thinking esbuild does not have access to symbols exported in a dependency module when parsing another file, but that would not make sense when bundling (I hope I am not mistaken here :smile: )

Back to the drawing board with this MR then, I guess :slightly_smiling_face: I'll try to experiment a bit more with erasing exported const enums, but leaving the ones inside namespaces intact.

Gelio:
I've got some sad news. I give up on inlining const enum values across files. I've been trying and failing to do that for the last 9h and it looks like a major endeavor 😄 

Some findings:
1. The problem is that `parser.knownEnumValues` is local to only 1 file. There is no way to know if an identifier imported from another file is an enum, and what values it has.
2. We could probably implement some global store for `knownEnumValues`, or attach known enum values to each exported identifier, so it is accessible from its dependents. However, that seems like a major task, so I did not try to do that
3. I tried to make it so that instead of transpiling the enum statement right away in the parser, we do it lazily, in the printer. That did not work too well, because then the linker did not pick up that the enum is exported, and did not see the identifier. I abandoned that path since I thought it may be more time-consuming.

All in all, @evanw was definitely right and I underestimated the effort 😅 

I managed to (hopefully) implement the easier solution that @evanw suggested in https://github.com/evanw/esbuild/pull/534#issuecomment-727667807. I will update the description of the PR

@evanw This is a small enhancement, as it only provides value for local const enums, but I reckon we still could try to merge it. Let me know what you think

Gelio:
Hey @evanw, I'm wondering whether you are still interested in this addition to esbuild. My intent is not to apply any pressure, but just want to be sure if I should be hoping to get this landed any time soon 🙂

kzc:
TS has an odd feature where new members can be added to pre-existing enums:
```
$ cat enum.ts 
enum Foo {BAR=1, BAZ=2};
enum Foo {BOO=3};
console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
```
Notice that no warnings are generated by tsc:
```
$ node_modules/.bin/tsc enum.ts --outFile enum.js && node enum.js
1 2 3
```
This PR produces:
```
$ esbuild enum.ts | node
 > enum.ts:3:12: warning: Unknown member BAR on enum Foo
    3 │ console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
      ╵             ~~~

 > enum.ts:3:21: warning: Unknown member BAZ on enum Foo
    3 │ console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
      ╵                      ~~~

2 warnings
1 2 3
```

Gelio:
@kzc Oh, that's surprising 😄 Thanks for pointing it out! I'll try to incorporate it into the solution and add a test for this case.

I guess it's a matter of not creating a new `knownEnumValues` every time.

kzc:
Here's a test case variation using `const enum` that generates incorrect pure annotations...

Given:
```
$ cat const_enum.ts
const enum Foo {BAR=1, BAZ=2};
const enum Foo {BOO=3};
console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
```
Expected:
```
$ node_modules/.bin/tsc const_enum.ts --outFile const_enum.js && node const_enum.js
1 2 3
```
With this PR:
```
$ esbuild const_enum.ts --log-level=error
var Foo;
/* @__PURE__ */ (function(Foo2) {
  Foo2[Foo2["BAR"] = 1] = "BAR";
  Foo2[Foo2["BAZ"] = 2] = "BAZ";
})(Foo || (Foo = {}));
;
/* @__PURE__ */ (function(Foo2) {
  Foo2[Foo2["BOO"] = 3] = "BOO";
})(Foo || (Foo = {}));
;
console.log(Foo.BAR, Foo.BAZ, 3);
```
it's fine without minification:
```
$ esbuild const_enum.ts --log-level=error | node
1 2 3
```
but when run through terser the incorrect annotations produce:
``` 
$ esbuild const_enum.ts --log-level=error | node_modules/.bin/terser -bc
var Foo;

Foo || (Foo = {}), Foo || (Foo = {}), console.log(Foo.BAR, Foo.BAZ, 3);
```
which leads to the incorrect result:
```
$ esbuild const_enum.ts --log-level=error | node_modules/.bin/terser -bc | node
undefined undefined 3
```
Likewise with esbuild:
```
$ esbuild const_enum.ts --log-level=error --bundle --minify-syntax
(() => {
  // const_enum.ts
  /* @__PURE__ */ (function(Foo2) {
    Foo2[Foo2.BOO = 3] = "BOO";
  })(Foo || (Foo = {}));
  console.log(Foo.BAR, Foo.BAZ, 3);
})();
```
```
$ esbuild const_enum.ts --log-level=error --bundle --minify-syntax | node
[stdin]:5
  })(Foo || (Foo = {}));
     ^

ReferenceError: Foo is not defined
```

kzc:
Side note - not specific to this PR...

In my opinion instead of the following esbuild generated output:
```
$ echo 'export enum E {A, B};' | esbuild --loader=ts
export var E;
(function(E2) {
  E2[E2["A"] = 0] = "A";
  E2[E2["B"] = 1] = "B";
})(E || (E = {}));
;
```
this ought to be generated:
```
$ cat proposed.js 
// export enum E {A, B};
var E = /*@__PURE__*/ (function(E2) {
  return E2[E2["A"] = 0] = "A", E2[E2["B"] = 1] = "B", E2;
})(E || {});
export {E};
```
since it can be dropped altogether if not used:
```
$ echo 'import {E} from "./proposed.js";' | esbuild --bundle --minify
(()=>{})();
```
```
$ echo 'import {E} from "./proposed.js";' | rollup --silent
// no output
```
and still function correctly if used:
```
$ echo 'import {E} from "./proposed.js"; console.log(E.A, E.B);' | esbuild --bundle --minify | node
0 1
```
```
$ echo 'import {E} from "./proposed.js"; console.log(E.A, E.B);' | rollup --silent | node
0 1
```

gregveres:
The suggested output that @kzc suggests seems to be the best approach. In the issue thread, we said that the real issue of not getting rid of the enums was the code that got generated. If the code gen was changed to output the suggested code, and it was removed by rollup, that seems like the perfect solution.

Gelio:
I've rebased the branch on the latest `master` and added 2 new commits that fix the issues mentioned by @kzc above. I have also added tests that are based on the snippets you provided, so thanks 🙂

It turns out that I had to change the symbol merge strategy for enums - it was `mergeReplaceWithNew`, and it needs `mergeKeepExisting`. The tests passed, but I am not 100% certain why the strategy was different initially.

One behavior that is different from `tsc` is error reporting for prematurely used enum members. In the following snippet:

```ts
enum Foo {BAR=1, BAZ=2};
console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
enum Foo {BOO=3};
```

`tsc` is fine with using `Foo.BOO` before it's defined. The current behavior of esbuild will print a warning here (validated by a test), which IMO is a useful enhancement.

Unfortunately, due to the single-pass parsing nature of esbuild, the following valid code also reports a warning and `Foo.ZOO` is not inlined 😟:

```ts
const enum Foo {BAR=1, BAZ=2};
function falsePositiveWarningInsideFunction() {
	console.log(Foo.BAR, Foo.BAZ, Foo.BOO);
}
const enum Foo {BOO=3};
falsePositiveWarningInsideFunction();
```

This will likely be a problem with any use of an enum above the place of its declaration.

---

On the output proposed by @kzc - yeah, that would do the job. However, esbuild's output is similar to tsc's one, which also defines a variable in a separate statement, and assigns to it in an IIFE. See [this TS playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBAYg9nKBvAQgQQEoF4CMAaKdALSwCYBfAbgCgBjOEAZzgBtgA6FuAcwAp447dBgIChaIgEoaQA). I agree that having the syntax you proposed when it's possible (probably when there is a single enum declaration, without any enum merging) would be ideal. I won't pursue that in this PR, though

---

I see that the CI for windows failed, but that looks to be related to the failure on `master` caused by https://github.com/evanw/esbuild/commit/471e743fcaf27b8fcd8b1c3dabd2e5cb15c60ce7

Gelio:
Yet another rebase on `master`, since #1074 is now merged and the CI should start passing

kzc:
While the `const enum` values are indeed inlined, the unused pure annotated generated IIFEs for the enums are not always dropped by `esbuild`:

With the latest PR:
```
$ cat common.ts 
export namespace NS {
  const enum Foo {
    A = 1,
    B = A * 2,
    C,
    D = 'stringsWorkToo',
  }
  export let usage = bar(Foo.A, Foo.B, Foo.C, Foo.D)
}
```
```
$ cat entry.ts 
import { NS } from './common.ts';
NS.usage();
```
```
$ esbuild entry.ts --bundle --minify-syntax
(() => {
  // common.ts
  var NS;
  (function(NS2) {
    let Foo;
    /* @__PURE__ */ (function(Foo2) {
      Foo2[Foo2.A = 1] = "A", Foo2[Foo2.B = 2] = "B", Foo2[Foo2.C = 3] = "C", Foo2.D = "stringsWorkToo";
    })(Foo || (Foo = {})), NS2.usage = bar(1, 2, 3, "stringsWorkToo");
  })(NS || (NS = {}));

  // entry.ts
  NS.usage();
})();
```
`esbuild` is unable to drop the unused pure annotated IIFE above because it is not top level.

Even latest `rollup` (for whatever reason) is unable to drop the unused pure annotated IIFE generated by esbuild:
```
$ esbuild entry.ts --bundle --minify-syntax | rollup --silent
(() => {
  // common.ts
  var NS;
  (function(NS2) {
    let Foo;
    /* @__PURE__ */ (function(Foo2) {
      Foo2[Foo2.A = 1] = "A", Foo2[Foo2.B = 2] = "B", Foo2[Foo2.C = 3] = "C", Foo2.D = "stringsWorkToo";
    })(Foo || (Foo = {})), NS2.usage = bar(1, 2, 3, "stringsWorkToo");
  })(NS || (NS = {}));

  // entry.ts
  NS.usage();
})();
```
Although `uglify-js` and `terser` can drop the unused pure annotated IIFEs in question, most `esbuild` users would probably not elect to use them:
```
$ esbuild entry.ts --bundle --minify-syntax | npx uglify-js --toplevel -mc
var a;(a||(a={})).usage=bar(1,2,3,"stringsWorkToo"),a.usage();
```
```
$ esbuild entry.ts --bundle --minify-syntax | npx terser --toplevel -mc
var a;!function(a){let o;o||(o={}),a.usage=bar(1,2,3,"stringsWorkToo")}(a||(a={})),a.usage();
```
Using the pure annotated IIFE approach outlined in https://github.com/evanw/esbuild/pull/534#issuecomment-808787321 would allow `esbuild` (and `rollup`) to drop these unused enum IIFEs.

Ideally, this PR could not emit the unused enum IIFEs at all and avoid this issue altogether.

Gelio:
Thanks for the followup and more suggestions. I have been testing out various approaches now to get the output to match your suggestion @kzc. 

For the following input files:
```
			"/common.ts": `
				const enum Foo {
					A = 1,
					B = A * 2,
					C,
					D = 'stringsWorkToo',
				}

				export let usage = bar(Foo.A, Foo.B, Foo.C, Foo.D)
			`,
			"/entry.ts": `
				import { usage } from './common.ts';

				usage();
			`,
```

I have managed to get the following esbuild's output:

```
// common.ts
var usage = bar(1, 2, 3, "stringsWorkToo");

// entry.ts
usage();
```

which looks pretty good. 

I'll need to try some more approaches. Also, the change is quite messy at the moment, so I'll take some time to improve it.

However, as it does seem to be a different way to tackle the problem with const enums, I plan on releasing it as a separate PR. Both PRs could be merged independently, as IMO they introduce different ideas. Also, I don't want to get this one unstable.

A couple of notes that I have on my mind:
1. Is it common to run files through rollup after esbuild? esbuild is not as good as cutting out unused `PURE`-annotated expressions, so the benefit may not be noticeable for most users.
2. Is the benefit of this approach only related to removing unused enums? And in the context of const enums, the benefit would be that the closure generated by const enums that are inside of namespaces would be erased.

    If so, that seems pretty marginal to me.

    I'm not certain how often namespaces are used, but I'd say pretty rarely. At least I have not seen one used outside of declaration files in the last 2 years. ES modules are used more often, and in the case of ES modules, the generated closures are erased correctly (see [the output for `TestTSInlinableLocalConstEnum`](https://github.com/evanw/esbuild/pull/534/files#diff-6cb093d3444ec04df352db68d80b61659bd6b80d1d947e7d9c9168f56b05e359R309-R315))

    If that is the case, I'm on the fence about pursuing that idea, as it involves more work, and is a breaking change

Let me know if I'm mistaken in some point

kzc:
> I have managed to get the following esbuild's output:
> `var usage=bar(1,2,3,"stringsWorkToo");usage();`

That's great.

> Is it common to run files through rollup after esbuild?

For applications, probably not. But if esbuild is used to produce library bundles then it would be common for those library packages to be consumed by other bundlers.

> Is the benefit of this approach only related to removing unused enums?

If you already know which enums are in use, and they are not exported from the source file, then there's no advantage to emitting them. And if an enum is exported or if the enum identifier itself is used in a non-enumeration way then it would be a good thing if it could be tree shakable if not necessary.

> I'm not certain how often namespaces are used, but I'd say pretty rarely

100% agreement on that point.

> and is a breaking change

Not to my knowledge. The esbuild user would not notice the missing unused enum IIFEs.

kzc:
@Gelio Just want to mention that my comments are from the perspective of an esbuild user, and I was just mentioning what features that I'd like to see. So feel free to ignore my input. Ultimately it's up to you and @evanw to hash out.

Gelio:
Thanks for the answers again. To my mind, the issue about changing the emitted syntax for enums is orthogonal to this PR and the point about inlining const enum values (which was the main point of #128).

If you reckon changing the output syntax could have benefits, it deserves a separate issue, that may catch the attention of @evanw who definitely has moree knowledge than me about the possible implications of such a change (and does not seem to be active in this PR anymore, unfortunately). If you create an issue for it, I may chime in with some comments and findings I came across during the experiment yesterday :slightly_smiling_face: 

I want to keep the discussion in this PR related to the PR and the issue it tries to solve

kzc:
Changing the enum code generation was just mentioned as a generalization because this PR was emitting `/*@__PURE__*/` annotated IIFEs that were not assigned to a variable - meaning they ought to be unconditionally dropped, aside from the IIFE argument `Enum || (Enum = {})`. Because esbuild [only performs tree shaking on top level code](https://github.com/evanw/esbuild/issues/639#issuecomment-753683962), problems resulting from an incorrectly annotated IIFE would only be evident by subsequently using another tool like uglify-js, terser or rollup.

AlonMiz:
@Gelio thanks a lot. i wish we can merge this soon.
what's holding you back?

Gelio:
@AlonMiz This PR has not been reviewed by @evanw yet at all, which I believe is a prerequisite to getting it merged. I am waiting for his input (or any other review, for that matter).

I don't think there is much for me to do specifically, aside from maybe resolving merge conflicts and remembering the changes. Since there is no clear direction from @evanw, I don't want to spend time on this PR now only to have it be forgotten for a few months and require a revisit in the future

kzc:
Looks like https://github.com/evanw/esbuild/pull/534#issuecomment-808787321 was just implemented and merged: https://github.com/evanw/esbuild/commit/731f559940b1ad37037407d22118f2ec56bd7133

jods4:
Nice improvement, and great to see that `const enums` are not set in stone yet.

In my projects we generate tons of enums from server-side model and reference them in _other_ files, so I'm still hoping for basic direct reference simplification, i.e.:
```js
// Direct import of an exported enum const
import { E } from "./enums"
console.log(E.One)

// Compiles to:
console.log(1)
```

Hopefully, this simplification will eventually be doable without full type analysis as it only requires to mark exports that are const enums.

That still doesn't support more convoluted cases but it'd be good enough as it covers 99% cases for me.

evanw:
Closing this PR now that #128 has been fixed.

Issue #535, opened 2020-11-16
Start using GitHub releases to notify about new releases


timocov:
This is not issue of the tool.

GitHub has nice option to subscribe on repo's releases (instead of subscribing on any activity in the repo to see what's going on) - it allows you do not track every single activity in the repo, but being notified about new releases. Currently, it seems that there is changelog file and git tags.

Do you have a plan to start to use GitHub releases?

evanw:
It looks like there are already releases being generated, so I imagine you should already be able to subscribe to them: https://github.com/evanw/esbuild/releases. Are you just asking for the commit message to duplicate the information from the changelog for that release?

timocov:
> It looks like there are already releases being generated, so I imagine you should already be able to subscribe to them: evanw/esbuild/releases

Actually it's not releases, it's just git tags on that page I guess. I'm already subscribed on releases and haven't been notified so far :)

(for example how releases page looks like when you create releases - https://github.com/webpack/webpack/releases)

> Are you just asking for the commit message to duplicate the information from the changelog for that release?

Not really. To create github's release you need to open releases page and then select already created tag or just draft new release (without tag at the moment), fill the information about the release (name, git tag to attached to and description - I guess it could be copy-paste from changelog file) and then "publish release". After that all subscribers will be notified about that release. More info on https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository#creating-a-release.

Yeah, it might be additional work while releasing the new version so I'm not pushing you to use them at all, but I believe it could be useful to people who wants to look at progress of the tool but not get all notifications about every single comment/issue.

Also, it seems that releases could be part of the automatization on GitHub actions - for example in https://github.com/TypeStrong/ts-loader/blob/master/.github/workflows/release.yml when a new release is created (it could be attached to a tag only, so it certainly exists), GA runs publishing the package to npm.

rtsao:
You could probably do something much fancier with GitHub Actions and/or Releases, but https://github.com/evanw/esbuild/pull/539 should address this (and improve the releases page by displaying the relevant section of the changelog for each release) and doesn't alter your existing workflow in any way (you still commit locally, push the version tag, etc).

evanw:
This latest release was published as a GitHub release: https://github.com/evanw/esbuild/releases/tag/v0.8.9. This should be the case from now on.

timocov:
Yeah, it works perfectly now - just got an email with the new release. Thank you @rtsao!

PR #539, opened 2020-11-17
Automatically create GitHub Release on pushed tags

rtsao:
Fixes https://github.com/evanw/esbuild/issues/535

Adapted from https://github.com/actions/create-release#example-workflow---create-a-release

This job is trigged when a git tag matching `v*` is pushed. The relevant section of `CHANGELOG.md` is automatically extracted and used as the body of the created GitHub Release (if no changelog entry is found, an empty body is used). This ensures there is always corresponding GitHub Release for every newly published version.

The heredoc is needed to set the multiline changelog environment variable `CHANGELOG` that is consumed in the subsequent step that actually creates the release. This is based on the documentation at https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#multiline-strings

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,29 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Extract changelog
+        run: |
+          CHANGELOG=$(awk -v ver=$(cat version.txt) '/## / { if (p) { exit }; if ($2 == ver) { p=1; next} } p' CHANGELOG.md)
+          echo "CHANGELOG<<EOF" >> $GITHUB_ENV
+          echo "$CHANGELOG" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Extract changelog
+        run: |
+          CHANGELOG=$(awk -v ver=$(cat version.txt) '/## / { if (p) { exit }; if ($2 == ver) { p=1; next} } p' CHANGELOG.md)
+          echo "CHANGELOG<<EOF" >> $GITHUB_ENV
+          echo "$CHANGELOG" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+      - name: Create GitHub Release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rtsao, line 23: Note: this is automatically provided to actions, no need to create your own token

evanw:
Wow, cool! Thanks so much for putting this together. Let's give this a try with the next release.

evanw:
It worked! Thanks again.

PR #540, opened 2020-11-17
Upgrade actions/setup-node to 1.4.4

rtsao:
This should resolve the CI failures resulting from the [deprecation of the `set-path` command](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).


evanw:
Thanks so much for proactively fixing this!

Issue #541, opened 2020-11-17
Syntax error when using ternaries + async fat arrow + typecast


arcanis:
I admit the title is very specific 😄

The following structure doesn't parse with ESBuild, but seems fine per the TS playground:

```ts
_ as any ? async () => _ : _;
```

rtsao:
It appears this is actually a broader issue with typecasted ternary conditions, it's just that this particular case results in a parse error.

For example, esbuild currently transformed `a as any ? b : c` to `a;`

evanw:
The fix for this has been released. Thanks so much for reporting this.

PR #544, opened 2020-11-17
Fix typecasted ternary conditions

rtsao:
Fixes https://github.com/evanw/esbuild/issues/541

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -3420,7 +3420,7 @@ func (p *parser) parseSuffix(left js_ast.Expr, level js_ast.L, errors *deferredE
 			// Handle the TypeScript "as" operator
 			if p.TS.Parse && !p.lexer.HasNewlineBefore && p.lexer.IsContextualKeyword("as") {
 				p.lexer.Next()
-				p.skipTypeScriptType(js_ast.LLowest)
+				p.skipTypeScriptType(js_ast.LConditional)
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


evanw:
Thanks for the quick fix. Unfortunately it looks like something else is going on here. TypeScript compiles this to just `_`:

```ts
_ as number extends Object ? () => void : any;
```

We should read the source code for the TypeScript compiler to see what it does. It looks like `extends` is special-cased:

```ts
function parseTypeWorker(noConditionalTypes?: boolean): TypeNode {
    if (isStartOfFunctionTypeOrConstructorType()) {
        return parseFunctionOrConstructorType();
    }
    const pos = getNodePos();
    const type = parseUnionTypeOrHigher();
    if (!noConditionalTypes && !scanner.hasPrecedingLineBreak() && parseOptional(SyntaxKind.ExtendsKeyword)) {
        // The type following 'extends' is not permitted to be another conditional type
        const extendsType = parseTypeWorker(/*noConditionalTypes*/ true);
        parseExpected(SyntaxKind.QuestionToken);
        const trueType = parseTypeWorker();
        parseExpected(SyntaxKind.ColonToken);
        const falseType = parseTypeWorker();
        return finishNode(factory.createConditionalTypeNode(type, extendsType, trueType, falseType), pos);
    }
    return type;
}
```

rtsao:
Oh interesting, I guess we need to special-case `extends` as well. Based on that, it seems TypeScript interprets that as:

```ts
_ as (number extends Object ? () => void : any);
```

evanw (approved):
Ah I see. The `extends` expression results in a boolean, so it requires a conditional afterwards to resolve the boolean value back into a type. Thanks for the fix!

PR #552, opened 2020-11-20
Call all js plugins `onResolve` and `onLoad` functions

### LINKED ISSUE #501: Call all plugins resolvers instead of using only the first one
Currently only the first plugin onResolve result will be used, this means there is no way to override a previous plugin

This makes it impossible to compose plugins together and it will force you to create one big plugin to handle all necessary cases and prevent other plugins from early returning the wrong result

This also makes it impossible to use more than one onResolve plugin, for example you cannot use a yarn pnp resolver together with an http url resolver

I propose instead to call all plugins onResolve functions and use the previous plugins result as input for the next plugin, this means that plugin can override previous plugins

To do this 

This also makes plugins similar to a middlewares, this way one plugin can use other plugins results and build complex logic in a composable manner

remorses:
This makes it possible to override previous plugins, previously only the first matching plugin `onResolve` and `onLoad` was called

Fix #501


remorses:
Now plugins don't change the arguments passed to subsequent plugins anymore.

To enable data sharing between plugins maybe we could add a context argument that plugins can populate in a middleware like fashion, what do you think?

PR #561, opened 2020-11-25
types: overload `build` definitions

### LINKED ISSUE #559: Cannot invoke an object which is possibly 'undefined'
I'm not sure if this is intended, but the `.rebuild()` method from the `BuildResult` interface, accordingly to the snippet below, can also be undefined:

https://github.com/evanw/esbuild/blob/2fa473e4c2f5374e066353c9d37bdc8afa8da734/lib/types.ts#L88

Because of this, is necessary to adhere to some hacky solutions like this:

```typescript
const buildResult = await esbuildService.build({ ...esbuildConfig, incremental: true })
buildResult.rebuild!() // non-null assertions defeats the purpose of using "strict: true" on tsconfig.json
```

or also:

```typescript
const buildResult = await esbuildService.build({ ...esbuildConfig, incremental: true })
buildResult.rebuild && buildResult.rebuild()
```

or even this:

```typescript
const buildResult = await esbuildService.build({ ...esbuildConfig, incremental: true })
buildResult.rebuild?.call(this)
```

Is there anything that i'm missing? as shown in the example from [`0.8.12`](https://github.com/evanw/esbuild/blob/master/CHANGELOG.md#0812) release https://github.com/evanw/esbuild/blob/master/CHANGELOG.md), there were no need for any checking to use the `.rebuild()` method

```javascript
require('esbuild').build({
  entryPoints: ['app.js'],
  bundle: true,
  outfile: 'out.js',
  incremental: true,
}).then(result => {
  // The "rebuild" method is present if "incremental" is true. It returns a
  // promise that resolves to the same kind of object that "build" returns.
  // You can call "rebuild" as many times as you like.
  result.rebuild().then(result2 => {
    // Call "dispose" when you're done to free up resources.
    result.rebuild.dispose()
  })
})
```

lukeed:
Adds TypeScript [overload definitions](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) for `build` so that `write: false` and `incremental: true` remove the conditional existence of their relevant `BuildResult` properties.

---

Closes #559

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -82,10 +82,16 @@ export interface OutputFile {
   text: string; // "contents" as text
 }
 
+export type BuildInvalidate = () => Promise<BuildIncremental>;
+
+export interface BuildIncremental extends BuildResult {
+  rebuild: BuildInvalidate & { dispose(): void };
+}
+
 export interface BuildResult {
   warnings: Message[];
   outputFiles?: OutputFile[]; // Only when "write: false"
-  rebuild?: (() => Promise<BuildResult>) & { dispose(): void }; // Only when "incremental" is true
+  rebuild?: BuildInvalidate; // Only when "incremental" is true
 }
 
 export interface BuildFailure extends Error {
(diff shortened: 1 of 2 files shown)


evanw (approved):
Nice! This looks great. Thanks for doing this.

PR #563, opened 2020-11-26
Add support for custom host in serve api

jamalc:
This change would allow users to specify an alternate host when using the serve API. This is useful for developing with esbuild inside of a virtual machine/docker container or to request development assets from a remote testing device on the same network at a different IP address.


evanw (approved):
Looks great! Nice job. Thanks for adding this.

PR #564, opened 2020-11-26
fix: allow and pick darwin x86_64 binary on apple silicon macs

sod:
When you try to npm install esbuild on a m1 mac, you see this error:

Allow and pick darwin x86_64 binary on apple silicon macs. It'll run just fine through rosetta 2 translation, at least throwing the 8.4MB typescript.js file against it works without complaining:

Native darwin arm64 seems not to be supported yet by golang. It's planned for go v1.16 (from https://github.com/golang/go/issues/42756)

### DIFF
--- a/lib/install.ts
+++ b/lib/install.ts
@@ -265,6 +265,7 @@ const knownWindowsPackages: Record<string, string> = {
 };
 const knownUnixlikePackages: Record<string, string> = {
   'darwin x64 LE': 'esbuild-darwin-64',
+  'darwin arm64 LE': 'esbuild-darwin-64',
   'freebsd arm64 LE': 'esbuild-freebsd-arm64',
   'freebsd x64 LE': 'esbuild-freebsd-64',
   'linux arm64 LE': 'esbuild-linux-arm64',
--- a/npm/esbuild-darwin-64/package.json
+++ b/npm/esbuild-darwin-64/package.json
@@ -8,7 +8,8 @@
     "darwin"
   ],
   "cpu": [
-    "x64"
+    "x64",
+    "arm64"
   ],
   "directories": {
     "bin": "bin"


evanw (approved):
Oh huh, that's pretty neat. Definitely a good fix in the meantime. Thanks for trying it out and for putting this together. I'll have to trust you that it works since I don't have access to a M1 myself.

evanw:
This has been released. Please let me know if this works for you or not.

sod:
Yes, can confirm. `npm install esbuild`. And vite via `npm i -g vite; npm init vite-app vite-test; cd vite-test; npm run build` works as well.

RoXuS:
@sod I get
image
Do you do something else?

sod:
Do you have rosetta 2 installed?

RoXuS:
nop, oh I have to you use rosetta2 on this ?

sod:
x86 binaries don't run natively on m1. They have to be transpiled by rosetta 2 to run function.

Native arm64 binary for esbuild is not possible at this time, as golang has no support for it yet.

sod:
And you don't have to actively "use rosetta 2", it just has to be installed and then just transparently compiles x86 binaries to arm64 when it encounters one.

RoXuS:
Oh thx! Work as expected!

evanw:
Heads up that this might not be completely smooth sailing. The Hugo project (another Go-based web development tool) is getting reports of emulation issues with running Go-based code with Rosetta 2: https://github.com/gohugoio/hugo/issues/7968. I'm not yet aware of any reports like this from esbuild users but I wouldn't be surprised if it comes up.

RoXuS:
@sod, it is ok if you install go from head if you don't want to wait Go 1.16.
With this GIST -> https://gist.github.com/felixbuenemann/d08875fc3e1a69fe8d6eaa724829c950

Working great!

PR #568, opened 2020-11-30
Multiple exports for the same concrete type result in a build error that is inconsistent with TypeScript

ggoodman:
This PR is intended to encode behaviour that I _think_ should be considered valid but which is currently considered invalid in `esbuild`.

When multiple modules re-export the same underlying type (via different paths) this appears to produce an 'ambiguous import' error: https://github.com/evanw/esbuild/blob/1f3bd9d1834c5e411b2a9ad74dfa683cbb8b81dd/internal/bundler/linker.go#L1786-L1790

I think TypeScript is taking a different approach which is to say that since the 'duplicate' exports refer to the same concrete type, they can be deduplicated instead of being treated as conflicting. In the included test, there is only one `Foo` class but it gets exported via the entrypoint through two competing paths. Perhaps the linker needs a concept of source identity beyond identifier identity.

### DIFF
--- a/internal/bundler/bundler_importstar_ts_test.go
+++ b/internal/bundler/bundler_importstar_ts_test.go
@@ -510,10 +510,7 @@ func TestTSDuplicateExportRefs(t *testing.T) {
 			`,
 			"/re-export.ts": `
 				export * from './types1'
-				export * from './nested/re-export.ts'
-			`,
-			"/nested/re-export.ts": `
-				export * from '../types1'
+				export { Foo } from './types1'
 			`,
 			"/types1.ts": `
 				export class Foo {}
--- a/internal/bundler/bundler_importstar_ts_test.go
+++ b/internal/bundler/bundler_importstar_ts_test.go
@@ -500,3 +500,29 @@ func TestTSReExportTypeOnlyFileES6(t *testing.T) {
 		},
 	})
 }
+
+func TestTSDuplicateExportRefs(t *testing.T) {
+	importstar_ts_suite.expectBundled(t, bundled{
+		files: map[string]string{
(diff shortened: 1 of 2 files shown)


evanw:
Sorry, I'm confused. This test case seems to work fine? This is what the CI failure says:

```
--- FAIL: TestTSDuplicateExportRefs (0.00s)
    --- FAIL: TestTSDuplicateExportRefs/#00 (0.00s)
        bundler_importstar_ts_test.go:505: No snapshot saved for TestTSDuplicateExportRefs
            ---------- /out.js ----------
            // /types1.ts
            var Foo = class {
            };
            
            // /entry.ts
            console.log(Foo);
```

That means there was no build error and the build successfully generated an output file.

ggoodman:
Hi @evanw, sorry about that. Missed the correct workflow and have since added the test. Interestingly, with the snapshot, the test passes.

This tells me that I've been unable to reproduce the issue that I was seeing in a much larger codebase where an `export * from './catinthehat_crate'` was conflicting with a more specific export like `export { Thing1, Thing2 } from './catinthehat_crate'`.

I've since successfully gotten the offending code 'fixed' in the project on which I was trying to run `esbuild` so this is less of an issue now.

I leave to your judgement whether the (now passing) test adds value or not.

evanw:
> This tells me that I've been unable to reproduce the issue that I was seeing in a much larger codebase where an `export * from './catinthehat_crate'` was conflicting with a more specific export like `export { Thing1, Thing2 } from './catinthehat_crate'`.

Thanks for the tip. I'll treat this as an issue report and start investigating. I think I might know what this is about.

PR #582, opened 2020-12-06
Support JSON objects and arrays in `define`

### LINKED ISSUE #581: Support object literals for define
Currently the `define` transform only supports identifiers, strings, numbers, boolean, and null.  It would be convenient to also support object literals, or ideally arbitrary JavaScript expressions.  Object literals in particular are an important use case for me.

Is there a reason that can't be supported?

jbms:
Fixes #581.


viktor-mezhenskyi (approved):

evanw:
This was closed in favor of a more complicated inject-related implementation for the reasons described in #581.

PR #597, opened 2020-12-14
Use env variable ESBUILD_BINARY_PATH to use different binary path

### LINKED ISSUE #592: How to use local esbuild binary in the npm package?
I am trying to debug an error in esbuild golang code but i cannot find an easy way of using my local esbuild binary in the `esbuild` npm package

Maybe we can add an option `binaryPath` or use an environment variable `ESBUILD_BINAERY_PATH` as a solution, what do you think?

remorses:
Fix #592 

I also noticed that Yarn berry custom node binary is not used in windows, is that intentional?

### DIFF
--- a/lib/install.ts
+++ b/lib/install.ts
@@ -237,7 +237,10 @@ function installWithWrapper(name: string, fromPath: string, toPath: string): voi
     binPath,
     `#!/usr/bin/env node
 const path = require('path');
-const esbuild_exe = path.join(__dirname, '..', ${JSON.stringify(toPath)});
+let esbuild_exe = path.join(__dirname, '..', ${JSON.stringify(toPath)});
+if (process.env.ESBUILD_BINARY_PATH) {
+  esbuild_exe [path.resolve(process.env.ESBUILD_BINARY_PATH), []];
+}
 const child_process = require('child_process');
 const { status } = child_process.spawnSync(esbuild_exe, process.argv.slice(2), { stdio: 'inherit' });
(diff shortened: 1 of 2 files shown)


evanw:
Is there a reason you're modifying the wrapper as well as the JavaScript API? Modifying the wrapper to behave differently when that environment variable is present seems unimportant if you're also changing the JavaScript API to avoid calling the wrapper in the first place when that environment variable is present. Unless your code is calling the executable wrapper directly without going through the JavaScript API, but in that case your code could just call your new executable directly without any assistance from this code. Can you say more about your use case? How are you calling esbuild when you want to debug it?

> I also noticed that Yarn berry custom node binary is not used in windows, is that intentional?

A wrapper script is needed either if the installer is Yarn 2 or if the current OS is Windows. It's needed for Yarn 2 because Yarn 2 is incompatible with binary executables. If I remember correctly, it's needed with Windows because npm on Windows is incompatible with binary executables whose names don't end in `.exe` and I want the cross-platform command for esbuild to be `esbuild` not `esbuild.exe`. The check for Yarn 2 isn't done on Windows because the custom wrapper is always used on Windows regardless of whether you're using Yarn 2 or not, so there's nothing to do with the result of the check.

remorses:
> Is there a reason you're modifying the wrapper as well as the JavaScript API? 

Sorry i didn't understand the reason of the yarn berry handling, it's because in yarn berry the location of the esbuild binary is different

Issue #590, opened 2020-12-10
Faster sync API via  worker_thread


cspotcode:
I'm brand new to esbuild, but am looking for excuses to cut my teeth on more golang projects.

I see from various issues and pull requests that esbuild's sync APIs spawn a new external process for each transformation call, because this is the only way to synchronously pipe data to the external process and wait for a response.

To reduce the per-transformation overhead, I'm wondering if either of these solutions are a) possible and b) something I can work on as a pull request:

1. Use a `worker_thread` to perform async communication with an esbuild process, blocking the main thread.  I've implemented code that blocks the main thread while doing async stuff in a worker, so I could publish that as a reusable library and use it in esbuild.  (here is my code, specifically the `sync_worker` package: https://github.com/cspotcode/typescript-http-support)  I imagine usage would look like this pseudocode. (forgive me if this is obvious, as I say I'm brand new to esbuild)

```
const service = createEsBuildSyncCompileService() // <-- spawns worker_thread and esbuild process
service.transformSync(); // blocks main thread calling out to worker_thread, which uses async to talk to esbuild
```

2. Compile esbuild to WebAssembly and run it in-process.  I know less about the state of go-to-WASM compilation, but would enjoy the excuse to learn more.
EDIT I see there is already a webassembly package.  In that case, I can implement option (1) using the WebAssembly build.

evanw:
Oh wow what a hack. I love it. It looks like you're using the `SharedArrayBuffer` atomics API to block the main thread. I don't know what the cross-platform support for this feature is. MDN says it's supported in node 8+, which should be fine. It looks like `worker_thread` was added in node 10+. So I think it's ok? Hopefully there aren't any issues with node's atomics and less common processor architectures.

I think this could be good to add to esbuild, assuming benchmarks show that it's a big win (I expect them to). I want to keep esbuild dependency-free so it should be implemented as a part of esbuild itself instead of as a library that esbuild uses. This shouldn't require much additional code to support so including it in esbuild is fine.

It definitely shouldn't be implemented such that it requires WebAssembly though. The WebAssembly version is up to 10x slower than the native version for a variety of reasons. Instead, it should be implemented such that it works equally well with either the [native version](https://www.npmjs.com/package/esbuild) or the [WebAssembly version](https://www.npmjs.com/package/esbuild-wasm). Both versions use the same TypeScript library code (they just create a different executable as a child process) so this should naturally happen without any extra work.

[...]

PR #612, opened 2020-12-20
worker_threads

cspotcode:
I started a pull request for #590

One big change: it uses 2x threads: a watchdog and the worker.  The watchdog detects when the worker terminates and notifies all blocked callers, to avoid hanging in case of unexpected errors in the worker.

I also added safer workerData access to avoid node bugs where workerData isn't available right away.


evanw:
I'm closing this since this PR is old. Sync APIs with `worker_threads` was re-enabled after this PR was made and has been enabled by default for a while without issues.

PR #620, opened 2020-12-23
Add transpile options

thr0w:
* Add `--remove-console` option that avoids print console calls to target
* Add `--remove-debugger` option that avoids print debugger statement to target
* Add `--debug-tool=...` enable debug tool in target ([more info](https://github.com/throwforks/esbuild/blob/master/docs/debugtool.md))
* Add `--remove-debug-tool` remove debug tool from target([more info](https://github.com/throwforks/esbuild/blob/master/docs/debugtool.md))


evanw:
I understand the desire for these options but I am not planning on building such specific transforms into esbuild itself. People need all kinds of AST-related transforms and it would be unmaintainable to include them all in esbuild.

The way to do this kind of thing with esbuild is to use another AST transformation library in a plugin to transform the source code first before passing it to esbuild. And you can also get a similar code-removal effect with esbuild without using a plugin by putting the affected code behind [a `--define` variable](https://esbuild.github.io/api/#define) in a branch or potentially by [marking certain global functions as `--pure`](https://esbuild.github.io/api/#pure).

thr0w:
No problem! I understand @evanw . Thank you for your attention and congratulations for that great tool.
Happy 2021

PR #628, opened 2020-12-28
Fix realpath function

### LINKED ISSUE #627: Initial resolveDir is invalid when using Yarn Berry, paths in metafile.json are invalid too
When using yarn berry the initial `resolveDir` passed to plugins is invalid, for example `/usr/morse/github/github` becomes `/usr/morse/github/github`

The paths in metafile.json are invalid too, they have one more `../` than necessary, for example what should be `./index.ts` becomes `../index.ts`, this is probably because esbuild thinks cwd is `/usr/morse/github/github` instead of `/usr/morse/github`

This does not happen when using yarn 1, only with yarn berry

Reproduction:
https://github.com/remorses/esbuild-reproduction-wrong-resolveDir-in-yarn-berry

remorses:
Fix #627 

I use panic when there is an error in eval symlinks, this will panic when there is a cycle of symblic links, previously it would try to resolve the links forever˙

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -2,7 +2,6 @@ package fs
 
 import (
 	"errors"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


evanw:
Thanks for the fix. I'm removing the panic because esbuild shouldn't crash in this case. I added a comment with more rationale about why the error is ignored.

Unfortunately I was unable to reproduce the infinite loop case, neither in the old code nor in the new code. I just get `ELOOP`. Regardless there is now test coverage for a symlink cycle so it should catch problems if they come up in the future.

PR #656, opened 2021-01-09
Allows NodeJS to terminate also if there is an active service.

SalvatorePreviti:
Issue https://github.com/evanw/esbuild/issues/659

Curently, NodeJS will not terminate if there is an esbuild service running.

Running this, will not allow NodeJS to exit.
```
const esbuild = require('esbuild');
esbuild.startService();
```

This pull request allows the possibility to keep the esbuild service running in background ready to receive requests without the need to manually call stop() - NodeJS will terminate automatically when there is no operation in progress (transform, build or serve).

See https://nodejs.org/api/child_process.html#child_process_subprocess_unref for more technical details.

This change should not affect the normal usage, and will allow the implementation of features like experimental module loaders without hanging the NodeJS process and without resorting to restarting the child process every transformation. Will also simplify the usage of this library by bundler and build tools.

More than happy to discuss this.
And thank you again for esbuild!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,9 @@ jobs:
     - name: JS API Tests
       run: node scripts/js-api-tests.js
 
+    - name: NodeJS Unref Tests
+      run: node scripts/node-unref-tests.js
+
     - name: Plugin Tests
       run: node scripts/plugin-tests.js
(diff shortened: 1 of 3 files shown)


evanw (approved):
Looks good to me. Thanks!

PR #681, opened 2021-01-15
update webpack 5 benchmark

chenxsan:
Hi there, would be great if you can upgrade the benchmark for webpack 5 as it has been released for a while https://webpack.js.org/blog/2020-10-10-webpack-5-release/, and the performance had been improved compared to the current one in my test:

image

And here's the data for esbuild and rollup:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ require/webpack/node_modules:
 require/webpack5/node_modules:
 	mkdir -p require/webpack5
 	echo '{}' > require/webpack5/package.json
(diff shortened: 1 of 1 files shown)


evanw (approved):
Thanks. I will include this in the next benchmark update.

Note that the time you selected isn't the real time, and is not the one used in the benchmark. The real time is the one marked `real` below what you selected. You can't rely on Webpack's self-reported time because it doesn't include startup overhead.

PR #683, opened 2021-01-16
Move sourcesContent to common flags in JS API

### LINKED ISSUE #682: sourcesContent only added to build options not transform options in JS API
The docs (https://esbuild.github.io/api/#sources-content) list sourcesContent as a valid option for both transform and build but the code for the JS API defines sourcesContent only for build flags:

https://github.com/evanw/esbuild/blob/08ca8cfbf697f4c13af62931f84983f5210f39c9/lib/common.ts#L164

This means esbuild will error when you specify `sourcesContent: false` for the transform API:

```
  const esbuildOpts = {
    format: 'cjs',
    target: 'esnext',
    loader: loader,
    sourcemap: true,
    sourcesContent: false,
    sourcefile: sourcePath,
  };
  const result = esbuild.transformSync(sourceText, esbuildOpts);
```

Errors with:

    Invalid option in transformSync() call: "sourcesContent"

jschaf:
Fixes #682.

### DIFF
--- a/lib/common.ts
+++ b/lib/common.ts
@@ -91,6 +91,7 @@ function pushLogFlags(flags: string[], options: CommonOptions, keys: OptionKeys,
 }
 
 function pushCommonFlags(flags: string[], options: CommonOptions, keys: OptionKeys): void {
+  let sourcesContent = getFlag(options, keys, 'sourcesContent', mustBeBoolean);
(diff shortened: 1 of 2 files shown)


evanw (approved):
Thanks for the fix. Sorry about the oversight.

PR #695, opened 2021-01-19
fix missing 'neutral' in platform description

hardfist:


evanw (approved):
Whoops. Thanks for the catch.

PR #717, opened 2021-01-27
Fix exporting star when the export base is function

misak113:
* Until now, only objects (no functions) were supported to be exported as a star.
* The reason is that `Object.create` called on "function prototype" will not create callable function.
* The clonning "function" before it's exported, will make it work.
* Adding default and other properties was kept as before.


evanw:
It's not totally clear to me what problem this is solving because you didn't include any code demonstrating a problem. But if I had to guess, it looks like you're maybe trying to enable `import * as fn from 'path'` and then use `fn` as a function. Unfortunately that isn't how ECMAScript modules work. The ECMAScript standard says that `fn` here is a [module namespace exotic object](https://262.ecma-international.org/11.0/#sec-module-namespace-exotic-objects) and is not callable.

I understand the desire to make this work because some older non-conformant JavaScript cross-compilers (e.g. TypeScript without the `esModuleInterop` setting enabled) do allow this to work. But JavaScript tooling is converging on the actual behavior in the ECMAScript specification and esbuild will not be deviating from the specification here. You will have to change your code instead. You should be able to do `import fn from 'path'` which will work in node, browsers, and esbuild.

misak113:
Hi,
thx for the quick reaction. Sorry that I didn't provide more info about the use case. But it's exactly what you've mentioned. 
I totally agree that since this is not a part of ECMA standard, it should not be merge as it is. However, sometimes, in the real world we have to live with a decision of our past us. This is the exact moment. We have a long-living project written in TypeScript compiled using webpack which is using this style of imports out of standards.
I made this change and tried it on our company project and works great as a compatibility hack.

So what I propose is to allow configure esbuild to do this optionally based on options passed to esbuild.
Just one brainstormed proposal for a name: `exportStarFunctionsCompatibility: boolean`.

TL;DR
Our project is currently using webpack configuration for bundling typescript, sass, images etc. This setup takes almost 2 minutes to complete the build from scratch (with no caches).
Currently, I experimentally rewrote the configuration to esbuild and it takes under 5 seconds. WoW. This is really a must-have.
The only blocker is this "feature" to perform everything with zero changes in code.
If this "feature" is not possible, I'll be forced to maintain the fork with those few changes. I believe that more companies are dealing with the same problem.

P.S.: Could you point me where should I write E2E tests for this once we manage the final form?

eyelidlessness:
@misak113 your use case sounds like it could be solved either with a plugin (to transform those import statements automatically) or a one time automated codemod to correct those cases in the source.

PR #719, opened 2021-01-28
esbuild-wasm: don't create a new exit0 native module for each run

kzc:
* has a stable file name based on the hash of the contents of the shared library
* saves disk space for repeated runs

### DIFF
--- a/npm/esbuild-wasm/bin/esbuild
+++ b/npm/esbuild-wasm/bin/esbuild
@@ -2,6 +2,7 @@
 
 // Forward to the automatically-generated WebAssembly loader from the Go compiler
 
+const crypto = require('crypto');
 const path = require('path');
 const zlib = require('zlib');
 const fs = require('fs');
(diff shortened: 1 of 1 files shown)


npm/esbuild-wasm/bin/esbuild
```diff
@@ -44,9 +45,14 @@ process.on('exit', code => {
   const data = require('../exit0')[nativeModule];
   if (data) {
     try {
-      const tempFile = path.join(os.tmpdir(), `${Math.random().toString(36).slice(2)}-${nativeModule}`);
-      fs.writeFileSync(tempFile, zlib.inflateRawSync(Buffer.from(data, 'base64')));
-      require(tempFile);
+      const hash = crypto.createHash('sha256').update(data).digest().toString('hex').slice(0, 16);
+      const tempFile = path.join(os.tmpdir(), `${hash}-${nativeModule}`);
+      try {
+        require(tempFile);
```
evanw, line 51: I think the contents of the file should probably be validated here before using it.

kzc: I had thought about that. It would be yet another read, but sure.

kzc: done

npm/esbuild-wasm/bin/esbuild
```diff
@@ -41,12 +42,20 @@ process.on('exit', code => {
...
-  const data = require('../exit0')[nativeModule];
-  if (data) {
+  const base64 = require('../exit0')[nativeModule];
+  if (base64) {
     try {
-      const tempFile = path.join(os.tmpdir(), `${Math.random().toString(36).slice(2)}-${nativeModule}`);
-      fs.writeFileSync(tempFile, zlib.inflateRawSync(Buffer.from(data, 'base64')));
-      require(tempFile);
+      const data = zlib.inflateRawSync(Buffer.from(base64, 'base64'));
+      const hash = crypto.createHash('sha256').update(data).digest().toString('hex').slice(0, 16);
```
kzc, line 49: It would be faster to hash `base64` here rather than `data` due to its length:
```js
const hash = crypto.createHash('sha256').update(base64).digest().toString('hex').slice(0, 16);
```
What do you think?

kzc: I made this change.

evanw (approved):
Thanks for the fix!

PR #726, opened 2021-01-30
Fix Metafile types to account for non-JS assets

iamakulov:
In the metafile output, some non-JS output assets don’t have `imports` and/or `exports` fields in the metafile. This change fixes the TS types accordingly.

Screen Shot 2021-01-30 at 13 07

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -246,11 +246,11 @@ export interface Metadata {
           bytesInOutput: number
         }
       }
-      imports: {
+      imports?: {
         path: string
         kind: MetadataImportKind
       }[]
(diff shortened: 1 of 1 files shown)


iamakulov:
I’m not sure where the source code for `esbuild.github.io` is, so I didn’t update it – but I believe https://esbuild.github.io/api/#metafile will also need to be changed accordingly.

evanw:
Thanks for the PR. However, I think the metafile would be easier to use if these properties were just always present. So I'm going to fix this problem that way instead. Then you shouldn't have to check for `undefined` in these places.

PR #531, opened 2020-11-15
Add support for banner and footer

Gelio:
This PR is an attempt to fix https://github.com/evanw/esbuild/issues/482.

It adds support for `banner` and `footer` options, that allow for prepending/appending a user-provided string to each output file.
Similar to [rollup's options](https://rollupjs.org/guide/en/#outputbanneroutputfooter).

I do not have full confidence that I have done everything correctly, as this is my first ever PR in go, which I began learning yesterday 😅 

Things I did (hopefully correctly):
* add `Banner` and `Footer` options to configs
* add emitting those in the linker
* take account of the banner in the sourcemap
* add test cases for sourcemaps when `banner` and `footer` are specified
* add end-to-end test cases when `banner` and `footer` are specified
* add handling those options in the JS service

I'm looking for any feedback you have 🙂 

## Example

`test.js`:

```js
console.log('hello world');
```

```sh
14:11 $ ./esbuild test.js --sourcemap --banner="/* Hello world */" --footer="/* footer */"
/* Hello world */
console.log("hello world");
/* footer */
//# sourceMappingURL=data:application/json;base64,ewogICJ2ZXJzaW9uIjogMywKICAic291cmNlcyI6IFsidGVzdC5qcyJdLAogICJzb3VyY2VzQ29udGVudCI6IFsiY29uc29sZS5sb2coJ2hlbGxvIHdvcmxkJyk7XG4iXSwKICAibWFwcGluZ3MiOiAiO0FBQUEsUUFBUSxJQUFJOyIsCiAgIm5hbWVzIjogW10KfQo=
```

[...]

PR #775, opened 2021-02-09
Feat: add `footer` and `banner` for css file

### LINKED ISSUE #712: banner and footer flags are not honored in minified CSS
In package.json:
```json
    "dist-2": "esbuild src/css/main.css --bundle --minify --footer='/* Copyright 2021 Foo */' --outfile=dist/main.css",
```
Result: CSS is minified, but no banner/footer is added.  Also `@license` and `/*! -- */` comments in the CSS source are not honored as they are with JS bundling.

v0.8.32

susiwen8:
Close #712

add `css-footer` and `css-banner` for `CSS` file

### DIFF
--- a/cmd/esbuild/main.go
+++ b/cmd/esbuild/main.go
@@ -54,16 +54,18 @@ var helpText = func(colors logger.Colors) string {
   --watch               Watch mode: rebuild on file system changes
 
 ` + colors.Bold + `Advanced options:` + colors.Default + `
-  --banner=...              Text to be prepended to each output file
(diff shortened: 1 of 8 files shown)


susiwen8:
@evanw Hi, could you take a look for this, something similar with #531

evanw:
I am planning for this to be a breaking API change instead of fixing it in a non-breaking change like this. I want to keep the API surface small and use a map like `outExtension` instead of having a separate flag per content type, especially since there may be even more content types than JS and CSS in the future.

PR #778, opened 2021-02-09
fix #776: css skip BOM

### LINKED ISSUE #776: Skip over a leading BOM in CSS files
Hello, it seems the @import statement in .css file when trying to bundling it, does not work. I get the following error ( the resulting bundle does not include correctly the code where @import points:

```
css/app.css: warning: Unexpected "@import"
    1 │ ﻿@import "./bootstrap/bootstrap.min.css";
      ╵    ~~~~~~~

1 warning
```

-> _app.css_ 

```css
@import "./bootstrap/bootstrap.min.css";
html, body {
    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
/* ..... */
}
```

->  _esbuild-config.js_

```js
require('esbuild').build({
            entryPoints: ['./css/app.css'],
            bundle: true,
            outfile: './wwwroot/css/app.min.css',
            minify: true,
        })
```

But if i place @import statement at bottom or the in middle of the app.css file, then it works. The issue seems to happen just if the @import statement is placed on top.

Maybe there's something i wrong but i don't know what. Please check it about.

esbuild version is: 0.8.43

susiwen8:
Close #776

Logic in [`printWithEscape`](https://github.com/evanw/esbuild/blob/master/internal/css_printer/css_printer.go#L436) for escaping `BOM` is not right. it has already written `\feff` before escape.

Add a addition `isEscapeBom` params for BOM

### DIFF
--- a/internal/css_printer/css_printer.go
+++ b/internal/css_printer/css_printer.go
@@ -51,7 +51,7 @@ func (p *printer) printRule(rule css_ast.R, indent int, omitTrailingSemicolon bo
 	case *css_ast.RAtNamespace:
 		if r.Prefix != "" {
 			p.print("@namespace ")
-			p.printIdent(r.Prefix, identNormal, canDiscardWhitespaceAfter)
+			p.printIdent(r.Prefix, identNormal, canDiscardWhitespaceAfter, false)
 		} else {
 			p.print("@namespace")
 		}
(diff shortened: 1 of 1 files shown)


evanw:
Sorry, this isn't the correct fix. The problem is that it's being parsed in the first place, not that the U+FEFF character needs to be omitted from the incorrectly-parsed CSS. This should be fixed in the parser instead. I'm going to close this since I have a fix for the parser.

PR #787, opened 2021-02-11
fix #786: avoid absolute paths for disabled modules

### LINKED ISSUE #785: Relative imports ignored via “browser” field cause absolute paths in source maps
While looking into why our builds were not perfectly reproducing between different systems, I ended up finding that in some cases, absolute paths end up in the source map file, causing the hash to change.

Example (snippet from .map file):
```
"empty:/mnt/ramdisk/node_modules/paper/dist/node/self.js", "empty:/mnt/ramdisk/node_modules/paper/dist/node/extend.js", "../../../../node_modules/paper/dist/paper-core.js"
```

Turns out, this is caused by the `browser` field in the `package.json` of dependencies. In this case: https://github.com/paperjs/paper.js/blob/095671073b9cfed2221b48aea54776ca6f2aed70/package.json#L45-L52
```
  "browser": {
    "canvas": false,
    "jsdom": false,
    "jsdom/lib/jsdom/living/generated/utils": false,
    "source-map-support": false,
    "./dist/node/self.js": false,
    "./dist/node/extend.js": false
  },
```

I believe this is caused by the `resolver`, specifically: https://github.com/evanw/esbuild/blob/9580aa8a6164b1896abcc7ae5423da062efa2faf/internal/resolver/resolver.go#L400

And

https://github.com/evanw/esbuild/blob/9580aa8a6164b1896abcc7ae5423da062efa2faf/internal/resolver/resolver.go#L452-L457

I’m not sure why this has to be an absolute path, nor why these should be part of the sourcemap, and as this causes differences between builds on different systems, I’d love to get rid of it.

### LINKED ISSUE #786: Module imports ignored via “browser” field cause absolute paths in source maps
Follow-up of https://github.com/evanw/esbuild/issues/785, absolute paths can still happens with module imports.

Example from https://github.com/socketio/engine.io-client/blob/master/package.json#L69-L74:
```
  "browser": {
    "ws": false,
    "xmlhttprequest-ssl": "./lib/xmlhttprequest.js",
    "./lib/transports/websocket-constructor.js": "./lib/transports/websocket-constructor.browser.js",
    "./lib/globalThis.js": "./lib/globalThis.browser.js"
  },
```

Causes an absolute path in the source map:
```
"(disabled):/Users/eelco/Projects/Example/node_modules/engine.io-client/node_modules/ws/browser.js"
```

eelco:
Use the “file” namespace for resolved remapped modules, to ensure the paths are relative in the sourcemap.

Fixes #786.

Follow-up of 419b8011e768d132de3577cb9414909beb6a9bbf (which fixed #785).

### DIFF
--- a/internal/bundler/snapshots/snapshots_packagejson.txt
+++ b/internal/bundler/snapshots/snapshots_packagejson.txt
@@ -50,7 +50,7 @@ var index;
 ================================================================================
 TestPackageJsonBrowserMapModuleDisabled
 ---------- /Users/user/project/out.js ----------
-// (disabled):/Users/user/project/node_modules/node-pkg/index.js
+// (disabled):Users/user/project/node_modules/node-pkg/index.js
 var require_node_pkg = __commonJS(() => {
 });
 
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -448,9 +448,9 @@ func (r *resolver) resolveWithoutSymlinks(sourceDir string, importPath string, k
 					if remapped == nil {
 						// "browser": {"module": false}
 						if absolute, ok := r.loadNodeModules(importPath, kind, sourceDirInfo); ok {
-							absolute.Primary = logger.Path{Text: absolute.Primary.Text, Flags: logger.PathDisabled}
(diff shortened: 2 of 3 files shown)


evanw:
Thanks for the fix. Sorry about the oversight. This will go out with the next release.

PR #799, opened 2021-02-12
fix #791: onResolve is not fired on `entrypoints` when namespace equals 'file'

### LINKED ISSUE #791: onResolve is not fired on entrypoints when namespace equals 'file'
Hey there!

I was writing a plugin and realized that `onResolve` is not fired on entrypoints when namespace equals `'file'`. 

```js
build.onResolve({ filter: /someregexp/, namespace: 'file' }, args => {});
```

↑ callback here **won't be fired** on entrypoint JS file. 

```js
build.onResolve({ filter: /someregexp/ }, args => {});
```

↑ callback here **will be fired** on entrypoint JS file. And `args.namespace` will be an empty string.

It is probably related to #546, but I'm not sure. Is it a normal thing, or a bug? I want to make sure that my plugin do not process anything not related to it, and have decided to use `namespace: 'file'`. But I also need to process entrypoints. So right now I have to omit namespace setting in `onResolve` filtering.

(I can create a small example project to illustrate the behavior.)

susiwen8:
Close #791 
normally `esbuild` default `namespace` is [`file`](https://esbuild.github.io/plugins/#namespaces), but in this [commit](https://github.com/evanw/esbuild/commit/05eaca4d15d8971eb51d8a9da00a2e147ee27a97), it seems it has been forgotten to set `file` as default.

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -635,7 +635,7 @@ func runOnResolvePlugins(
 		ResolveDir: absResolveDir,
 		PluginData: pluginData,
 	}
-	applyPath := logger.Path{Text: path}
+	applyPath := logger.Path{Text: path, Namespace: "file"}
 	if importSource != nil {
 		resolverArgs.Importer = importSource.KeyPath
 		applyPath.Namespace = importSource.KeyPath.Namespace


evanw:
Sorry but I am not going to land this because the current behavior is by design. See #791 for details.

PR #810, opened 2021-02-15
add cjsWrap info in metafile

### LINKED ISSUE #680: [Feature] Add isCommonjs information in metafile
I need to know when a bundle has been generated from a commonjs module, previously i was using the exports field and checking if there was only a default export but this would give too many false positives

```json
{
  "outputs": {
    "react/index.js": {
      "isCommonjs": true,
      "exports": ["default"],
      "imports": [
        ...
      ],
      "inputs": {
        ...
      },
      "bytes": 119
    }
}
```

remorses:
close #680

Add the new `cjsWrap` information in the metafile

`cjsWrap` means that the output bundle exports have been wrapped in a default export, tools like [bundless](https://github.com/remorses/bundless) can use this information to rewrite named imports

```json
{
  "outputs": {
    "react/index.js": {
      "cjsWrap": true,
      "exports": ["default"],
      "imports": [
        ...
      ],
    
    }
}
```

### DIFF
--- a/internal/bundler/linker.go
+++ b/internal/bundler/linker.go
@@ -3743,9 +3743,19 @@ func (repr *chunkReprJS) generate(c *linkerContext, chunk *chunkInfo) func(gener
 			if !isFirstMeta {
 				jMeta.AddString("\n      ")
 			}
+			jMeta.AddString("],\n      ")
+
+			// cjsWrap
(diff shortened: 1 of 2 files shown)


evanw:
Sorry, but this is an internal implementation detail that I don't want to expose. It's an artifact of the bundling strategy and may change in the future (and has already changed once since this PR was created). So I'm not going to go with this PR.

PR #811, opened 2021-02-15
entryPoint information in metafile

### LINKED ISSUE #711: Feature request: facadeModuleId in metafile outputs
When building a multi-entry bundle with `splitting: true`, it is often necessary to map output chunks to their original entry modules. Currently, the metafile doesn't provide enough information to do this 100% accurately because two output chunks may contain the same inputs module (if entry B transitively imports entry A). What I am doing right now is checking which output chunk has the same ending sub path segments with the input modules, but this can potentially yield false positives if the paths happen to collide.

Rollup entry chunks contains a `facadeModuleId` property which points to the original entry module that this chunk was created from. It would be nice to have something similar in esbuild's metafile outputs.

remorses:
Close #711

Adds a `entryPoint` field in the metafile

```json
{
  "outputs": {
    "react/index.js": {
      "entryPoint": "src/entry.js",
      "exports": ["default"],
      "imports": [
        ...
      ],
    
    }
}
```

### DIFF
--- a/internal/bundler/linker.go
+++ b/internal/bundler/linker.go
@@ -3782,9 +3782,8 @@ func (repr *chunkReprJS) generate(c *linkerContext, chunk *chunkInfo) func(gener
 				jMeta.AddString("\n      ")
 			}
 			if chunk.isEntryPoint {
-				file := c.files[chunk.sourceIndex]
-				facadeModuleID := file.source.PrettyPath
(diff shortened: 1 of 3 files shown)


evanw:
> Not sure we want to use this name, maybe `inputEntrypoint` is better?

Yeah "facade" doesn't tell me anything about what this property means. Maybe just `entryPoint` to mirror the `entryPoints` API option?

Note to self: This PR doesn't mark CSS files generated by importing CSS from a JavaScript entry point with this property. However, I think that is actually a good idea because that keeps it a 1:1 mapping. And in the future, I may need to generate more than one CSS file in this case so annotating the CSS files with this property is potentially confusing. I think I should instead add a new property to link the JavaScript entry point to the necessary CSS files instead.

Issue #560, opened 2020-11-25
esbuild prints warning on usage of require.main


vladmandic:
nodejs v15 deprecated usage of `module.parent` to detect if module is loaded or running as standalone:
`[DEP0144] DeprecationWarning: module.parent is deprecated due to accuracy issues. Please use require.main to find program entry point instead.`

however, esbuild throws warning when using `require.main`:
`warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)`
`    178 │ if (require.main === module) test();`

like the suggestion says, placing try/catch block around it works, but that is anything but elegant.

evanw:
Thanks for pointing this out. This is a common pattern so you're right, esbuild should deal with this. I think it should be sufficient to skip warning about `require.main` if the output format is CommonJS. With other output formats, `require` may be undefined so it still seems appropriate to warn (and to wrap with `try/catch`).

metachris:
This seems to be still an issue; this typical pattern doesn't work in a bundled output script:

```typescript
if (require.main === module) {
  // ...
}
```

PR #813, opened 2021-02-15
Stop indirect call warning for 'require.cache' for CommonJS modules

### LINKED ISSUE #812: Warnings for use of require.cache
Hi, thanks for esbuild!

We've been using it with an Amazon CDK project, and noticed that we get a bunch of warnings about [the `require.cache` property](https://nodejs.org/api/modules.html#modules_require_cache) from the following file (or at least, the compiled JS version):

https://github.com/aws/aws-cdk/blob/c0944d2fcd4351997b8b8e981f88cf3145e065b4/packages/%40aws-cdk/core/lib/private/runtime-info.ts

```
 > ../node_modules/@aws-cdk/aws-lambda/node_modules/@aws-cdk/core/lib/private/runtime-info.js: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    15 │     for (const fileName of Object.keys(require.cache)) {
       ╵                                        ~~~~~~~

 > ../node_modules/@aws-cdk/aws-lambda/node_modules/@aws-cdk/core/lib/private/runtime-info.js: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    64 │     const mod = require.cache[fileName];
       ╵                 ~~~~~~~

...
```

AIUI, these are explicitly not designed to be affecting bundling, are are just for introspecting the current set of packages. As such, they should be okay and not warned about, similar to `require.main` in #560.

(Fair warning: I don't understand all the moving parts here, so this could easily be a valid warning that needs to be fixed/addressed/managed upstream.)

huonw:
This follows/expands https://github.com/evanw/esbuild/commit/4e503ba2ada87da775d918d77e6ce035a51dce37 (#560) to suppress the warning about an indirect call to 'require' for `require.cache`, in addition to `require.main`, when building CommonJS modules. This is because manipulating/reading the cache shouldn't affect the overall bundling (AIUI).

I've validated that this suppresses the warning for the file from #812:

```shell
curl 'https://raw.githubusercontent.com/aws/aws-cdk/c6256992902fc4237ceb9f965e970e2c2ef00777/packages/%40aws-cdk/core/lib/private/runtime-info.ts' > runtime-info.ts
./esbuild --format=cjs runtime-info.ts
```

Output before:

```
 > runtime-info.ts: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    16 │   for (const fileName of Object.keys(require.cache)) {
       ╵                                      ~~~~~~~

 > runtime-info.ts: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    69 │   const mod = require.cache[fileName];
       ╵               ~~~~~~~

var __create = Object.create;
var __defProp = Object.defineProperty;
var __getProtoOf = Object.getPrototypeOf;
...
```

Output after:

```
var __create = Object.create;
var __defProp = Object.defineProperty;
var __getProtoOf = Object.getPrototypeOf;
...
```

(I'm not sure if I've updated the CHANGELOG correctly.)

Fixes #812

### DIFF
--- a/internal/bundler/bundler_default_test.go
+++ b/internal/bundler/bundler_default_test.go
@@ -1185,6 +1185,51 @@ func TestRequireWithoutCallInsideTry(t *testing.T) {
 	})
 }
 
+func TestRequirePropertyAccessCommonJS(t *testing.T) {
+	default_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.js": `
+				// These shouldn't warn
+				console.log(Object.keys(require.cache))
+				console.log(Object.keys(require.extensions))
+				delete require.cache['fs']
+				delete require.extensions['.json']
+			`,
+		},
+		entryPaths: []string{"/entry.js"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			OutputFormat:  config.FormatCommonJS,
+			AbsOutputFile: "/out.js",
+		},
+	})
+}
+
+func TestRequirePropertyAccessES6(t *testing.T) {
+	default_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.js": `
+				// These shouldn't warn
+				console.log(Object.keys(require.cache))
+				console.log(Object.keys(require.extensions))
+				delete require.cache['fs']
+				delete require.extensions['.json']
+			`,
+		},
+		entryPaths: []string{"/entry.js"},
+		options: config.Options{
+			Mode:          config.ModeBundle,
+			OutputFormat:  config.FormatESModule,
+			AbsOutputFile: "/out.js",
+		},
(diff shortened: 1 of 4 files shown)


internal/bundler/bundler_default_test.go
```diff
@@ -1185,6 +1185,51 @@ func TestRequireWithoutCallInsideTry(t *testing.T) {
...
+			AbsOutputFile: "/out.js",
+		},
+	})
+}
+
+func TestRequirePropertyAccessES6(t *testing.T) {
+	default_suite.expectBundled(t, bundled{
+		files: map[string]string{
+			"/entry.js": `
+				// These shouldn't warn
```
UncleBill, line 1212:
```diff
-				// These shouldn't warn
+				// These shouldn warn`
```

This should be a typo?

Issue #550, opened 2020-11-19
Add darwin arm64 LE build


mhart:
Currently get this on the latest Macs:

```
Unsupported platform: darwin arm64 LE
```

mhart:
I realize this won't be possible until you've got access to Go that supports it: https://github.com/golang/go/issues/38485#issuecomment-730247329

But I figured it's worth having an issue to track it

evanw:
I was wondering when this would come in. From what I understand, Go is currently in feature freeze for 1.15 and this isn't going to be supported by Go until the 1.16 release in February.

evanw:
Sounds like esbuild should be working on the M1 soon. It turns out the simple way to fix this is just to ship the `darwin x64 LE` version. This was done in #564.

RoXuS:
I got the following error:

![image](https://user-images.githubusercontent.com/3296671/101170982-09583480-35f4-11eb-8668-f85571716829.png)

This is the only one that does not work :(

If I can help?

285851937:
@evanw is there any way to install the esbuild on Raspberry Pi 4 ?
My Pi's info:
```
# uname -a
Linux raspberrypi 4.19.97-v7l+ #1294 SMP Thu Jan 30 13:21:14 GMT 2020 armv7l GNU/Linux
```
and after install esbuild,there is an error:
```
Unsupported platform: linux arm LE
```
ps: It's ``` arm``` NOT ```arm64```
Thanks~~

evanw:
> Unsupported platform: linux arm LE

That's tracked by issue #528, not this issue (this issue is about macOS).

In theory your device should now be supported in version 0.8.22. I don't have a Raspberry Pi myself, so I have not tested it. Let me know in issue #528 if you have any problems.

285851937:
excuse me got a wrong issuse. I've been upgraded raspbian to 64bit version, like this:```Linux raspberrypi 5.4.79-v8+ #1373 SMP PREEMPT Mon Nov 23 13:32:41 GMT 2020 aarch64 GNU/Linux``` and install the **0.8.22** successfully. thank you a lot...
PS:[steps of upgrading **arm** to **arm64** for Raspberry Pi](https://www.mmuaa.com/post/0c9188ffde4e2cff.html) if any one need to upgrad  Raspberry Pi.

sudarshansb143:
If someone is facing same issue with angular please refer here 

[...]

PR #820, opened 2021-02-16
Add native builds for ARM-based Macs

rtsao:
Resolves https://github.com/evanw/esbuild/issues/550

### DIFF
--- a/lib/install.ts
+++ b/lib/install.ts
@@ -86,12 +86,7 @@ async function installBinaryFromPackage(name: string, fromPath: string, toPath: 
 }
 
 function validateBinaryVersion(binaryPath: string): void {
-  let stdout;
-  try {
-    stdout = child_process.execFileSync(binaryPath, ['--version']).toString().trim();
-  } catch (err) {
-    throw err;
-  }
+  const stdout = child_process.execFileSync(binaryPath, ['--version']).toString().trim();
   if (stdout !== version) {
     throw new Error(`Expected ${JSON.stringify(version)} but got ${JSON.stringify(stdout)}`);
   }
--- a/lib/install.ts
+++ b/lib/install.ts
@@ -90,14 +90,6 @@ function validateBinaryVersion(binaryPath: string): void {
   try {
     stdout = child_process.execFileSync(binaryPath, ['--version']).toString().trim();
   } catch (err) {
-    if (platformKey === 'darwin arm64 LE')
-      throw new Error(`${err && err.message || err}
-
-This install script is trying to install the x64 esbuild executable because the
-arm64 esbuild executable is not available yet. Running this executable requires
-the Rosetta 2 binary translator. Please make sure you have Rosetta 2 installed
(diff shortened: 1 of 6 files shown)


npm/esbuild-wasm/bin/esbuild
```diff
@@ -81,4 +81,4 @@ fs.read = function () {
 };
 
 const argv = ['node', wasm_exec, esbuild_wasm].concat(process.argv.slice(2));
-wrapper(require, require.main, Object.assign(Object.create(process), { argv }), { instantiate });
+wrapper(require, require.main, Object.assign(Object.create(process), { argv }), Object.assign(Object.create(WebAssembly), { instantiate }));
```
rtsao, line 84: As of Golang 1.16.0, the generated `run` function now has this check:
```js
			if (!(instance instanceof WebAssembly.Instance)) {
				throw new Error("Go.run: WebAssembly.Instance expected");
			}
```

This seemed like the simplest solution as opposed to patching the generated code like in https://github.com/evanw/esbuild/blob/add2790afd5f72b39c7eea9e48b2022e56f4cfe3/scripts/esbuild.js#L61

lib/install.ts
```diff
@@ -90,14 +90,6 @@ function validateBinaryVersion(binaryPath: string): void {
...
   } catch (err) {
-    if (platformKey === 'darwin arm64 LE')
-      throw new Error(`${err && err.message || err}
-
-This install script is trying to install the x64 esbuild executable because the
-arm64 esbuild executable is not available yet. Running this executable requires
-the Rosetta 2 binary translator. Please make sure you have Rosetta 2 installed
-before installing esbuild.
-`);
     throw err;
```
lukewarlow, line 93: This try catch is now redundant.

evanw:
Thanks so much for putting this together.

However, it's missing the most important part: has anyone with a M1 processor tried this build command on Go 1.16.0 and checked that esbuild works? I don't want to ship a broken build if no one has tested it, but I don't have a M1 myself. A good check could be to build one of the benchmarks (e.g. `make bench-three-esbuild`).

I'm also curious to hear what the build times are like for native vs. Rosetta emulated. For example, what the build times are like for `make bench-three-esbuild` with `GOARCH=arm64` vs. `GOARCH=amd64`.

sod:
> I'm also curious to hear what the build times are

m1 macbook pro 8gb | <- left arm64 ... right amd64 ->
image

evanw:
> if I run the bench 10 times, the arm64 avg. at ~290ms and the amd64 at ~405ms

That's awesome. Thanks for checking.

muuvmuuv:
Adding my tests here with Mac mini M1 16 GB Ram but with Go native arm64 build:

|amd64|arm64|
|-|-|
|Bildschirmfoto 2021-02-18 um 18 48 57|Bildschirmfoto 2021-02-18 um 18 49 46|

Issue #265, opened 2020-07-16
TSX Loader fails to parse Generics for Arrow Functions


macarie:
Hi 🙈

I've found this bug/weird (not really) behaviour: when generics are being used with arrow functions while using the `tsx` loader, ESBuild will throw an error.

Reproducible with:

```js
import esbuild from 'esbuild'

const ts = `
export const useless = <T>(array: T[]): T[] => {
	return array
}
`


esbuild.transformSync(ts, {
	loader: 'tsx',
})

// => Fails with: <stdin>:3:1: error: Unexpected "return"
```

This seems to be a known issue with TypeScript as well (https://github.com/Microsoft/TypeScript/issues/4922), so I was a bit hesitant whether to report it or not.

Do you think that this can be handled somehow by ESBuild?

By the way, thank you for this amazing project 🙂

---

A workaround for anyone landing here:

```ts
export const useless = <T,>(array: T[]): T[] => {
	return array
}
```

```ts
export const useless = <T extends unknown>(array: T[]): T[] => {
	return array
}
```

evanw:
The `tsx` loader intentionally follows TypeScript's behavior since that's basically the "specification" for how TypeScript is supposed to work. So if that's the way the TypeScript compiler works, then esbuild should work the same way and this is not a bug in esbuild. If you want this behavior to be changed you'll have to get it changed in the official TypeScript compiler first, at which point it will become a bug in esbuild because they will behave differently.

I'm also not going to try to turn esbuild into a "better TypeScript" because it wouldn't do any good. The point of TypeScript is the types, and with esbuild you still have to run the official TypeScript compiler to do type checking. So even if esbuild could parse this syntax when TypeScript can't, that means you'd no longer be able to type check your source code.

macarie:
Understood, thank you anyway 👍🏼

PR #825, opened 2021-02-17
Add two test cases for tsx and generics interaction.

brandonbloom:
I ran in to a parser bug related to #265 and was going to help fix them, so added these two test cases. However, it turns out these were already fixed on master! Figured I'd share the test cases in case they are useful.


evanw:
Sure, sounds good 👍

PR #843, opened 2021-02-19
Serve single page applications

ephjos:
Introduces a boolean flag that changes the server to fallback to `/index.html` on unknown routes. Without this flag, some Single Page Applications will not work when navigating to a "fake" route. With this flag, the index page will always be served and the client-side router can be used to render a new page.

The change adds the `--servespa` flag to the CLI and adds members to both the `ServeOptions` and `apiHandler` structs. At the end of the `ServeHTTP` function, if all of the other fallback methods fail, the index page is manually loaded and sent to the client.

This small change will enable esbuild to function as a development server for SPA projects, which can greatly simplify the project setup.

### DIFF
--- a/cmd/esbuild/main.go
+++ b/cmd/esbuild/main.go
@@ -77,6 +77,8 @@ var helpText = func(colors logger.Colors) string {
   --resolve-extensions=...  A comma-separated list of implicit extensions
                             (default ".tsx,.ts,.jsx,.mjs,.cjs,.js,.css,.json")
   --servedir=...            What to serve in addition to generated output files
+  --servespa                Serve a Single Page Application by falling back to
+                            index.html on unknown route
   --sourcefile=...          Set the source file for the source map (for stdin)
   --sourcemap=external      Do not link to the source map with a comment
   --sourcemap=inline        Emit the source map with an inline data URL
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -362,6 +362,7 @@ type ServeOptions struct {
 	Port      uint16
 	Host      string
 	Servedir  string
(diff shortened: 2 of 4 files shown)


evanw:
I think this is too specific for esbuild. There are many possible "almost like esbuild but with one small change" features that could be added to the local web server, but I don't think they all belong in esbuild. The local server is meant to be a composable primitive that you can make use of in your application. It does the hard part of incremental builds but other than that you can customize it by composing it.

For example, one way you can do this is to forward HTTP requests from another server. I think you can do this with a few lines of JavaScript:

```js
const http = require('http');
http.createServer((req, res) => {
  const proxyReq = http.request({
    hostname: 'localhost',
    port: 8000,
    path: req.url,
    method: req.method,
    headers: req.headers,
  }, proxyRes => {
    if (proxyRes.statusCode === 404) {
      res.writeHead(200, { 'Content-Type': 'text/html' });
      res.end('<h1>Some fallback html page</h1>');
    } else {
      res.writeHead(proxyRes.statusCode, proxyRes.headers);
      proxyRes.pipe(res, { end: true });
    }
  });
  req.pipe(proxyReq, { end: true });
}).listen(3000);
```

This code has barely been tested but it seems to do the trick. You'd set esbuild to listen on port 8000 while loading port 3000 in the browser.

ephjos:
I understand and appreciate your commitment to keeping this project acutely focused on your goals. The proxy is a cool approach and for what it's worth I am currently running esbuild in watch mode concurrently with [serve](https://www.npmjs.com/package/serve), which works quite well. 

This is a very useful tool and interesting piece of software, so thank you for your work and your time!

PR #872, opened 2021-02-22
Fix data race: collect results, then append.

### LINKED ISSUE #871: Data race in tests
Observed in my workstation Linux amd64, with all versions of Go1.13+
```
$ go test -race github.com/evanw/esbuild/internal/bundler
==================
WARNING: DATA RACE
Read at 0x00c0001247e0 by goroutine 20:
  github.com/evanw/esbuild/internal/bundler.(*scanner).preprocessInjectedFiles.func2()
      esbuild/internal/bundler/bundler.go:1085 +0x75

Previous write at 0x00c0001247e0 by goroutine 63:
  github.com/evanw/esbuild/internal/bundler.(*scanner).preprocessInjectedFiles()
      esbuild/internal/bundler/bundler.go:1078 +0x13f7
  github.com/evanw/esbuild/internal/bundler.ScanBundle()
      esbuild/internal/bundler/bundler.go:891 +0x4f6
  github.com/evanw/esbuild/internal/bundler.(*suite).expectBundled.func1()
      esbuild/internal/bundler/bundler_test.go:96 +0x597
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Goroutine 20 (running) created at:
  github.com/evanw/esbuild/internal/bundler.(*scanner).preprocessInjectedFiles()
      esbuild/internal/bundler/bundler.go:1084 +0x15b1
  github.com/evanw/esbuild/internal/bundler.ScanBundle()
      esbuild/internal/bundler/bundler.go:891 +0x4f6
  github.com/evanw/esbuild/internal/bundler.(*suite).expectBundled.func1()
      esbuild/internal/bundler/bundler_test.go:96 +0x597
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Goroutine 63 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1239 +0x5d7
  github.com/evanw/esbuild/internal/bundler.(*suite).expectBundled()
      esbuild/internal/bundler/bundler_test.go:84 +0x1ae
  github.com/evanw/esbuild/internal/bundler.TestInject()
      esbuild/internal/bundler/bundler_default_test.go:3393 +0x8bc
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202
==================
--- FAIL: TestInject (0.00s)
    --- FAIL: TestInject/#00 (0.00s)
        testing.go:1093: race detected during execution of test
    testing.go:1093: race detected during execution of test
FAIL
FAIL	github.com/evanw/esbuild/internal/bundler	1.317s
FAIL
```

Deleplace:
Fixes #871

I preserved the apparent semantics, i.e. the order in which results are appended to `InjectedFiles`.

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -1057,6 +1057,8 @@ func (s *scanner) preprocessInjectedFiles() {
 		go func() { s.resultChannel <- result }()
 	}
 
+	results := make([]config.InjectedFile, len(s.options.InjectAbsPaths))
+	j := 0
 	for _, absPath := range s.options.InjectAbsPaths {
 		prettyPath := s.res.PrettyPath(logger.Path{Text: absPath, Namespace: "file"})
 		lowerAbsPath := lowerCaseAbsPathForWindows(absPath)
@@ -1074,20 +1076,22 @@ func (s *scanner) preprocessInjectedFiles() {
 			continue
 		}
 
-		i := len(injectedFiles)
-		injectedFiles = append(injectedFiles, config.InjectedFile{})
 		channel := make(chan config.InjectedFile)
 		s.maybeParseFile(*resolveResult, prettyPath, nil, logger.Range{}, nil, inputKindNormal, channel)
 
 		// Wait for the results in parallel
+		// The results slice is large enough, it is not reallocated during the computations
 		injectWaitGroup.Add(1)
-		go func(i int, prettyPath string, resolveResult *resolver.ResolveResult) {
-			injectedFiles[i] = <-channel
+		go func(i int) {
+			results[i] = <-channel
 			injectWaitGroup.Done()
-		}(i, prettyPath, resolveResult)
+		}(j)
+		j++
 	}
 
 	injectWaitGroup.Wait()
+	injectedFiles = append(injectedFiles, results[:j]...)
+
 	s.options.InjectedFiles = injectedFiles
 }
 


kzc:
Good stuff. Until this issue I didn't know that Go has a built in race detector. Might as well use it in testing by default.

```patch
--- a/Makefile
+++ b/Makefile
@@ -22,3 +22,3 @@ check-go-version:
 test-go:
-       go test ./internal/...
+       go test -race ./internal/...
```

Deleplace:
Yes, and the race detector is awesome!

To be clear, it doesn't do any static analysis of the code. It compiles a special instrumented version of the package, which keeps track at runtime of all of the read and write memory accesses of the [current execution](https://en.wikipedia.org/wiki/Tracing_(software)). Then when the program is run (or when the tests are run) and a violation of the memory model happens, it detects it and reports it.

kzc:
With this PR and the patch above I'm still seeing a test failure due to a race condition:
```
$ make clean test
...
fatal error: too many address space collisions for -race mode
...
FAIL	github.com/evanw/esbuild/internal/css_parser	0.017s
```
I'm using:
```
go version go1.15.7 darwin/amd64
```

kzc:
Just to be clear, this error no longer occurs on my machine with this PR:
```
FAIL    github.com/evanw/esbuild/internal/bundler       3.027s
```

Deleplace:
The race detector does incur a large overhead in memory and compute. It may make sense to use `-race` in the Makefile, and that should be discussed in its own issue.

The data we have about the current PR is positive: sensible diff, fixes the problem on my machine, passes the CI tests. The current PR doesn't add any automated use of the race detector.

kzc:
@Deleplace So you don't see the internal/css_parser failure in https://github.com/evanw/esbuild/pull/872#issuecomment-783554832? Would you mind sharing the output of `go version` on your machine?

kzc:
The fatal error I'm seeing in `go test -race ./internal/css_parser` is a known issue on the older now unsupported macos 10.10, although it was a supported OS version at the time of the following bug report in 2018:

https://github.com/golang/go/issues/26475#issuecomment-427091408

The stack trace in the link is identical to the one on my machine.

evanw:
Thanks so much for finding this, and for the fix. Looks good to me.

evanw:
Wait actually now that I look at this more, I'm confused about why this is an issue. Sorry for not taking a closer look first. The array should already be big enough since there is this at the top:

```go
injectedFiles := make([]config.InjectedFile, 0, len(s.options.InjectedDefines)+len(s.options.InjectAbsPaths))
```

The fix in this PR isn't quite right because the `continue` statements in the loop now leave holes in the array.

Deleplace:
For the "why was this an issue" in the first place:
- see [comment #871](https://github.com/evanw/esbuild/issues/871#issuecomment-783469206): we were doing several hazardous things concurrently
- the race detector complained, and it is designed to not report any false positive
- I understand the rationale "in this case there should not be any reallocation", however it is still a memory model violation to be doing `files = append(files, f)` and `files[i] = <-channel` concurrently. This is not just theoretical, the compiler does many weird things with the assumption that the code is not racy, and bad things may happen anytime, today, or next month.

Deleplace:
For the "leave holes in the array", I don't think we have them. Consider:
- a new distinct slice `results` which has fixed length M = `len(s.options.InjectAbsPaths)`
- `j` is incremented only when 1 extra computation has just been launched
- after all the computations have finished, only the sublist `results[:j]` is used
- we alter `injectedFiles` only once, after all the computations have finished to write to `results[:j]`

evanw:
Ah right there are no holes because of the `results[:j]`. 👍

Issue #789, opened 2021-02-11
Different strategy for installing platform-specific binaries


eduardoboucas:
Hi! 👋🏻 

We're looking to integrate esbuild with the [Netlify CLI](https://www.npmjs.com/package/netlify-cli) and some users are [reporting permission-related errors](https://github.com/netlify/cli/issues/1870) when installing the application, which trace back to https://github.com/evanw/esbuild/issues/369.

I understand that this is a result of the UID/GID switching that npm does when running scripts, which is problematic when the package is installed as root.

Rather than manually fetching the appropriate binary as part of the `postinstall` script, could the `esbuild` package list the various [platform-specific packages](https://github.com/evanw/esbuild/tree/master/npm) as `optionalDependencies`, and shift to npm itself the responsibility of figuring out which packages to installed based on the `cpu` and `os` properties? We're doing this for one of our packages that also includes a Go binary and it works well.

I'm happy to contribute with a pull request, but I wanted to check whether this is something you'd be interested in, or even if it's something you've already considered and discarded for a particular reason.

Thanks in advance!

evanw:
> could the `esbuild` package list the various [platform-specific packages](https://github.com/evanw/esbuild/tree/master/npm) as `optionalDependencies`

I have considered implementing the install script this way but I didn't go down that route because if I recall correctly it makes the installation output really noisy with all of the errors about unmet optional dependencies, which is a poor user experience. I'm also not totally sure what the guarantees around optional dependencies are (since I need them to be required, not optional) and if there are any situations where that could backfire and leave you with a broken package. Especially with all of the various alternate package managers other than npm that people like to use. Having my own install script gives me more control over these things.

I'd really rather not change the whole approach of the install script at this point because it's undergone a lot of evolution and has accumulated a lot of special cases that solve real problems. It's relatively stable at this point and changing it would risk breaking people, and would likely be a ton of work for me since I'd expect dozens of support requests from the fallout. It's at least not something I should do without a breaking change in the version number.

Depending on what you're doing, one possibility could be for you to just make your own package with all of esbuild's platform-specific packages as optional dependencies as you suggested. That should let you do this without any changes to esbuild itself. Doing this would at least be a less disruptive way of trying out the approach than involving esbuild's whole user base in the experiment.

eduardoboucas:
Thanks for the additional context, @evanw. I completely understand your position.

[...]

PR #881, opened 2021-02-24
Add ESBUILD_BIN_PATH

sushain97:
Installation already appears to support a couple modes:
- `ESBUILD_BIN_PATH_FOR_TESTS` which symlinks an `esbuild` binary into `node_modules`
- `installBinaryFromPackage` which either installs from a local cache or downloads the binary from `npm` (directly or via `npm`)

However, this doesn't match our needs exactly. In particular, we run `yarn install` in `--offline` mode inside a micro VM without network access. Thus, we need a way to tell the `esbuild` installer about a binary that we've already downloaded. Since we need the `node_modules` to be standalone, `ESBUILD_BIN_PATH_FOR_TESTS` doesn't work since it creates a symlink.

We handle a similar issue with `node-sass` by using `npm_config_sass_binary_cache`. However, the local caching logic in `esbuild` is a little more involved and I'm loathe to attempt replicating it our end (especially since we support both `darwin` and `linux` -- see the hack mentioned later).

So, there are two potential solutions:
- Support `ESBUILD_CACHE_PATH` which overrides the cache directly
- Support `ESBUILD_BIN_PATH` which acts just like `ESBUILD_BIN_PATH_FOR_TESTS` but copies the file instead

I decided on the latter since that resolves the problem directly (although I do need to test this change with our build system... opening the PR early to see if there's any appetite for this tweak).

For completeness, my current hack is:
```
mkdir -p ~/.cache/esbuild/bin/
cp external/esbuild/bin/esbuild ~/.cache/esbuild/bin/esbuild-linux-64@0.8.34
```
(within the `yarn install` Bazel action)

### DIFF
--- a/lib/install.ts
+++ b/lib/install.ts
@@ -221,7 +221,10 @@ function installDirectly(name: string) {
   if (process.env.ESBUILD_BIN_PATH_FOR_TESTS) {
     fs.unlinkSync(binPath);
     fs.symlinkSync(process.env.ESBUILD_BIN_PATH_FOR_TESTS, binPath);
-    validateBinaryVersion(process.env.ESBUILD_BIN_PATH_FOR_TESTS);
+    validateBinaryVersion(binPath);
+  } else if (process.env.ESBUILD_BIN_PATH) {
+    fs.copyFileSync(process.env.ESBUILD_BIN_PATH, binPath);
+    validateBinaryVersion(binPath);
   } else {
     installBinaryFromPackage(name, 'bin/esbuild', binPath)
       .catch(e => setImmediate(() => { throw e; }));


evanw:
I'm open to doing something like this but I'd like to reserve the right to break hacks like this this in the future. This is a super specific request and the implementation details of the installer isn't something I intended to be part of esbuild's public API. For example, in the future the installer may be changed to an entirely different approach such as #789, at which point anything relying on this will break.

sushain97:
That's fair! I'm sure we can adjust our wrappers accordingly if it breaks.

While this use case is certainly rare, it's definitely something which folks have invested in: https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/. The current installation setup doesn't allow for offline installs unless you guess the cache path.

My skimming of #789 suggests that if `esbuild` switched to an approach where the package manager itself downloads the deps, the downloads would be no different from any other `npm` module and everything would Just Work.

evanw:
FYI I'm going to change `ESBUILD_BIN_PATH` to `ESBUILD_BINARY_PATH` since that is already supported for overriding the binary path of the JavaScript API code at run-time (see #597).

sushain97:
> FYI I'm going to change `ESBUILD_BIN_PATH` to `ESBUILD_BINARY_PATH` since that is already supported for overriding the binary path of the JavaScript API code at run-time (see #597).

👍 Thanks for merging!

PR #882, opened 2021-02-25
fix: ts class with decorator should use var

echosoar:
A class has a Decorator:
```test.ts
@Decoratr('test')
export class Test {
  
}
```
execute:
```sh
esbuild test.ts --target=node15 --platform=node --format=cjs
```

output:
```js
// output
__markAsModule(exports);
__export(exports, {
  Test: () => Test // variable cannot be used here
});
let Test = class {
};
Test = __decorate([
  Decoratr("test")
], Test);
```

result should be like this:
```js
// output
__markAsModule(exports);
__export(exports, {
  Test: () => Test
});
var Test = class {
};
Test = __decorate([
  Decoratr("test")
], Test);
```

### DIFF
--- a/internal/js_parser/js_parser_lower.go
+++ b/internal/js_parser/js_parser_lower.go
@@ -2160,7 +2160,7 @@ func (p *parser) lowerClass(stmt js_ast.Stmt, expr js_ast.Expr, shadowRef js_ast
 
 		// Generate the variable statement that will represent the class statement
 		stmts = append(stmts, js_ast.Stmt{Loc: classLoc, Data: &js_ast.SLocal{
-			Kind:     p.selectLocalKind(js_ast.LocalLet),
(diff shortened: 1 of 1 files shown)


evanw:
The choice of `let` here is deliberate, and is required to preserve the semantics of `class` which behaves like `let`, not like `var`. This PR doesn't motivate why you are changing this.

I assume you're doing this because you're having some problem with esbuild. In that case, the more productive thing to do would be to file an issue about your specific problem.

echosoar:
> The choice of `let` here is deliberate, and is required to preserve the semantics of `class` which behaves like `let`, not like `var`. This PR doesn't motivate why you are changing this.
> 
> I assume you're doing this because you're having some problem with esbuild. In that case, the more productive thing to do would be to file an issue about your specific problem.

Thanks for your reply, I described the issue under my problems later on.

maysupan:
Thank

PR #902, opened 2021-03-01
feat: enabled commonjs code splitting

jacob-ebey:
Enabled commonjs code splitting. Looks like all the existing tests are in-tact.

Are there any tests that run against the generated output to verify functionality?

### DIFF
--- a/internal/bundler/bundler_splitting_cjs_test.go
+++ b/internal/bundler/bundler_splitting_cjs_test.go
@@ -1,0 +1,527 @@
+package bundler
+
+import (
+	"testing"
+
+	"github.com/evanw/esbuild/internal/config"
+)
+
+var splittingSuiteCJS = suite{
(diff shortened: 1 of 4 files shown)


evanw:
I want to let you know that I won't be accepting this PR. I am already in the middle of a major rewrite of the linker. The rewrite is to address fundamental import order issues with code splitting as well as add new features (including code splitting for CommonJS). The problem is much deeper than it may first appear. For example, exports between chunks need to be live references instead of dead copies the way you've done it in this PR. There is an extraordinary amount of context required to do such a big rewrite so for efficiency's sake I'm not looking for someone else to work on this with.

jacob-ebey:
@evanw Fair enough. Would love to follow the work if you could do it in public.

talentlessguy:
Probably worth closing the PR?

PR #941, opened 2021-03-10
Bound stdout shifting to prevent excessive allocations

jridgewell:
I noticed this when using an incremental rebuild. In my project, the `stdout` can grow to 100mb+ during the initial build. During rebuild, the `stdout` is mostly empty. But the old code used an unbounded `slice`, which allocated and shifted almost the full 100mb.


maysupan:
Thank you.

evanw:
Good catch! I had intended to use `subarray` here instead of `slice`. Thanks for fixing this.

PR #967, opened 2021-03-13
Avoid stdout buffer allocation

jridgewell:
Your [comment](https://github.com/evanw/esbuild/pull/941#issuecomment-795118580) reminded me to fix this one, too.

### DIFF
--- a/lib/common.ts
+++ b/lib/common.ts
@@ -434,7 +434,7 @@ export function createChannel(streamIn: StreamIn): StreamOut {
         break;
       }
       offset += 4;
-      handleIncomingPacket(stdout.slice(offset, offset + length));
+      handleIncomingPacket(stdout.subarray(offset, offset + length));
       offset += length;
     }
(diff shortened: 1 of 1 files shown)


maysupan:
Thank!

lib/common.ts
```diff
@@ -434,7 +434,7 @@ export function createChannel(streamIn: StreamIn): StreamOut {
         break;
       }
       offset += 4;
-      handleIncomingPacket(stdout.slice(offset, offset + length));
+      handleIncomingPacket(stdout.subarray(offset, offset + length));
```
evanw, line 437: One concern here is that something later on would save a copy of some subarray of this subarray, which could then get overwritten later. But I checked the decoding code and it appears to always make a copy, so this should be ok I think:

https://github.com/evanw/esbuild/blob/afa91d7dd3e235578f0500b66aab26796691cf18/lib/stdio_protocol.ts#L316-L322

PR #970, opened 2021-03-13
attempt to fix pnpm hard link issues

### LINKED ISSUE #963: compatible problem with pnpm when multiple esbuild version exists
It seems that pnpm will do some special tricks to bin files, which will cause esbuild fails when multiple esbuild version exisits,(it may be a bug of pnpm https://github.com/pnpm/pnpm/issues/3238), but It's kind of unusual to make js entry files rely on bin files(it's usual that bin file relys on main js entry files), maybe not put the real bin into bin folder could solve the issue.

evanw:
Fixes #963.


lib/install.ts
```diff
@@ -222,7 +222,13 @@ function installDirectly(name: string) {
...
     validateBinaryVersion(binPath);
   } else {
-    installBinaryFromPackage(name, 'bin/esbuild', binPath)
+    // Write to a temporary file, then move the file into place. This is an
+    // attempt to avoid problems with package managers like pnpm which will
+    // usually turn each file into a hard link. We don't want to mutate the
+    // hard-linked file which may be shared with other files.
+    const tempBinPath = binPath + '__';
+    installBinaryFromPackage(name, 'bin/esbuild', tempBinPath)
+      .then(() => fs.renameSync(tempBinPath, binPath))
```
merceyz, line 231: Do note you can't rename across drives, it will throw `EXDEV` so you need to copy instead

merceyz: I didn't notice the temp path was right next to it in the same folder, nvm

PR #983, opened 2021-03-16
fix comments

Zuckjet:
just a small fix for comment typo


evanw:
Thanks for catching this.

Issue #988, opened 2021-03-17
[bug] "Transforming destructuring is not supported yet" since 0.9.3


jurijzahn8019:
Hi,

since 0.9.3 i get this error. Works in 0.9.2

```plaintext
> vite -c vite.config.ts build

vite v2.1.1 building for production...
✓ 114 modules transformed.
error during build:
Error: Transform failed with 4 errors:
assets/index.5b8f7deb.js:4:6: error: Transforming destructuring to the configured target environment ("chrome61", "edge16", "es2019", "firefox60", "safari11") is not supported yet
assets/index.5b8f7deb.js:6:8: error: Transforming destructuring to the configured target environment ("chrome61", "edge16", "es2019", "firefox60", "safari11") is not supported yet
assets/index.5b8f7deb.js:7:8: error: Transforming destructuring to the configured target environment ("chrome61", "edge16", "es2019", "firefox60", "safari11") is not supported yet
assets/index.5b8f7deb.js:15:8: error: Transforming destructuring to the configured target environment ("chrome61", "edge16", "es2019", "firefox60", "safari11") is not supported yet
    at failureErrorWithLog ( <root_dir>/node_modules/esbuild/lib/main.js:1195:15)
    at <root_dir>/node_modules/esbuild/lib/main.js:1056:33
    at <root_dir>/node_modules/esbuild/lib/main.js:566:9
    at handleIncomingPacket ( <root_dir>/node_modules/esbuild/lib/main.js:655:9)
    at Socket.readFromStdout ( <root_dir>/node_modules/esbuild/lib/main.js:533:7)
    at Socket.emit (events.js:315:20)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9)
    at Socket.Readable.push (internal/streams/readable.js:223:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:188:23)
```

zhangshuaidan:
I meet the same error, here is detail info
`assets/vendor.93f489e5.js:629:22: error: Transforming destructuring to the configured target environment ("chrome61", "edge16", "es2019", "firefox60", "safari11") is not supported yet
...
    at failureErrorWithLog (D:\dev\vue\vite-project\node_modules\esbuild\lib\main.js:1195:15)
    at D:\dev\vue\vite-project\node_modules\esbuild\lib\main.js:1056:33
[...]

PR #990, opened 2021-03-17
fix: modify Edge version for destructuring

Zuckjet:
[Fix](https://github.com/evanw/esbuild/issues/988)

From MDN [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment),  Edge support `Destructuring assignment` from version 14.

### DIFF
--- a/internal/compat/js_table.go
+++ b/internal/compat/js_table.go
@@ -203,7 +203,7 @@ var jsTable = map[JSFeature]map[Engine][]int{
 	},
 	Destructuring: {
 		Chrome:  {51},
-		Edge:    {18},
+		Edge:    {14},
 		ES:      {2015},
 		Firefox: {53},
 		IOS:     {10},


Zuckjet:
> Raise certain VM versions in the JavaScript feature compatibility table

JavaScript VM feature compatibility data is derived from this dataset: https://kangax.github.io/compat-table/. The scripts that process the dataset expand the data to include all VM versions that support a given feature (e.g. chrome44, chrome45, chrome46, ...) so esbuild takes the minimum observed version as the first version for which the feature is supported.

However, some features can have subtests that each check a different aspect of the feature. In this case the desired version is the minimum version within each individual subtest, but the maximum of those versions across all subtests (since esbuild should only use the feature if it works in all cases). Previously esbuild computed the minimum version across all subtests, but now esbuild computes the maximum version across all subtests. This means esbuild will now lower JavaScript syntax in more cases.

From change log, I know why you would like to raise VM version. But it seems that raising VM version will bring breaking changes to other tools built by esbuild

evanw:
Heads up that right now I'm not planning to land this because doing this could lead to correctness bugs. More context: https://github.com/evanw/esbuild/issues/988#issuecomment-801315072.

Zuckjet:
Nice explanation, I will close this.

PR #1023, opened 2021-03-20
implement lowering for-of loops

DevSide:
Lowering for-of loops into regular for loops.

@evanw, would you like I infer the iterable identifier type then inject a runtime iterator helper ? 
Before going further, I need to know if it's the way to go.


evanw:
Yes unfortunately implementing this correctly requires emulating all of the specifics of iterators, which are somewhat complex.

PR #1028, opened 2021-03-21
Add Source Map sourceRoot Support

### LINKED ISSUE #675: Relative source map file paths not found with vscode debugger
I'm currently debugging why breakpoints in vscode are not found with our testing setup over at [Preact](https://github.com/preactjs/preact). For that each test file is a fully self-contained bundle and our test runner serves those files relative to the server's process root:

```bash
# Folder structure
/my-project
  /src
    source.js
  /tests
    test1.js   # imports src/source.js
    /sub
      test2.js # imports src/source.js
```

How the files are served by the test runner:

```txt
/base/test1.js
/base/sub/test2.js
```

Now onto the source map resolution. The [spec](https://sourcemaps.info/spec.html#h.75yo6yoyk7x5) has this to say about resolution:

> If the sources are not absolute URLs after prepending of the “sourceRoot”, the sources are resolved relative to the SourceMap (like resolving script src in a html document).

The source maps generated by esbuild look like this for me:

```json
{
  "version": 3,
  "sources": ["src/source.js", "..."],
  "mappings": "...",
  "sourcesContent": [...]
} 
```

They don't have a `sourceRoot` property and therefore the source maps are resolved relative to the js file.

```txt
/base/test1.js -> /base/tests/test1.js
                  /base/src/source.js
/base/sub/test2.js -> /base/sub/tests/sub/test2.js
                      /base/sub/src/source.js
```

This leads to duplicate path segments as above. Here is a screenshot of what it looks like in Chrome's Devtools in Preact's suite:

Screenshot 2021-01-14 at 17 08 47

In the screenshot `compat/test/browser` is repeated twice. So when a user creates a breakpoint in vscode the requested paths don't match and therefore the breakpoint is not found.

The way webpack handles this is by using a custom protocol `webpack://` to force the resolution logic to always be absolute. The vscode debugger plugin has some [built-in mappings](https://github.com/microsoft/vscode-chrome-debug/blob/c9bb7d2d793eeeefe06535baa39a4a4620a8ff91/src/chromeDebugAdapter.ts#L25-L31) that map this to the projects working directory.

I don't think using something like `esbuild://` would be a good fit, but some form of absolute resolution is needed for a consistent experience. Maybe this can be done via `sourceRoot` by setting it to a value which can be easily mapped by debuggers.

jridgewell:
This adds the `--source-root`, `sourceRoot` (JS), and `SourceRoot` (Go) options to insert a `sourceRoot` field in any output source maps. When a `sourceRoot` exists on the map, any sources are resolved relative to it.

 This is particularly useful when hosting compiled code on a server, and pointing the source files to a GitHub repo. Eg, as [AMP does](https://cdn.ampproject.org/v0.mjs.map).

Fixes #675.


evanw:
Some thoughts:

* I think this should be `sourceRoot` instead of `sourcemapRoot`. This matches the name of the property in the source map, and [esbuild's `sourcesContent` setting](https://esbuild.github.io/api/#sources-content) is an existing precedent for these settings matching the name of the corresponding source map property.

* Right now the paths in the source map are relative to the [`absWorkingDir`](https://esbuild.github.io/api/#working-directory) setting. Another option could be to change this to be relative to the [`outbase`](https://esbuild.github.io/api/#outbase) setting instead. I'm wondering which one is more appropriate.

jridgewell:
> I think this should be sourceRoot instead of sourcemapRoot. 

Sure, I can make that change tonight.

> Right now the paths in the source map are relative to the absWorkingDir setting.

I actually like that it's relative to `absWorkingDir`, although setting it relative to could work. My experience is with Closure compiler, which uses the working directory.  Rollup seems to make everything relative to the directory of the output sourcemap (thought they provide a [function option](https://rollupjs.org/guide/en/#outputsourcemappathtransform) to transform it). Webpack seems to use working directory, though they use an absolute `webpack://package/PATH` URL.

When making npm packages, I generally don't include the `src` directory in the package's published files. So making the sources relative to the build directory isn't correct, I use the `sourceRoot` to point to the git code at the specific version tag.

Perhaps we could just create an option for `sourcemapOutbase`, and have it default to whichever of `absWorkingDir` or `outbase` you prefer?

jridgewell:
This has been updated to `--source-root`, `sourceRoot` (JS), and `SourceRoot` (Go).

evanw (approved):
Looks reasonable to me! Thanks for the PR.

PR #1036, opened 2021-03-23
Use WebAssembly.instantiateStreaming when available

Jarred-Sumner:
This makes `esbuild.initialize` automatically use `WebAssembly.instantiateStreaming` in supported browsers by default. 

For browsers that support `WebAssembly.instantiateStreaming` and if the `wasmURL` is same-origin, it adds an additional (default `true`) option, `verifyWasmURL` which sends a `HEAD` request to the `wasmUrl` to check that the `Content-Type` header is `application/wasm` and that the response is ok. If not, it fallsback to `WebAssembly.instantiate` and logs to `console.info`. **This makes it a backwards-compatible change**, incase a developer's webserver is not configured to correctly send the `application/wasm` header [such as nginx](https://github.com/nginx/nginx/blob/master/conf/mime.types) or if the `wasmUrl` is under a different origin (like a CDN).

To do this, it inlines `wasmUrl` into `global.ESBUILD_WASM_URL`. If `global.ESBUILD_WASM_URL` is truthy and `WebAssembly.instantiateStreaming` is defined, it uses that instead.

~I have not manually tested this, but the browser test failed until I made it check for `HEAD` requests to `/esbuild.wasm`, which is a good sign. I had some build environment issues I think due to having a later version of Go installed~

I added tests that cover:
- No `WebAssembly.instantiateStreaming` support
- `WebAssembly.instantiateStreaming` support, `verifyWasmURL` false
- `WebAssembly.instantiateStreaming` support, `verifyWasmURL` true
- `WebAssembly.instantiateStreaming` support, `verifyWasmURL` true, return wrong content type. It should fallback to `WebAssembly.instantiate`

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -56,14 +56,17 @@ export const initialize: typeof types.initialize = options => {
 
 const startRunningService = async (wasmURL: string, useWorker: boolean, verifyWasmURL: boolean ): Promise<void> => {
   let wasm: ArrayBuffer | void;
+  let url: URL;
   // Per https://webassembly.org/docs/web/#webassemblyinstantiatestreaming, 
   // Automatically use 'instantiateStreaming' when available and:
   // - The Response is CORS-same-origin
   // - The Response represents an ok status
   // - The Response Matches the `application/wasm` MIME type
-  if ('instantiateStreaming' in WebAssembly && new URL(wasmURL, location.href).origin === location.origin) {
-    if (verifyWasmURL) {
+  if ('instantiateStreaming' in WebAssembly && (url = new URL(wasmURL, location.href), url.origin === location.origin)) {
+    // If its a relative URL, it must be made absolute since the href of the worker might be a blob
+    wasmURL = url.toString();
 
+    if (verifyWasmURL) {
       const resp = await fetch(wasmURL, {
         method: "HEAD",
         // micro-optimization: try to keep the connection open for longer to reduce the added latency for fetching the WASM
@@ -76,7 +79,7 @@ const startRunningService = async (wasmURL: string, useWorker: boolean, verifyWa
         let res = await fetch(wasmURL);
         if (!res.ok) throw new Error(`Failed to download ${JSON.stringify(wasmURL)}`);
         // Log this after so they don't see two logs.
-        console.info(`Make esbuild-wasm load faster by setting the \"Content-Type\" header to \"application/wasm\" in \"${JSON.stringify(wasmURL)}\". Learn more at https://v8.dev/blog/wasm-code-caching#stream.`)
+        console.info(`Make esbuild-wasm load faster by setting the "Content-Type" header to "application/wasm" in ${JSON.stringify(wasmURL)}. Learn more at https://v8.dev/blog/wasm-code-caching#stream.`)
         wasm = await res.arrayBuffer();
       }
     }
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -56,7 +56,12 @@ export const initialize: typeof types.initialize = options => {
 
 const startRunningService = async (wasmURL: string, useWorker: boolean, verifyWasmURL: boolean ): Promise<void> => {
   let wasm: ArrayBuffer | void;
-  if ('instantiateStreaming' in WebAssembly) {
+  // Per https://webassembly.org/docs/web/#webassemblyinstantiatestreaming, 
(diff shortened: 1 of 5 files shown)


Jarred-Sumner:
Here are timings from running the tests:
```
[console.timeEnd] iife: 2943.001220703125 ms
[console.timeEnd] iifeStreaming: 2432.276123046875 ms
[console.timeEnd] iifeStreamingVerifyWasmURL: 2445.62109375 ms
[console.timeEnd] iifeStreamingVerifyWasmURLFallback: 2900.329833984375 ms
[console.timeEnd] iifeAsync: 2798.52294921875 ms
[console.timeEnd] iifeAsyncStreaming: 594.915771484375 ms
[console.timeEnd] iifeAsyncStreamingVerifyWasmURL: 643.916015625 ms
[console.timeEnd] iifeAsyncStreamingVerifyWasmURLFallback: 2827.716796875 ms
[console.timeEnd] iifeMin: 2899.09423828125 ms
[console.timeEnd] iifeMinStreaming: 2420.141845703125 ms
[console.timeEnd] iifeMinStreamingVerifyWasmURL: 2440.100341796875 ms
[console.timeEnd] iifeMinStreamingVerifyWasmURLFallback: 2905.22509765625 ms
[console.timeEnd] iifeMinAsync: 2837.287109375 ms
[console.timeEnd] iifeMinAsyncStreaming: 633.72802734375 ms
[console.timeEnd] iifeMinAsyncStreamingVerifyWasmURL: 632.817138671875 ms
[console.timeEnd] iifeMinAsyncStreamingVerifyWasmURLFallback: 2830.69580078125 ms
[console.timeEnd] esm: 2863.778076171875 ms
[console.timeEnd] esmStreaming: 2400.2158203125 ms
[console.timeEnd] esmStreamingVerifyWasmURL: 2437.08984375 ms
[console.timeEnd] esmStreamingVerifyWasmURLFallback: 2897.178955078125 ms
[console.timeEnd] esmAsync: 2815.09912109375 ms
[console.timeEnd] esmAsyncStreaming: 619.460205078125 ms
[console.timeEnd] esmAsyncStreamingVerifyWasmURL: 613.30810546875 ms
[console.timeEnd] esmAsyncStreamingVerifyWasmURLFallback: 2831.27587890625 ms
[console.timeEnd] esmMin: 2890.48291015625 ms
[console.timeEnd] esmMinStreaming: 2478.572998046875 ms
[console.timeEnd] esmMinStreamingVerifyWasmURL: 2432.202880859375 ms
[console.timeEnd] esmMinStreamingVerifyWasmURLFallback: 2888.947021484375 ms
[console.timeEnd] esmMinAsync: 2788.493896484375 ms
[console.timeEnd] esmMinAsyncStreaming: 622.496337890625 ms
[console.timeEnd] esmMinAsyncStreamingVerifyWasmURL: 697.15185546875 ms
[console.timeEnd] esmMinAsyncStreamingVerifyWasmURLFallback: 2822.542724609375 ms
```

This is with the cache disabled, 30mbps down, 15mbps up, and 20ms latency: (my home internet is 100mbps down)
```
     // Connect to Chrome DevTools
          const client = await page.target().createCDPSession()

          // Set throttling property
          await client.send('Network.emulateNetworkConditions', {
            'offline': false,
            // 30 mbps down
            'downloadThroughput': 3932160,
            // 15 mbps up
            'uploadThroughput': 15728640,
            // Little latency
            'latency': 20,
            
          })
          await client.send("Network.setCacheDisabled", {cacheDisabled: true});
```

I'm a little suspicious of whether or not the `Network.setCacheDisabled` actually works for this case.

If you'd like, I can commit the code that adds these timings (its wrapped in a boolean so it can be easily turned off)

evanw:
I do think this change should be done (I even have a draft of this change somewhere), and I appreciate the thoughtfulness that went into this. However, I've been thinking that this should be done by instead just trying `WebAssembly.instantiateStreaming` and then automatically falling back to `WebAssembly.instantiate` if that fails before finally giving up.

My thinking is that not serving WebAssembly with the right MIME type is really a bug on the server, albeit one with a graceful fallback, and is somewhat of an exceptional situation that shouldn't be catered to for performance. The default behavior should assume that everything is set up correctly for maximum speed.

So the way I'd like to see this implemented is a) there is no configuration option to change behavior and b) `WebAssembly.instantiateStreaming` is automatically tried first. In that case I think the main thread doesn't need to perform the fetch at all. The worker thread will need to handle errors and post any error back to the main thread, however.

PR #1054, opened 2021-03-25
Avoid a buffer copy when decoding utf-8 in node 10

jridgewell:


lib/stdio_protocol.ts
```diff
@@ -352,7 +352,7 @@ else if (typeof Buffer !== 'undefined') {
 
     return buffer;
   };
-  decodeUTF8 = bytes => Buffer.from(bytes).toString();
+  decodeUTF8 = bytes => Buffer.from(bytes.buffer).toString();
```
evanw, line 355: These aren't equivalent:

```js
var array = new Uint8Array([51, 52, 53, 54, 55])
var slice = array.subarray(1, 4)
console.log({
  a: Buffer.from(slice).toString(),
  b: Buffer.from(slice.buffer).toString(),
})
```

This prints `{ a: '456', b: '34567' }`.

jridgewell: Ah, I didn't think we called this with anything but a full-view `Uint8Array`. I can update to use `byteOffset` and `byteLength` for decoding.

Zuckjet: @jridgewell I don't understand why this change will **avoid buffer copy**, can you give  an explanation ?

jridgewell: See https://nodejs.org/api/buffer.html#buffer_static_method_buffer_from_arraybuffer_byteoffset_length:

> This creates a view of the ArrayBuffer without copying the underlying memory. For example, when passed a reference to the .buffer property of a TypedArray instance, the newly created Buffer will share the same allocated memory as the TypedArray's underlying ArrayBuffer.

Passing an `ArrayBuffer` means the new `Buffer` instance will be backed by the same memory. Passing an array-like forces a the `Buffer` to allocate a contiguous memory block, and it then has to iterate over the array to copy the data to the new block.

PR #1067, opened 2021-03-27
Add warning when writing to private method

jridgewell:
Regrettably, private methods aren't writable. The downleveling seems correct, but I noticed that there wasn't a warning like the getter-only/setter-only private accessors.


evanw:
This is a good addition. Thanks.

Issue #399, opened 2020-09-20
Incorrect import order with code splitting and multiple entry points


iamakulov:
Consider the following code:

```js
// entry1.js
import "./init-dep-1.js";
import "./run-dep.js";

// entry2.js
import "./init-dep-2.js";
import "./run-dep.js";

// init-dep-1.js
global.foo = {
  log: () => console.log("foo.log() (from entry 1) called"),
};

// init-dep-2.js
global.foo = {
  log: () => console.log("foo.log() (from entry 2) called"),
};

// run-dep.js
global.foo.log();
```

When you bundle this code with `esbuild --bundle --format=esm --splitting ./entry1.js ./entry2.js`, ESBuild discovers that `run-dep.js` is a module shared between both entry points. Because code splitting is enabled, ESBuild moves it into a separate chunk:

```js
// build/entry1.js
import "./chunk.P2RPFHF3.js";

// init-dep-1.js
global.foo = {
  log: () => console.log("foo.log() (from entry 2) called")
};

// ../../../tmp/esbuild-code-splitting/entry1.js
```

However, by doing so, ESBuild puts `run-dep.js` above the `init-dep-1.js` or `init-dep-2.js` code. This changes the import order – and breaks the code:

```sh
$ esbuild --bundle --outdir=build --format=esm --splitting --platform=node --out-extension:.js=.mjs ./entry1.js ./entry2.js
$ node build/entry1.mjs
global.foo.log();
           ^

TypeError: Cannot read property 'log' of undefined
```

evanw:
As far as I know the execution order of imported ECMAScript modules is undefined (see also [this question](https://stackoverflow.com/questions/35551366)). The run-time is free to, for example, asynchronously download all imports in parallel and evaluate them in the order that their downloads finish, which could be different on different runs. The only guarantee is that all imports will be evaluated before any code in the script doing the importing is evaluated.

Using ECMAScript module imports means having to write your code to be robust to different import execution orders. The bundling algorithm in esbuild takes advantage of this flexibility to do the above transformation.

If you need deterministic order of imported code, you can either use `require()` instead of `import` or use `import` but defer the execution of certain code by wrapping it in a function and calling it later.

guybedford:
Module evaluation in ECMA-262 is synchronous from beginning to end. It is defined strictly as a post-order graph execution in https://tc39.es/ecma262/#sec-innermoduleevaluation. There isn't any room for misinterpretation of this and no JS engines violate this behaviour.

[...]

Issue #253, opened 2020-07-12
support for top level await


buttercubz:
I am currently working with deno which has top level support, is there any way to use esbuild with top level await?

evanw:
Sorry, top-level await is not supported. It messes with a lot of things and adding support for it is quite complicated. It likely won't be supported for a long time.

guybedford:
Out of interest / for the issue history, is it possible to give some idea
of the sort of issues that arise here?

evanw:
Right now there is an equivalence in esbuild between CommonJS modules and ES6 modules. A module can be written in either syntax and can even mix both syntaxes. That no longer works with CommonJS because the implementation of `require()` can't suspend execution in JavaScript. So I'll have to figure out something for that. It's probably some form of conditionally disabling features along the module dependency graph when a module contains a top-level await.

I'd like to punt this feature until later because there are already a lot of edge cases to work through with all of the features added so far (tree shaking, scope hoisting, code splitting, star re-exports, etc.) and this feature adds another dimension to the test case matrix. This feature is also lower priority for me than other features such as watch mode, plugins, finishing code splitting, and adding additional CSS and HTML file types, especially given that top-level await is a new feature and not yet widely used.

Let's keep this issue open because I'm sure it'll come up again. At the very least esbuild should be able to parse top-level await and pass it through in non-bundling mode, which will be useful for people using esbuild as a transformer and not a bundler.

I haven't added parsing support for top-level await yet because it's not clear exactly how to parse it. The way it works the JavaScript spec is that you're supposed to know beforehand whether the parsing goal is "script" or "module" and then only parse the `await` operator with the "module" goal. However, esbuild doesn't work that way because it makes no distinction between the two goals.

TypeScript is another tool that does this so I've been waiting to see how they handle things. When I looked into this last they had a [pretty severe bug](https://github.com/microsoft/TypeScript/issues/39186) so I'm going to wait a while longer for their implementation to settle before I consider how to approach this in esbuild. It looks like they still have open bugs on this such as [this one](https://github.com/microsoft/TypeScript/pull/38518), which interestingly appears to require a symbol table to parse correctly kind of like C. Not sure how I feel about that approach.

Also, while I'm on the topic, apparently this feature also [requires supporting top-level `for await` loops](https://github.com/tc39/proposal-top-level-await/issues/132). Just adding that as a note to myself for when I get to implementing this.

guybedford:
[...]

Issue #207, opened 2020-06-30
[Feature] Manual chunks


intrnl:
Similar to Rollup's [`manualChunks`](https://rollupjs.org/guide/en/#outputmanualchunks) option, where you can manually define custom common chunks

```json
{
  "manualChunks": {
    "common": ["react", "react-dom"]
  }
}
```

evanw:
Why would that be a better thing to do than automatically-computed chunks? I'm sure there are reasons, I just want to make sure I know why you would want to do this so that I know what this feature request is about. Is it because aesthetically seeing multiple files called `chunk` isn't desirable? Is it wanting to improve caching for repeated page loads? Is it about wanting to control the number of chunks? Something else?

intrnl:
it's primarily about caching. when your application code and its dependencies is bundled into one, users will have to keep downloading everything even though only your app code changes. so i think that it would be nice if the dependencies are split into its own.

an alternative proposal would be to have a toggle to move everything under node_modules to be its own, but i think providing manual control like this would be better especially if other devs have their own strategy for splitting.

evanw:
If this is about caching, then should this disable tree shaking? Otherwise when tree shaking is active, the size of even manually-specified chunks can grow and shrink between builds.

Also it's worth noting that esbuild will need other changes to achieve these caching goals. Right now all symbols in the bundle are minified together. Optimal caching will require symbols to be minified independently per chunk.

intrnl:
Tree shaking might be problematic yeah, which is quite the tradeoff

ref: rollup/rollup#2788

Would've liked to say that devs should only be specifying manual chunks for things that can't be tree shaken, but I think a lot will miss the point 😅

visj:
Closure Compiler allows [code splitting manual chunks](https://stackoverflow.com/questions/10395810/how-do-i-split-my-javascript-into-modules-using-googles-closure-compiler). If esbuild were to allow manual chunks in the same way, it could also be used as a very strong developer tool for applications using Closure Compiler as esbuild is a magnitude faster.

If you for instance have 5 entry points [a,b,c,d,e], and some code is shared only between [a,b], you could split that code out as well as code shared by all 5 chunks and make only routes [a,b] depend on the first chunk. I think that's how Google splits out their JS when you get different widgets in their search (such as the calculator or stock graphs).

mjackson:
> If this is about caching, then should this disable tree shaking?

[...]

PR #1078, opened 2021-03-28
PR for version 0.11.0

### LINKED ISSUE #667: Circular dependency issue
I'm porting the build system for Yarn and it seems we hit a pathological case where ESBuild incorrectly resolves cyclic dependencies. To reproduce:

```
git clone git@github.com:yarnpkg/berry
git checkout 9f7809c2b5f620203626b8f9c9f47e451863f5f5
yarn build:cli --no-minify
YARN_IGNORE_PATH=1 node ./packages/yarnpkg-cli/bundles/yarn.js --version
```

The problem seems to be around the two following lines:

https://github.com/yarnpkg/berry/blob/17beb526719627a677585f5ce0b9ced146a58def/packages/plugin-pnp/sources/PnpLinker.ts#L8
https://github.com/yarnpkg/berry/blob/17beb526719627a677585f5ce0b9ced146a58def/packages/plugin-pnp/sources/index.ts#L6

It causes `PnpLinker` to be undefined in `index.ts`, thus crashing the rest of the program since it then gets stored [here](https://github.com/yarnpkg/berry/blob/master/packages/plugin-pnp/sources/index.ts#L115) as undefined as well. Since the codebase only ever access `PnpLinker.ts` via `index.ts`, I don't see why it would be `undefined`.

### LINKED ISSUE #706: ESM entry module's exports signature is not preserved when `required()` by other modules
Context: I am using esbuild to do dependency-pre-bundling in Vite. This is a multiple-entry build with `splitting: true`. The entries are resolved package entry files - e.g.

```
- node_modules/preact/index.module.js
- node_modules/react/index.js
```

I'm also using a plugin with `onResolve` that redirects any imports to `preact` or `react` to these already resolved file locations to avoid duplicated copies of them in the resulting bundle.

Preact is a pure ESM module, so its corresponding output entry chunk is expected to preserve the same export signature as the original entry module. However, this is broken if any transitive module in the bundle contains a cjs `require()` call to it. E.g. if somewhere down the tree a dependency does this:

```js
const preact = require('preact')
```

This causes esbuild to wrap the ESM preact module with a CommonJS wrapper and expose only a `default` export. This makes sense for the internal interop, but currently this also affects the final output entry chunk for preact: now the output chunk for preact also only has a `default` export! This creates a mismatch between the output of the preact chunk and the original entry chunk.

What I imagine esbuild needs to do is to

- move both the preact module and its wrapped CommonJS code to a shared chunk
- re-export the same original exports from the preact entry chunk
- import and use the wrapped CommonJS version in the sub dependency that uses `require()`

---

Currently I have to work around this by detecting exports mismatch between input/output entry chunks and perform additional runtime interop. Would be great to have this fixed in esbuild itself.

### LINKED ISSUE #791: onResolve is not fired on entrypoints when namespace equals 'file'

### LINKED ISSUE #945: onResolve shouldn't change output path for entry points
By default, `esbuild` replicates file structure into the output directory. So, the following setup:

```js
require('esbuild').build({
    entryPoints: [
        'out_test/bar/index.js',
        'out_test/foo/index.js'
    ],
    bundle: true,
    outdir: 'out_test_build'
});
```

Will produce a result on `outdir`:

```
out_test_build
   bar
      index.js
   foo
      index.js
```

That's what we expected. However, if we're using `onResolve` for an entry point and `onLoad` to generate a content this way:

```js
require('esbuild').build({
    entryPoints: [
        'out_test/bar/index.js',
        'out_test/foo/index.js'
    ],
    bundle: true,
    outdir: 'out_test_build',
    plugins: [{
        name: 'example',
        setup({ onResolve, onLoad }) {
            onResolve({ filter: /.*/ }, args => {
                if (args.kind === 'entry-point') {
                    return {
                        namespace: 'test',
                        path: 'ref',
                        pluginData: args
                    };
                }
            });
            onLoad({ namespace: 'test', filter: /.*/}, args => {
                return {
                    contents: 'console.log("generated content")'
                };
            });
        }
    }]
});
```

We get an error:

```
 > error: Duplicate entry point "test:ref"

~/tmp/esbuild/node_modules/esbuild/lib/main.js:1195
  let error = new Error(`${text}${summary}`);
              ^

Error: Build failed with 1 error:
error: Duplicate entry point "test:ref"
    at failureErrorWithLog (~/tmp/esbuild/node_modules/esbuild/lib/main.js:1195:15)
    at buildResponseToResult (~/tmp/esbuild/node_modules/esbuild/lib/main.js:920:32)
    at ~/tmp/esbuild/node_modules/esbuild/lib/main.js:1019:20
    at ~/tmp/esbuild/node_modules/esbuild/lib/main.js:566:9
    at handleIncomingPacket (~/tmp/esbuild/node_modules/esbuild/lib/main.js:655:9)
    at Socket.readFromStdout (~/tmp/esbuild/node_modules/esbuild/lib/main.js:533:7)
    at Socket.emit (node:events:378:20)
    at addChunk (node:internal/streams/readable:313:12)
    at readableAddChunk (node:internal/streams/readable:288:9)
    at Socket.Readable.push (node:internal/streams/readable:227:10) {
  errors: [
    {
      detail: undefined,
      location: null,
      notes: [],
      text: 'Duplicate entry point "test:ref"'
    }
  ],
  warnings: []
}
```

It is surprising that we can't resolve entry points to the same reference, since later its content may be loaded/generated (using `onLoad`) depending on other resolve parameters. But ok, let's keep original path and pass extra parameters through `pluginData`:

```js
require('esbuild').build({
    entryPoints: [
        'out_test/bar/index.js',
        'out_test/foo/index.js'
    ],
    bundle: true,
    outdir: 'out_test_build',
    plugins: [{
        name: 'example',
        setup({ onResolve, onLoad }) {
            onResolve({ filter: /.*/ }, args => {
                if (args.kind === 'entry-point') {
                    return {
                        namespace: 'test',
                        path: args.path,
                        pluginData: {
                            ref: 'something'
                        }
                    };
                }
            });
            onLoad({ namespace: 'test', filter: /.*/}, args => {
                return {
                    contents: 'console.log("generated content")'
                };
            });
        }
    }]
});
```

In this case, we get another error:

```
 > error: Two output files share the same path but have different contents: out_test_build/index.js

~/tmp/esbuild/node_modules/esbuild/lib/main.js:1195
  let error = new Error(`${text}${summary}`);
              ^

Error: Build failed with 1 error:
error: Two output files share the same path but have different contents: out_test_build/index.js
    at failureErrorWithLog (~/tmp/esbuild/node_modules/esbuild/lib/main.js:1195:15)
    at buildResponseToResult (~/tmp/esbuild/node_modules/esbuild/lib/main.js:920:32)
    at ~/tmp/esbuild/node_modules/esbuild/lib/main.js:1019:20
    at ~/tmp/esbuild/node_modules/esbuild/lib/main.js:566:9
    at handleIncomingPacket (~/tmp/esbuild/node_modules/esbuild/lib/main.js:655:9)
    at Socket.readFromStdout (~/tmp/esbuild/node_modules/esbuild/lib/main.js:533:7)
    at Socket.emit (node:events:378:20)
    at addChunk (node:internal/streams/readable:313:12)
    at readableAddChunk (node:internal/streams/readable:288:9)
    at Socket.Readable.push (node:internal/streams/readable:227:10) {
  errors: [
    {
      detail: undefined,
      location: null,
      notes: [],
      text: 'Two output files share the same path but have different contents: out_test_build/index.js'
    }
  ],
  warnings: []
}
```

Looks like `esbuild` is losing dir path for entry points. I tried `outbase` but it has no effect.

Is this the expected behaviour?

I believe an entry point path should be untouched disregarding `onResolve`/`onLoad` is used or not. Otherwise it's lead to unexpected errors. Also make hard to recognise entry points in a result, since paths are different from those which we pass to `entryPoints`. As a workaround for now, we map entry point paths to an unique name (like `1.js`, `2.css`, `3.js` etc) and re-map them back in result's `outputFiles`.

### LINKED ISSUE #1056: --entry-names=[name] errors out for some dynamic imports
Hey Evan! Thanks for your hard work on ESBuild.

### Issue

Framer stumbled upon a case when `--entry-names=[name]` generates a file conflict with dynamic imports. Specifically, a code like:

```js
// entry.js
import("date-fns");
import("axios");
```

with a command like:

```sh
yarn esbuild entry.js --outdir=build --bundle --splitting --format=esm --entry-names=[name]
```

generates the following error:

```
> error: Two output files share the same path but have different contents: build/index.js
```

Here’s the repo with the reproduction: https://github.com/iamakulov/esbuild-entrynames-repro

### Cause

The error happens because both `date-fns`’ and `axios`’ main file is called `index.js`. ESBuild takes that file name as the entry point name. This causes it to generate two different chunks with an identical name of `index`.

The challenge here is that this is a third-party code, and we can’t control how the main file of a specific library is called.

### Potential solutions

Here are a couple potential solutions I can think of:

1. **Explicitly rename entrypoints with identical names.** As in, if there’re two `index` entrypoints, rename them to something else so that names don’t clash.
 
   One deterministic way to do this is to include `[dir]` into `[name]` when (and only when) there are multiple identical `[name]`s. In this case, with the example above, instead of generating two `build/index.js` chunks and erroring out, ESBuild would generate:
   
   ```
   build/node_modules-date-fns-index.js 
   build/node_modules-axios-index.js
   ```

2. **Manage names of `import()` chunks with a different flag.** As in, don’t use `--entry-names` for `import()` – use `--chunk-names` or eg a new `--dynamic-import-names` flag instead. 

   It’s not obvious to me why dynamic imports are managed by `--entry-names` (and are considered to be entry points).
   
   To me, an _entry point_ is a script that’s inserted straight into the HTML page and starts the app execution. Names of entry points matter because an external system may be inserting an entry point into an HTML response.
    
    Unlike a real entry point, a _dynamic import_ is loaded from ESBuild-generated JS code, not from HTML. Its name doesn’t really matter – ESBuild takes full care of loading that import, and no external system normally deals with that chunk. Pretty much like a regular code-split chunk.

evanw:
This is a PR for version 0.11.0, which is a release that contains the following breaking changes:

* Change how `require()` and `import()` of ESM works (fixes #667, fixes #706)

    Previously if you call `require()` on an ESM file, or call `import()` on an ESM file with code splitting disabled, esbuild would convert the ESM file to CommonJS. For example, if you had the following input files:

    ```js
    // cjs-file.js
    console.log(require('./esm-file.js').foo)
    // esm-file.js
    export let foo = bar()
    ```

    The previous bundling behavior would generate something like this:

    ```js
    var require_esm_file = __commonJS((exports) => {
      __markAsModule(exports);
      __export(exports, {
        foo: () => foo
      });
      var foo = bar();
    });
    console.log(require_esm_file().foo);
    ```

    This behavior has been changed and esbuild now generates something like this instead:

    ```js
    var esm_file_exports = {};
    __export(esm_file_exports, {
      foo: () => foo
    });
    var foo;
    var init_esm_file = __esm(() => {
      foo = bar();
    });
    console.log((init_esm_file(), esm_file_exports).foo);
    ```

    The variables have been pulled out of the lazily-initialized closure and are accessible to the rest of the module's scope. Some benefits of this approach:

    * If another file does `import {foo} from "./esm-file.js"`, it will just reference `foo` directly and will not pay the performance penalty or code size overhead of the dynamic property accesses that come with CommonJS-style exports. So this improves performance and reduces code size in some cases.

    * This fixes a long-standing bug (#706) where entry point exports could be broken if the entry point is a target of a `require()` call and the output format was ESM. This happened because previously calling `require()` on an entry point converted it to CommonJS, which then meant it only had a single `default` export, and the exported variables were inside the CommonJS closure and inaccessible to an ESM-style `export {}` clause. Now calling `require()` on an entry point only causes it to be lazily-initialized but all exports are still in the module scope and can still be exported using a normal `export {}` clause.

    * Now that this has been changed, `import()` of a module with top-level await (#253) is now allowed when code splitting is disabled. Previously this didn't work because `import()` with code splitting disabled was implemented by converting the module to CommonJS and using `Promise.resolve().then(() => require())`, but converting a module with top-level await to CommonJS is impossible because the CommonJS call signature must be synchronous. Now that this implemented using lazy initialization instead of CommonJS conversion, the closure wrapping the ESM file can now be `async` and the `import()` expression can be replaced by a call to the lazy initializer.

    * Adding the ability for ESM files to be lazily-initialized is an important step toward additional future code splitting improvements including: manual chunk names (#207), correct import evaluation order (#399), and correct top-level await evaluation order (#253). These features all need to make use of deferred evaluation of ESM code.

    In addition, calling `require()` on an ESM file now recursively wraps all transitive dependencies of that file instead of just wrapping that ESM file itself. This is an increase in the size of the generated code, but it is important for correctness (#667). Calling `require()` on a module means its evaluation order is determined at run-time, which means the evaluation order of all dependencies must also be determined at run-time. If you don't want the increase in code size, you should use an `import` statement instead of a `require()` call.

* Dynamic imports now use chunk names instead of entry names (fixes #1056)

    Previously the output paths of dynamic imports (files imported using the `import()` syntax) were determined by the `--entry-names=` setting. However, this can cause problems if you configure the `--entry-names=` setting to omit both `[dir]` and `[hash]` because then two dynamic imports with the same name will cause an output file name collision.

    Now dynamic imports use the `--chunk-names=` setting instead, which is used for automatically-generated chunks. This setting is effectively required to include `[hash]` so dynamic import name collisions should now be avoided.

    In addition, dynamic imports no longer affect the automatically-computed default value of `outbase`. By default `outbase` is computed to be the [lowest common ancestor](https://en.wikipedia.org/wiki/Lowest_common_ancestor) directory of all entry points. Previously dynamic imports were considered entry points in this calculation so adding a dynamic entry point could unexpectedly affect entry point output file paths. This issue has now been fixed.

* Allow custom output paths for individual entry points

    By default, esbuild will automatically generate an output path for each entry point by computing the relative path from the `outbase` directory to the entry point path, and then joining that relative path to the `outdir` directory. The output path can be customized using `outpath`, but that only works for a single file. Sometimes you may need custom output paths while using multiple entry points. You can now do this by passing the entry points as a map instead of an array:

    * CLI
        ```
        esbuild out1=in1.js out2=in2.js --outdir=out
        ```

    * JS
        ```js
        esbuild.build({
          entryPoints: {
            out1: 'in1.js',
            out2: 'in2.js',
          },
          outdir: 'out',
        })
        ```

    * Go

        ```go
        api.Build(api.BuildOptions{
          EntryPointsAdvanced: []api.EntryPoint{{
            OutputPath: "out1",
            InputPath: "in1.js",
          }, {
            OutputPath: "out2",
            InputPath: "in2.js",
          }},
          Outdir: "out",
        })
        ```

    This will cause esbuild to generate the files `out/out1.js` and `out/out2.js` inside the output directory. These custom output paths are used as input for the `--entry-names=` path template setting, so you can use something like `--entry-names=[dir]/[name]-[hash]` to add an automatically-computed hash to each entry point while still using the custom output path.

* Derive entry point output paths from the original input (fixes #945)

    Previously esbuild would determine the output path for an entry point by looking at the post-resolved path. For example, running `esbuild --bundle react --outdir=out` would generate the output path `out/index.js` because the input path `react` was resolved to `node_modules/react/index.js`. With this release, the output path is now determined by looking at the pre-resolved path. For example, For example, running `esbuild --bundle react --outdir=out` now generates the output path `out/react.js`. If you need to keep using the output path that esbuild previously generated with the old behavior, you can use the custom output path feature (described above).

* Use the `file` namespace for file entry points (fixes #791)

    Plugins that contain an `onResolve` callback with the `file` filter don't apply to entry point paths because it's not clear that entry point paths are files. For example, you could potentially bundle an entry point of `https://www.example.com/file.js` with a HTTP plugin that automatically downloads data from the server at that URL. But this behavior can be unexpected for people writing plugins.

    With this release, esbuild will do a quick check first to see if the entry point path exists on the file system before running plugins. If it exists as a file, the namespace will now be `file` for that entry point path. This only checks the exact entry point name and doesn't attempt to search for the file, so for example it won't handle cases where you pass a package path as an entry point or where you pass an entry point without an extension. Hopefully this should help improve this situation in the common case where the entry point is an exact path.


zaydek:
Thank you! The ability to alias entry names should be really useful.

PR #1085, opened 2021-03-29
no test for `TDotDotDot`

SyMind:


evanw:
Many punctuation tokens aren't covered here, so the addition of this case doesn't seem like it adds anything. I'm going to close this issue.

PR #1097, opened 2021-04-01
Add cross-origin isolation headers

haywirez:
This PR adds COOP and COEP headers when serving locally in development using `--servedir`, as they are needed if working with the re-enabled `SharedArrayBuffer`s in your code:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/Planned_changes
https://developer.chrome.com/blog/enabling-shared-array-buffer/
https://web.dev/coop-coep/

### DIFF
--- a/pkg/api/serve_other.go
+++ b/pkg/api/serve_other.go
@@ -117,6 +117,8 @@ func (h *apiHandler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 	// Handle get requests
 	if req.Method == "GET" && strings.HasPrefix(req.URL.Path, "/") {
 		res.Header().Set("Access-Control-Allow-Origin", "*")
+		res.Header().Set("Cross-Origin-Embedder-Policy", "require-corp")
+		res.Header().Set("Cross-Origin-Opener-Policy", "same-origin")
 		queryPath := path.Clean(req.URL.Path)[1:]
 		result := h.build()
 


Jarred-Sumner:
Do API requests to different endpoints still work in localhost with COOP and COEP and CORS enabled? Such as an API server on localhost:3001 when the esbuild server runs on port 3000

haywirez:
Not sure if I understood your question completely, but it seems to work fine using a different proxied origin (e.g. esbuild running on `:8000`, node process serving proxied request on `:3000`). I used the steps from the [custom server behavior example](https://esbuild.github.io/api/#customizing-server-behavior) in the docs.

evanw:
I don't think this is a good default to have. This is (by design) adding a restriction to all content served by esbuild. If you do this, all external scripts break (the error in Chrome is `net::ERR_BLOCKED_BY_RESPONSE`). Breaking all external scripts to enable `SharedArrayBuffer` doesn't seem like the right trade-off since I expect many more people to use external scripts than to use `SharedArrayBuffer`.

You should be able to do this with a proxy if you need to do this. With the example proxy server in the docs, you can do that with this code:

```js
proxyRes.headers['Cross-Origin-Embedder-Policy'] = 'require-corp'
proxyRes.headers['Cross-Origin-Opener-Policy'] = 'same-origin'
```

haywirez:
Sounds reasonable. If anyone needs it, this is a working setup for a COOP/COEP enabled proxy server:

```ts
// server.ts
const esbuild = require("esbuild");
const http = require("http");

// Start esbuild's server on a random local port
esbuild
  .serve(
    {
      servedir: "www",
    },
    {
      entryPoints: ["src/app.tsx"],
      outdir: "www/js",
      bundle: true,
      target: "es2020",
      define: { "process.env.NODE_ENV": "'production'" },
    }
  )
  .then((result) => {
    // The result tells us where esbuild's local server is
    const { host, port } = result;

    console.log(`started esbuild proxy server on host ${host}:${port}`);

    const proxyPort = 3000;

    // Then start a proxy server on defined port
    http
      .createServer((req, res) => {
        // Add cross-origin isolation headers
        res.setHeader("Cross-Origin-Embedder-Policy", "require-corp");
        res.setHeader("Cross-Origin-Opener-Policy", "same-origin");

        const options = {
          hostname: host,
          port: port,
          path: req.url,
          method: req.method,
          headers: req.headers,
        };

        // Forward each incoming request to esbuild
        const proxyReq = http.request(options, (proxyRes) => {
          res.writeHead(proxyRes.statusCode, proxyRes.headers);
          proxyRes.pipe(res, { end: true });
        });

        // Forward the body of the request to esbuild
        req.pipe(proxyReq, { end: true });
      })
      .listen(proxyPort);

    console.log(`listening on ${proxyPort}`);
  });

```
```json
// package.json
"scripts": {
    "proxy": "node src/server.ts",
    ...
}
```

Issue #1142, opened 2021-04-12
Would love a way to silence the "indirect calls to require" warnings for 3rd party dependencies


lgandecki:
I understand that I can silent the warnings with try catch around a require, but that's not really a solution when the require is in node_modules and it's a dependency of one of my dependencies. I guess I can try to require my dependency instead of import it, but then I'm losing TypeScript types and all in all it is a bit silly :) 
I would love to be able to "acknowledge" warnings, if I see a new one, I'd investigate and check whether it might be an issue or not, if it's not an issue, I want to silence it (an array of packages to silence is one option that comes to mind).

Did anyone find a nice way to deal with that? 

BTW: great job on this package, the speed improvement and the different workflow that this enables is mind-blowing. Thanks!

evanw:
You can disable the printing of warnings by setting the [`logLevel`](https://esbuild.github.io/api/#log-level) setting to `error` in esbuild's API. If you would still like to see certain warnings, you can print them yourself by iterating over the [`warnings`](https://github.com/evanw/esbuild/blob/d616dac2ca7cb0f9bd918b96fa44ece3578ee9b3/lib/types.ts#L120) array returned from esbuild's API. If you would like to pretty-print the warnings using esbuild's log style, you can call the [`formatMessages`](https://github.com/evanw/esbuild/blob/d616dac2ca7cb0f9bd918b96fa44ece3578ee9b3/lib/types.ts#L334-L340) API.

evanw:
Closing this issue as I have now completely removed the warning in #1155.

lgandecki:
ok, interesting approach. but is there a chance to SEE those warning? One of the warnings did help me to fix my build, it would be super hard to figure out without that.

evanw:
With the latest release, you can now opt-in to seeing some of these messages that you could have seen previously (along with some additional ones) with `--log-level=debug`.

Issue #1132, opened 2021-04-09
This call to "require" will not be bundled because the argument is not a string literal (surround with a try/catch to silence this warning)


Valexr:
Can't bundling mongoose module

![Screenshot 2021-04-09 at 19 00 47](https://user-images.githubusercontent.com/5656366/114209836-c1616980-9967-11eb-9139-c735d08b0737.jpg)

ggoodman:
Looks like a dupe of #1127.

evanw:
Yes, it's not possible to bundle things like this. The warnings are correct. They are warnings instead of errors because your bundle could hypothetically still work if these code paths are not relied upon. But if you need this to work then you're out of luck. You'll either have to run the code unbundled or use another tool instead of esbuild.

evanw:
Closing this because this is by design.

Valexr:
Ok. Tx for res @evanw 🤓

ndr47:
Does somebody know any workaround to bundle mongoose module with this?

Issue #1045, opened 2021-03-24
require.main not working


metachris:
This typical pattern doesn't work in a bundled output script:

```typescript
if (require.main === module) {
  // ...
}
```

See also #560 

Output during build:

```bash
 > src/main.ts:8:4: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    8 │ if (require.main === module) {
      ╵     ~~~~~~~
```

metachris:
Sorry my mistake, needed to add `--platform=node`

PR #1155, opened 2021-04-15
remove "require" and "import" warnings

### LINKED ISSUE #1153: Checking for existence of require in if throws warning
Sentry checks for `require` inside an if to throw an if no `appName` is provided and it can't require Electron. Unfortunately, esbuild thinks that this is a dynamic `require` and throws a warning.

Screen Shot 2021-04-14 at 14 40 35

Is it possible to not warn in case some code is just checking for the existence of require?

evanw:
Previously esbuild had warnings when bundling about uses of `require` and `import` that are not of the form `require(<string literal>)` or `import(<string literal>)`. These warnings existed because the bundling process must be able to statically-analyze all dynamic imports to determine which files must be included. Here are some real-world examples of cases that esbuild doesn't statically analyze:

* From [`mongoose`](https://www.npmjs.com/package/mongoose):

    ```js
    require('./driver').set(require(global.MONGOOSE_DRIVER_PATH));
    ```

* From [`moment`](https://www.npmjs.com/package/moment):

    ```js
    aliasedRequire = require;
    aliasedRequire('./locale/' + name);
    ```

* From [`logform`](https://www.npmjs.com/package/logform):

    ```js
    function exposeFormat(name) {
      Object.defineProperty(format, name, {
        get() { return require(`./${name}.js`); }
      });
    }
    exposeFormat('align');
    ```

All of these dynamic imports will not be bundled (i.e. they will be left as-is) and will crash at run-time if they are evaluated. Some of these crashes are ok since the code paths may have error handling or the code paths may never be used. Other crashes are not ok because the crash will actually be hit.

The warning from esbuild existed to let you know that esbuild is aware that it's generating a potentially broken bundle. If you discover that your bundle is broken, it's nice to have a warning from esbuild to point out where the problem is. And it was just a warning so the build process still finishes and successfully generates output files. If you didn't want to see the warning, it was easy to turn it off via `--log-level=error`.

However, there have been quite a few complaints about this warning. Some people seem to not understand the difference between a warning and an error, and think the build has failed even though output files were generated. Other people do not want to see the warning but also do not want to enable `--log-level=error`.

This PR removes this warning for both `require` and `import`. Now when you try to bundle code with esbuild that contains dynamic imports not of the form `require(<string literal>)` or `import(<string literal>)`, esbuild will just silently generate a potentially broken bundle. This may affect people coming from other bundlers that support certain forms of dynamic imports that are not compatible with esbuild such as the [Webpack-specific dynamic `import()` with pattern matching](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import).

Related past issues:

* Fixes #1153
* #1142
* #1132
* #1045
* #812
* #661
* #574
* #512
* #495
* #480
* #453
* #410
* #80


mattfysh:
Nice :) Wondering if instead of a build time warning, could the offending code be modified to log out a warning if the code path is entered at runtime?

evanw:
A note for anyone who ends up here:

You may be able to get more information to debug a crash due to unbundled `require` or `import` by using `--log-level=debug`. Logging of debug information for some of these cases was added in a later release. The difference is that the log level defaults to `info` so debug information is not logged by default.

evanw:
With the [latest release](https://github.com/evanw/esbuild/releases/v0.14.42), you can now use `--log-override:unsupported-require-call=warning` to make this a warning again (or `--log-override:unsupported-require-call=error` to make this an error).

PR #1160, opened 2021-04-16
Fix invalid mangling of zero and negative zero

### LINKED ISSUE #1159: [bug] equality check upon -0 and 0
## Reproduction

code like this:
```js
var isTrue = document.body.innerHTML === '';

console.log(isTrue ? -0 : 0)
```
after compiled with `--minify-syntax` option enabled, code becomes:
```js
var isTrue = document.body.innerHTML === "";
console.log(-0);
```

The ternary operator is removed, which may cause bugs in math operations. In our case, we use `Decimal.js`，and the `isNegative` function produces completely wrong answer when the value is negative zero.

## intended behavior

These two numbers are basically the same, like if your are not familiar with JavsScript, you would even know that -0 actually exists in this language. But they are different:
```js
Object.is(-0 , 0); // false

1 / 0; // Infinity

1 / -0; // Infinity
```

esbuild should be able to tell the difference between these different types of zero.

## What I found so far

`checkEqualityIfNoSideEffects` is the function for comparing two numbers. The number section is:
```go
r, ok := right.(*js_ast.ENumber)
return ok && l.Value == r.Value, ok
```
When the two parameters are numbers,`==` is used. Are they transformed to numbers in go first?

rtsao:
Fixes #1159

It seemed better to just add a special case within `valuesLookTheSame` rather than introduce an equality mode configuration into `checkEqualityIfNoSideEffects`.

Or it may be better to put this logic into `mangleIfExpr` directly, I wasn't sure.

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -889,6 +889,15 @@ func valuesLookTheSame(left js_ast.E, right js_ast.E) bool {
 			}
 			return true
 		}
+
+	// Special-case to distinguish between negative an non-negative zero when mangling
+	// "a ? -0 : 0" => "a ? -0 : 0"
+	// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness
+	case *js_ast.ENumber:
+		b, ok := right.(*js_ast.ENumber)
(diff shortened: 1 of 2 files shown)


evanw (approved):
Yes I agree, this looks like a good fix. Thanks for doing this and for adding tests!

PR #1179, opened 2021-04-23
Minify property accesses on object literals

### LINKED ISSUE #1166: Simplify constant object access
Note: I opened this issue with a higher level question, which was how `{foo: 'bar'}.foo === 'bar'` should be minified to `!0`, but this is due to the property accessor not simplifying. If this issue is fixed, then that statement should be minified for free (there is already constant string equality checking). The original issue is at the bottom.

`{ foo: 'bar' }.foo` should be minified to `"bar"`.
This should work with a fair amount of depth:
`{ foo: { bar: 'baz' } }.foo.bar` should be minified to `"baz"`.

Would modifying js_parser maybeRewritePropertyAccess be the place for this to go?

<details>
  <summary>Original issue is here. Click to expand.</summary>
  
`console.log({foo: 'bar'}.foo === 'bar')` is not minified to `console.log(!0)`
esbuild does minify some other simple statements like this:
`console.log('foo' === 'foo')` _is_ minified to `console.log(!0)`

Terser minifies both statements.

---

My story: I was using Snowpack which uses `import.meta.env` to access the environment, e.g. `import.meta.env.NODE_ENV`. (instead of process.env.) I added support for this in rollup using the resolveMetaImport plugin hook.

My input:

```js
if (import.meta.env.NODE_ENV === 'development') {
    console.log('dev');
} else {
    console.log('prod');
}
```
Output from rollup which inlines the `import.meta.env` access:
```js
if ({"NODE_ENV":"production"}.NODE_ENV === "development") {
  console.log("dev");
} else {
  console.log("prod");
}
```
with esbuild's minification:
```js
{NODE_ENV:"production"}.NODE_ENV==="development"?console.log("dev"):console.log("prod");
```

I was hoping that esbuild would recognise this pattern and remove the dead code. Terser's output in this case:
```js
console.log("prod");
```

---

</details>

I understand that esbuild's minifier doesn't aim to support everything, so if that's the case here, all good!

evanw:
The code `{a: {b: 1}}.a.b` will now be minified to `1`. This optimization is relatively complex and hard to do safely. Here are some tricky cases that are correctly handled:

```js
var obj = {a: 1}
assert({a: 1, a: 2}.a === 2)
assert({a: 1, [String.fromCharCode(97)]: 2}.a === 2)
assert({__proto__: obj}.a === 1)
assert({__proto__: null}.a === undefined)
assert({__proto__: null}.__proto__ === undefined)
assert({a: function() { return this.b }, b: 1}.a() === 1)
assert(({a: 1}.a = 2) === 2)
assert(++{a: 1}.a === 2)
assert.throws(() => { new ({ a() {} }.a) })
```

Fixes #1166

### DIFF
--- a/internal/js_parser/js_parser_test.go
+++ b/internal/js_parser/js_parser_test.go
@@ -2930,6 +2930,18 @@ func TestMangleObject(t *testing.T) {
 	expectPrintedMangle(t, "x = {a, ...'123', b}", "x = {a, ...\"123\", b};\n")
 	expectPrintedMangle(t, "x = {a, ...[1, 2, 3], b}", "x = {a, ...[1, 2, 3], b};\n")
 	expectPrintedMangle(t, "x = {a, ...(()=>{})(), b}", "x = {a, ...(() => {\n})(), b};\n")
+
+	// Check simple cases of object simplification (advanced cases are checked in end-to-end tests)
+	expectPrintedMangle(t, "x = {['y']: z}.y", "x = {y: z}.y;\n")
+	expectPrintedMangle(t, "x = {['y']: z}.y; var z", "x = z;\nvar z;\n")
+	expectPrintedMangle(t, "x = {foo: foo(), y: 1}.y", "x = {foo: foo(), y: 1}.y;\n")
+	expectPrintedMangle(t, "x = {foo: /* @__PURE__ */ foo(), y: 1}.y", "x = 1;\n")
+	expectPrintedMangle(t, "x = {__proto__: null}.y", "x = void 0;\n")
+	expectPrintedMangle(t, "x = {__proto__: null, y: 1}.y", "x = 1;\n")
+	expectPrintedMangle(t, "x = {__proto__: null}.__proto__", "x = void 0;\n")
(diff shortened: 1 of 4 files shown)


internal/js_parser/js_parser_test.go
```diff
@@ -2930,6 +2930,18 @@ func TestMangleObject(t *testing.T) {
 	expectPrintedMangle(t, "x = {a, ...'123', b}", "x = {a, ...\"123\", b};\n")
 	expectPrintedMangle(t, "x = {a, ...[1, 2, 3], b}", "x = {a, ...[1, 2, 3], b};\n")
 	expectPrintedMangle(t, "x = {a, ...(()=>{})(), b}", "x = {a, ...(() => {\n})(), b};\n")
+
+	// Check simple cases of object simplification (advanced cases are checked in end-to-end tests)
+	expectPrintedMangle(t, "x = {['y']: z}.y", "x = {y: z}.y;\n")
```
lukeed, line 2935: These, by themselves, are amazing! 🙌

PR #1190, opened 2021-04-25
Speed up ProcessDefines by like 0.03ms 

Jarred-Sumner:
I thought this would have more impact but it does not

```go
func BenchmarkProcessDefines(b *testing.B) {
	smallMap := make(map[string]config.DefineData, 1)
	smallMap["process.env.NODE_ENV"] = config.DefineData{}

	for i := 0; i < b.N; i++ {
		config.ProcessDefines(smallMap)
	}

}
// Before:
// BenchmarkProcessDefines-16    	   15420	     75798 ns/op	  100711 B/op	     134 allocs/op
// After:
// BenchmarkProcessDefines-16    	   26157	     39537 ns/op	   55366 B/op	     107 allocs/op
```


evanw:
Closing. Thanks for the PR, but I don't think this hard-coding these numbers is worth it.

PR #1192, opened 2021-04-25
Provide options for how to handle legal comments

### LINKED ISSUE #919: Add option to not automatically include licenses in footer
I already embed different, more-concise licenses in a banner and want to opt-out of license inclusion in the footers.

Judging from [this code in linker.go](https://github.com/evanw/esbuild/blob/b72805747e10f4072de9defc851543727d2d8ef5/internal/bundler/linker.go#L3972-L3975), it appears that there is no way to opt-out of automatic license inclusion as footers are added are from an append-only list.

This could be implemented by special-casing `footer: false` in the build API to remove *every* footer.

evanw:
A "legal comment" is considered to be any comment that contains `@license` or `@preserve` or that starts with `//!` or `/*!`. These comments are preserved in output files by esbuild since that follows the intent of the original authors of the code.

However, some people want to remove the automatically-generated license information before they distribute their code. To facilitate this, esbuild now provides several options for how to handle legal comments (via `--legal-comments=` in the CLI and `legalComments` in the JS API):

* `none`: Do not preserve any legal comments
* `inline`: Preserve all statement-level legal comments
* `eof`: Move all statement-level legal comments to the end of the file
* `linked`: Move all statement-level legal comments to a `.LICENSE.txt` file and link to them with a comment
* `external`: Move all statement-level legal comments to a `.LICENSE.txt` file but to not link to them

The default behavior is `eof` when bundling and `inline` otherwise.

Fixes #919
Closes #1033

### DIFF
--- a/internal/bundler/linker.go
+++ b/internal/bundler/linker.go
@@ -94,6 +94,9 @@ type chunkInfo struct {
 	// is the substitution of the final hash into "finalTemplate".
 	finalRelPath string
 
+	// If non-empty, this chunk needs to generate an external legal comments file.
+	externalLegalComments []byte
+
 	// When this chunk is initially generated in isolation, the output pieces
 	// will contain slices of the output with the unique keys of other chunks
 	// omitted.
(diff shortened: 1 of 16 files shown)


internal/bundler/linker.go
```diff
@@ -4320,10 +4347,25 @@ func (c *linkerContext) generateChunkJS(chunks []chunkInfo, chunkIndex int, chun
...
-		j.AddString(text)
-		j.AddString("\n")
+	if len(legalCommentList) > 0 {
+		sort.Strings(legalCommentList)
+
+		switch c.options.LegalComments {
+		case config.LegalCommentsEndOfFile:
+			for _, text := range legalCommentList {
+				j.AddString(text)
+				j.AddString("\n")
```
eligrey, line 4357: I think this line should come before `j.AddString(text)` so that users don't end up with an extra newline at the end of the file.

internal/bundler/linker.go
```diff
@@ -4320,10 +4347,25 @@ func (c *linkerContext) generateChunkJS(chunks []chunkInfo, chunkIndex int, chun
...
+			}
+
+		case config.LegalCommentsLinkedWithComment,
+			config.LegalCommentsExternalWithoutComment:
+			jComments := helpers.Joiner{}
+			for _, text := range legalCommentList {
+				jComments.AddString(text)
+				jComments.AddString("\n")
```
eligrey, line 4365: Ditto for my comment above

internal/bundler/linker.go
evanw: Files generated by esbuild deliberately contain a trailing newline so that printing them to the terminal doesn't mess with the terminal prompt.

Issue #1044, opened 2021-03-24
CSS asset URLs are incorrect when --entry-names is used


kaksmet:
Running
```
esbuild index.css --bundle --outdir=build --entry-names=styles/[name] --asset-names=assets/[name] --loader:.png=file
```
with `index.css` looking like
```css
.foobar {
    background-image: url(./foobar.png);
}
```
results in `build/styles/index.css`:
```css
/* index.css */
.foobar {
  background-image: url(./assets/foobar.png);
}
```
Since `foobar.png` is copied to `build/assets/foobar.png` the URL should actually be `../assets/foobar.png`?

evanw:
Thanks for catching this. I'll have to investigate this more to figure out what's going on, but in the meantime you may be able to use the [public path](https://esbuild.github.io/api/#public-path) feature to work around this for now.

evanw:
It looks like these URLs are relative to the output directory instead of the directory containing the entry point. This can be a problem even without the `--entry-names` option. I believe this is also the case for JavaScript code too. Using the public path feature is the standard way to get this to work. I'll try to get these relative paths to work though.

Mostly a note for myself: The fundamental issue is that each URL token in each file is currently assigned a single URL, but fixing this requires the same URL token potentially being assigned different URLs when it's imported into different entry points, since the directories containing that entry point may be different. This will need to change for this to be fixed.

EHadoux:
Yes, I can confirm it's also the case with TS code. I have a tree like:
```
src
  |
  +- Dir1
       |
       + DirA
       |  |
       |  + index.ts
       |
       + DirB
         |
         + index.ts
         + file.txt
```
With ```DirB/index.ts``` importing ```file.txt```.
My build script contains (amongst others):
```js
{ outdir: './bins', outbase: './src', loader: { '.txt': 'file' }, bundle: true }
```
but I don't use ```---entry-names```.

My output tree looks like:
```
bins
  |
  +- Dir1
  |    |
  |    + DirA
  |    |  |
  |    |  + index.js
  |    |
  |    + DirB
  |      |
  |      + index.js
  |      + file.js
  + file.HASH.txt
```

PR #1196, opened 2021-04-25
support for [dir] in AssetNames

daaku:


Rada8732 (approved):

evanw:
This would be nice to add. However, this is being done at the wrong phase and will also need to be redone when #1044 is fixed. So I'm going to close this PR and instead solve this request and #1044 together.

PR #1229, opened 2021-04-30
Add primitive types

seahindeniz:
I'm not sure why the Define option has only `string` type for values. I think it'll be fine to support other types as well
image

### DIFF
--- a/lib/shared/types.ts
+++ b/lib/shared/types.ts
@@ -24,7 +24,7 @@ interface CommonOptions {
 
   jsxFactory?: string;
   jsxFragment?: string;
-  define?: { [key: string]: string };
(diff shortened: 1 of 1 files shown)


evanw:
Strings are the only supported type because the string contains the code to use for replacement, not the value to use for replacement. This is a deliberate design decision and allows replacement to be more general than just literal values. For example, you can use `define` to replace something with a JavaScript identifier. If esbuild worked the way you are suggesting, there would be no way to distinguish between replacing with an identifier and replacing with a string. So I'm not open to this change. See the documentation for more information: https://esbuild.github.io/api/#define.

seahindeniz:
I see. In that case, I'll close the PR. Thanks for the clarification!

PR #1234, opened 2021-05-01
Use `sourceRoot` as base for `sources` paths

eduardoboucas:
When using sourcemaps, the paths in `sources` are relative to the output file. This assumes that it's always possible to navigate from the output file to each input file using relative paths, which is not the case when the input and the output files are on different drive letters on Windows.

With this PR, if a `sourceRoot` path is provided _and_ it's an absolute path, it will be used as the base for the relative paths in `sources`.

_Example:_

```
C
├─ input/
│  ├─ in.js
D
├─ output/
│  ├─ out.js
```

```js
{
  sourcemap: true,
  sourceRoot: "C:\input"
}
```

Currently, `sources` will contain `../../C/input/in.js`, which is not a valid path.

With this PR, paths are resolved relatively to `sourceRoot` (if defined), so `sources` will contain "in.js".

This will also generate shorter paths in `sources`, since the common path segments can be removed. This is one of the main reasons of using a `sourceRoot` property, as per [the spec](https://sourcemaps.info/spec.html):

> An optional source root, useful for relocating source files on a server or removing repeated values in the “sources” entry.  This value is prepended to the individual entries in the “source” field.

### DIFF
--- a/internal/bundler/linker.go
+++ b/internal/bundler/linker.go
@@ -4995,7 +4995,15 @@ func (c *linkerContext) generateSourceMapForChunk(
 		// Modify the absolute path to the original file to be relative to the
 		// directory that will contain the output file for this chunk
 		if item.path.Namespace == "file" {
-			if relPath, ok := c.fs.Rel(chunkAbsDir, item.path.Text); ok {
+			basePath := chunkAbsDir
+
+			// If "SourceRoot" is an absolute path, use it as the base for the
+			// relative paths
+			if sourceRootIsAbs := c.fs.IsAbs(c.options.SourceRoot); sourceRootIsAbs {
+				basePath = c.options.SourceRoot
+			}
+
+			if relPath, ok := c.fs.Rel(basePath, item.path.Text); ok {
 				// Make sure to always use forward slashes, even on Windows
 				item.prettyPath = strings.ReplaceAll(relPath, "\\", "/")
 			}
--- a/scripts/js-api-tests.js
+++ b/scripts/js-api-tests.js
@@ -442,6 +442,33 @@ let buildTests = {
     assert.strictEqual(json.sourcesContent[1], content)
   },
 
+  async sourceMapWithDifferentDirectories({ esbuild, testDir }) {
+    const inputDir = path.join(testDir, "input");
+    const outputDir = path.join(testDir, "output");
+    await mkdirAsync(inputDir, { recursive: true });
+    await mkdirAsync(outputDir, { recursive: true });
+    const input = path.join(inputDir, "in.js");
+    const output = path.join(outputDir, "out.js");
+    const content = "exports.foo = 123";
+    await writeFileAsync(input, content);
+    await esbuild.build({
+      entryPoints: [input],
+      outfile: output,
+      sourcemap: true,
+      sourceRoot: inputDir,
+    });
+    const result = require(output);
+    assert.strictEqual(result.foo, 123);
+    const outputFile = await readFileAsync(output, "utf8");
+    const match = /\/\/# sourceMappingURL=(.*)/.exec(outputFile);
+    assert.strictEqual(match[1], "out.js.map");
+    const resultMap = await readFileAsync(output + ".map", "utf8");
+    const json = JSON.parse(resultMap);
+    assert.strictEqual(json.version, 3);
+    assert.strictEqual(json.sources[0], path.relative(inputDir, input));
+    assert.strictEqual(json.sourcesContent[0], content);
+  },
+
   async resolveExtensionOrder({ esbuild, testDir }) {
     const input = path.join(testDir, 'in.js');
     const inputBare = path.join(testDir, 'module.js')


eduardoboucas:
I feel that I should add a bit more context to the rationale behind this PR.

The path for each source should be computed by prepending its `sources` entry with `sourceRoot`, and I think esbuild should do its best to ensure that the resulting sourcemap respects this.

So, with this PR, the follow happens depending on the value of `sourceRoot`:

- **`sourceRoot` is missing**: The items in `sources` will be relative to the output file, which respects the spec because _«If the sources are not absolute URLs after prepending of the “sourceRoot”, the sources are resolved relative to the SourceMap»_. No changed functionality in this case.

- **`sourceRoot` is not an absolute path**: The user is responsible for ensuring that the paths in `sources` after prepending `sourceRoot` will point to the input files, since esbuild can't infer any meaning from `sourceRoot` (e.g. it can be an URL). No changed functionality in this case.

- **`sourceRoot` is an absolute path**: In this case, the only way to guarantee that prepending `sourceRoot` to each source correctly points to the input files is to make `sources` relative to `sourceRoot`. This is the only change in functionality.

As a consequence of this, it's possible to have shorter entries in `sources` by removing repeated path trails and, more importantly, it fixes the issue outlined above with different drive letters on Windows.

evanw:
Some thoughts:

* This seems like a breaking change, so if it were to go in then it would have to wait until the next breaking change release.

* The absolute path detection could lead to false-positives on non-Windows platforms. For example, I expect it would be common for a URL such as `/assets/js` to be used as a source root. That would be mistaken as an absolute path on non-Windows platforms which would then incorrectly try to compute the relative path from the output directory to the `/assets/js` directory on the file system. The intent is probably to have the path stay relative to the output directory but use the source root to prefix it with the URL `/assets/js` in the browser. I'm not sure how to address this issue.

eduardoboucas:
Good point. Perhaps a good first step would be to do this only if `fs.volumeName` for the input and output files are different, which means that we're in the problematic case of different drive letters on Windows?

If so, it could be considered a minor change since there's no risk of false positives on non-Windows platforms _and_ we'd be fixing incorrect behaviour for Windows platforms. 

What do you think?

jridgewell:
`--sources-root` flag was meant to affect the _output_ sourcemap, not the inputs. Input sourcemaps have their own `"sourcesRoot": "path"` field in JSON, which is where their sources are relative to.

Using `--sources-root` on both the inputs and output will lead to doubling (`path/path/foo.js`), which is probably part of the reason this PR has to be limited to absolute paths only.

eduardoboucas:
> `--sources-root` flag was meant to affect the _output_ sourcemap, not the inputs. Input sourcemaps have their own `"sourcesRoot": "path"` field in JSON, which is where their sources are relative to.
> 
> Using `--sources-root` on both the inputs and output will lead to doubling (`path/path/foo.js`), which is probably part of the reason this PR has to be limited to absolute paths only.

I'm not sure what you're responding to, but I wasn't suggesting using `sourceRoot` on both inputs and outputs. I was suggesting that when your sources are `["../../parent/child/file1.js", "../../parent/child/file2.js"]` and `sourceRoot: "/parent"`, then the `sources` would become `["child/file1.js", "child/file2.js"]`. There's no duplication.

Unless I misunderstood your comment?

jridgewell:
I'm basing this from the currently implemented code. https://github.com/evanw/esbuild/pull/1234/files#diff-7c136e0e2ecad99566cc25cf7268b0537dadfe387e46eea3b7d82886e8875d84R5002-R5004 is checking if `c.options.SourceRoot` is an absolute path, and using that as a base if so.

`c.options.SourceRoot` is the `--source-root` option, which is used in the output sourcemap to set `"sourceRoot": "path"`.  See https://github.com/evanw/esbuild/blob/master/internal/bundler/linker.go#L5008-L5011, where it's added to the output sourcemap. So when you resolve paths from this output sourcemap, you concat the `sourceRoot` and `sources[i]` together. With this code change, both `sourceRoot` and `sources[i]` in the output map will contain, eg, `/root`, so the combined would be `/root/root/file.js`, which is the doubling I'm talking about.

eduardoboucas:
> So when you resolve paths from this output sourcemap, you concat the `sourceRoot` and `sources[i]` together. With this code change, both `sourceRoot` and `sources[i]` in the output map will contain, eg, `/root`, so the combined would be `/root/root/file.js`, which is the doubling I'm talking about.

If `sourceRoot` is `/root` and `sources[i]` is originally `/root/file.js`, the source will become `file.js` when you resolve it relatively to `sourceRoot`, so when you concatenate `sourceRoot` and `sources[i]` you'll get `/root/file.js`, as expected. I still don't understand how you'd end up with that duplication.

You can see this [here](https://github.com/evanw/esbuild/pull/1234/files#diff-5933b2d1756b532424d65e132d3b17584a8dcc599d5e7e314dc8ede3fe5022dfR468) in the test I've added, where the `sources` item will be just `in.js` and `sourceRoot` will be `{TEST-DIR}/input`, which after concatenation will become `{TEST-DIR}/input/in.js`.

jridgewell:
Ohh, you're removing `/root` from `sources[i]` with a `fs.Rel` call. I missed that part, I thought you were adding it.

eduardoboucas:
I have addressed @evanw's point about false-positives on Windows paths in https://github.com/evanw/esbuild/pull/1234/commits/b51b8f5d95daead05a42f13c21e69454599e0bb2. Instead of using `fs.IsAbs` to detect absolute paths, I'm checking for the `file` scheme. This will unequivocally identify a filesystem path, according to [RFC 3986](https://tools.ietf.org/html/rfc3986).

PR #1236, opened 2021-05-02
Improve esbuild’s DX during performance profiling

iamakulov:
When you’re profiling esbuild’s bundle initialization, you might stumble upon a whole bunch of `(anonymous)` functions:

Screen Shot 2021-05-02 at 20 43 53

Frequently, many of these functions are esbuild’s own functions which don’t have a name:

Screen Shot 2021-05-02 at 20 43 59

This PR gives these functions a name and improves the debugging experience. Here’s how the first flamegraph looks after the change:

Screen Shot 2021-05-02 at 20 46 32

Minification / bundle size were not considered. Also, if you’d prefer an alternative function name, I’d be happy to update the PR!

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -142,10 +142,10 @@ func code(isES6 bool) string {
 		}
 
 		// This is for lazily-initialized ESM code
-		export var __esm = (fn, res) => () => (fn && (res = fn(fn = 0)), res)
+		export var __esm = (fn, res) => function __esmWrapper() { return fn && (res = fn(fn = 0)), res }
 
 		// Wraps a CommonJS closure and returns a require() function
-		export var __commonJS = (cb, mod) => () => (mod || cb((mod = {exports: {}}).exports, mod), mod.exports)
+		export var __commonJS = (cb, mod) => function __commonJSWrapper() { return mod || cb((mod = {exports: {}}).exports, mod), mod.exports }
 
 		// Used to implement ES6 exports to CommonJS
 		export var __export = (target, all) => {


evanw:
> Minification / bundle size were not considered

This is indeed the reason why these are arrow functions. Are you profiling minified or unminified code?

iamakulov:
> Are you profiling minified or unminified code?

Unminified! I don’t have any DX expectations when debugging minified code.

iamakulov:
> Ideally each CommonJS wrapper would also be named with something that communicates which module it represents

That’s true! The second half of `(anonymous)` functions in the screenshot above is this:

```js
__commonJS((exports, module) => { ... }))
```

If that function’s name matched the file name, that’d be great. (Perhaps an algorithm similar to what esbuild uses for entrypoint names may be used? To avoid having lots of functions called `index`.)

```js
// ../../../node_modules/semver/index.js
var require_1 = __commonJS(function semver(exports, module) {})

// ../../../node_modules/semver/getHash.js
var require_2 = __commonJS(function getHash(exports, module) {})
// or?
var require_2 = __commonJS(function semver_getHash(exports, module) {})
// or?
var require_2 = __commonJS(function semver_getHash(exports, module) {})
```

For comparison, webpack uses full module paths (relative to the project directory) for chunk names during development. This corresponds to [`optimization.moduleIds: 'named'`](https://webpack.js.org/configuration/optimization/#optimizationmoduleids):

```js
/******/ (() => { // webpackBootstrap
/******/ 	var __webpack_modules__ = ({

/***/ "./a/index.js":
/*!********************!*\
  !*** ./a/index.js ***!
  \********************/
/***/ ((__unused_webpack_module, __webpack_exports__, __webpack_require__) => {

"use strict";
__webpack_require__.r(__webpack_exports__);
/* harmony export */ __webpack_require__.d(__webpack_exports__, {
/* harmony export */   "a": () => (/* binding */ a)
/* harmony export */ });
const a = 5;


/***/ }),

/***/ "./node_modules/moment/locale/af.js":
/*!******************************************!*\
  !*** ./node_modules/moment/locale/af.js ***!
  \******************************************/
/***/ (function(__unused_webpack_module, __unused_webpack_exports, __webpack_require__) {

//! moment.js locale configuration
//! locale : Afrikaans [af]
//! author : Werner Mollentze : https://github.com/wernerm

;(function (global, factory) {
    true ? factory(__webpack_require__(/*! ../moment */ "./node_modules/moment/moment.js")) :
   0
}(this, (function (moment) { 'use strict';
...
```

evanw:
I'm planning to address this by also naming the per-module function wrappers, and by making sure that this feature doesn't impact non-minified size. My PR for that is here: #1244.

PR #1238, opened 2021-05-02
fix(deno): close all subprocess resources

lucacasonato:
PR is split into three independent commits with comments for easy reviewing :-)

### DIFF
--- a/scripts/deno-tests.js
+++ b/scripts/deno-tests.js
@@ -1,97 +1,74 @@
 // To run this, you must first build the Deno package with "make platform-deno"
 import * as esbuild from '../deno/mod.js'
-import * as path from 'https://deno.land/std@0.93.0/path/mod.ts'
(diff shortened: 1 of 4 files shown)


scripts/deno-tests.js
```diff
@@ -1,97 +1,74 @@
...
+  })
+  const result = await import(path.toFileUrl(output))
+  asserts.assertStrictEquals(result.default, true)
+})
 
-  if (!allTestsPassed) {
-    console.error(`❌ deno tests failed`)
-    Deno.exit(1)
-  } else {
-    console.log(`✅ deno tests passed`)
```
evanw, line 89: I had this in there because I run all tests in parallel and without this it's hard to see which top-level test failed. That's ok though; I can add this back.

evanw:
Thanks for contributing this! I appreciate it.

Issue #660, opened 2021-01-11
`inject` has higher priority than `define`


remorses:
The identifiers for `options.define` should be replaced before injecting modules in `options.inject`

Another way to put this: `define` identifiers should have higher priority than exports in `inject`

This would prevent many injections from running, for example replacing occurrences of `process.env.VAR` with define could skip the injection of a `./process.js` module

Currently the injections are run first, which also means that the identifiers in `define` are never used if they are properties of an object exported from a module in `inject`

Reproduction: https://github.com/remorses/esbuild-inject-overrides-defines

geoffharcourt:
The situation where this has been most relevant for us is that we need to define `process.cwd` as a function for some of our dependencies (just one that no-ops), so we have to pick a static set of `process.env` values from the injected `process` as a result.

PR #1261, opened 2021-05-07
Apply defines before injected global variables

rtsao:
Resolves https://github.com/evanw/esbuild/issues/660

This seemed like the most straightforward way of addressing this, would love feedback


internal/js_parser/js_parser.go
```diff
@@ -9508,8 +9508,10 @@ func (p *parser) isDotDefineMatch(expr js_ast.Expr, parts []string) bool {
 				return false
 			}
 
-			// The last symbol must be unbound
-			return p.symbols[result.ref.InnerIndex].Kind == js_ast.SymbolUnbound
+			p.ignoreUsage(result.ref)
```
rtsao, line 9511: If we replace an injected symbol with a define, we need to explicitly mark it as unused (as `p.findSymbol` has the effect of marking usage). Otherwise, the injected import won't be eliminated.

Calling `p.ignoreUsage` feels a bit hacky. I also suppose that `ignoreUsage` should be invoked closer to the actual replacement callsite rather than in `isDotDefineMatch`, which doesn't necessitate that a replacement actually occurs...

evanw:
I agree that #660 would be good to solve. However, it seems like a breaking change to me in that this may cause subtle and unintentional behavior changes. So I'd like to wait on this until the next round of breaking changes (i.e. the next minor version bump, since esbuild is <1.0.0).

evanw:
Thanks! I appreciate you taking the time to fix this.

PR #1285, opened 2021-05-14
feat: support type-only ImportEqualsStmt

g-plane:
This is a TypeScript 4.2 feature:

```ts
import type React = require('react')
```

[TypeScript playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnASighgY3gXjihQEcBXYYgCgHJjcYaBKIA)

Ref: https://github.com/microsoft/TypeScript/pull/41573


evanw:
Looks reasonable. Thanks for the contribution!

PR #1310, opened 2021-05-24
fix(css): single hyphen shouldn't be treated as css identifier

### LINKED ISSUE #1308: CSS minification Bug
The double quotes in the css selector has been removed as part of minification.
This behaviour occurs only for "-" (hyphen) character

Frame 27

lbwa:
close #1308 

## CSS spec

In [CSS Selector level 4](https://drafts.csswg.org/selectors-4/#attribute-selectors): 

> Attribute values must be [`<ident-token>`](https://drafts.csswg.org/css-syntax-3/#ref-for-typedef-ident-token%E2%91%A0)s or [`<string-token>`](https://drafts.csswg.org/css-syntax-3/#ref-for-typedef-string-token%E2%91%A0)

According to  [`<ident-token>`](https://drafts.csswg.org/css-syntax-3/#ref-for-typedef-ident-token%E2%91%A0) rules,  single hyphen `-` without quotes is not valid `ident-token` which means `"-"` shouldn't be unwrapped as `-`. On the other hand, `"-"` is a valid [`<string-token>`](https://drafts.csswg.org/css-syntax-3/#ref-for-typedef-string-token%E2%91%A0) and a valid part of CSS attribute selector.

Extra helpful informations about `CSS escape` in [w3c](https://www.w3.org/International/questions/qa-escapes#cssescapes)

cc @evanw

### DIFF
--- a/internal/css_parser/css_parser_test.go
+++ b/internal/css_parser/css_parser_test.go
@@ -225,6 +225,7 @@ func TestEscapes(t *testing.T) {
 	expectPrinted(t, ":pseudo(cl\\,ss) {}", ":pseudo(cl\\,ss) {\n}\n")
 
 	// SSAttribute
+	expectPrinted(t, "[attr=\"-\"] {}", "[attr=\"-\"] {\n}\n")
 	expectPrinted(t, "[\\61ttr] {}", "[attr] {\n}\n")
 	expectPrinted(t, "[\\2c attr] {}", "[\\,attr] {\n}\n")
 	expectPrinted(t, "[\\,attr] {}", "[\\,attr] {\n}\n")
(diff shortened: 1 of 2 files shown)


evanw:
Thanks for the fix! Looks like the underlying reason is that `utf8.DecodeRuneInString` treats the end of the string as the U+FFFD replacement character which is being interpreted here as a non-ASCII code point in the "name-start code point" rule. I'll fix that part myself.

lbwa:
> Thanks for the fix! Looks like the underlying reason is that `utf8.DecodeRuneInString` treats the end of the string as the U+FFFD replacement character which is being interpreted here as a non-ASCII code point in the "name-start code point" rule. I'll fix that part myself.

Thanks for your explanations

PR #1331, opened 2021-05-31
Introduce UMD as a new output format

gwuhaolin:
cherry-pick from https://github.com/evanw/esbuild/pull/513 with conflicts resolved


pateketrueke:
Definitely having no UMD support breaks legacy projects on some users, so much thanks @gwuhaolin for your effort on this dutie!

pateketrueke:
For the record, you can achieve this in a _hacky way_ by using the `footer.js` option with a snippet like this, e.g.

```js
(function (root, factory) {
  if (typeof define === 'function' && define.amd) {
    define(factory);
  } else if (typeof module === 'object' && module.exports) {
    module.exports = factory();
  } else {
    root.YourExportedModuleAsIIFE = factory();
  }
}(typeof self !== 'undefined' ? self : this, () => YourExportedModuleAsIIFE));
```

Of course, it would be better to having UMD support without having to worry about this nuance.

stakhov:
rebase and try to push this request to master @gwuhaolin 

by the way thanks a lot

lizy0329:
I can't find any information about umd in official document, why hasn't it been merged? Is there some problem?

spencercap:
agreed! merge ETA? 
thanks @prantlf 👏 + @gwuhaolin 👏

mourner:
This is the only feature preventing me from switching from Rollup to ESBuild as the author of [50+ open source JS libraries](https://github.com/mourner/projects). Gently bringing this to your attention @evanw 🙏

mistic100:
really would like this to be added

mistic100:
@gwuhaolin does this PR supports externals ?

for example https://github.com/mistic100/Photo-Sphere-Viewer/blob/master/dist/photo-sphere-viewer.js#L7-L11 (this is rollup.js output)

DualWield:
why is this pull request still open?

PR #1338, opened 2021-06-01
fix(linker): add missing esm flag

### LINKED ISSUE #1323: without __esModule property when format esm module cause load module error
from vite [issue](https://github.com/vitejs/vite/issues/3399) i think it's esbuild problem

### Describe the bug

when the thirdparty module is esm format, esbuild manage it without add `__esModule` property. but webpack will add `__esModule` property.

for examle, the source code is

```js
// swiper.esm.js
export { default as Swiper, default } from './esm/components/core/core-class';
export { default as Virtual } from './esm/components/virtual/virtual';
//...
```

After esbuild format ,will output

```js
import {default as default2, default as default3} from "./esm/components/core/core-class";
import {default as default4} from "./esm/components/virtual/virtual";
export {
  default14 as A11y,
  default3 as default
};

```

without `__esModule` property.

it will be cause an error. if a module build by webpack, the source code as below

```js
//  react-id-swiper sourceCode
import Swiper from 'swiper';
new Swiper()
```
after webpack build , output code will add helper function `__importDefault`
```js
// react-id-swiper
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
var swiper_1 = __importDefault(require("swiper"));
new swiper_1.default(swiperNodeRef.current, object_assign_1.default({}, props))
```

but because of `__esModule` is undefined, `__importDefault` function will add default property once again cause the return value `__importDefault(require("swiper")).default` is not the current object will be error

image
image
 

### Reproduction

https://github.com/zhangyuang/vite-react-swiper-error

### System Info

vite/2.3.2 darwin-x64 node-v12.18.3

### additional

when i add `__esModule` property in source code manually, it can be executed succeed like

```js
export { default as Swiper, default } from './esm/components/core/core-class';
export { default as Virtual } from './esm/components/virtual/virtual';
export const __esModule = true 
```

In this case，maybe vite can add `__esModule` in esbuild plugin . for example

```js
build.onLoad({ filter: /.*/, namespace: 'swiper' }, ({ path: id }) => {
        return {
          loader: 'js',
          resolveDir: root,
          contents: `export * from "filepath"
          export {default} from "filePath"
          export const __esModule = true 
          `
        }
      })
```

### LINKED ISSUE #1333: [CJS-ESM interop] Behavior of requiring ESM from CJS is different with `@rollup/plugin-commonjs`
Requiring ESM from CJS + TypeScript `__importDefault` helper results in a nested structure, all the exports of the ESM file are wrapped in the `.default` property of the required variable.

See https://github.com/sodatea/cjs-require-esm-tests.git for a minimal reproduction.

Rollup fixes this issue with https://github.com/rollup/plugins/pull/552

It is also the cause of https://github.com/evanw/esbuild/issues/1323

lbwa:
fix #1333, fix #1323

## why

This pr is related to a backward-incompatible change from 0.11.0([changelog](https://github.com/evanw/esbuild/blob/master/CHANGELOG.md#0110)).

esbuild now generates transformed code like this:

```js
var esm_file_exports = {};
__export(esm_file_exports, {
  default: () => defaultValue, // caution
  foo: () => foo
});
var foo;
var init_esm_file = __esm(() => {
  foo = bar();
});
```

The problem is `esm_file_exports`  will be treated as a commnjs export when there is no __esModule flag. As for __importDefault, our esm default export will be treated as a normal commonjs export named `default` that we need to interop default export manually. 

so I suggest adding __markAsModule evaluation for making `esm_file_exports ` include a __esModule flag. This is used to make __importDefault **interop default esm export automatically**, instead of manually importation.

```diff
var esm_file_exports = {};
+ __markAsModule(esm_file_exports)
__export(esm_file_exports, {
  default: () => defaultValue,  // caution
  foo: () => foo
});
var foo;
var init_esm_file = __esm(() => {
  foo = bar();
});
```

## related pr/issues/commit

- https://github.com/evanw/esbuild/commit/f4a730886fee0425c648dab1c6744bc9b1a10a3a


guybedford:
Please let's not add markers to ES modules. Rather lets fix this for commonjs importers with better branding checks for ESM.

lbwa:
> Please let's not add markers to ES modules. Rather lets fix this for commonjs importers with better branding checks for ESM.

Looking forward to a better solution.

guybedford:
@lbwa I still don't understand the original issue - how is Webpack loading an ES module external in the first place? That is the mechanism that will be responsible for the real ES module interop in this case.

evanw:
> Please let's not add markers to ES modules. Rather lets fix this for commonjs importers with better branding checks for ESM.

This only comes up when importing ESM from CommonJS and that isn't something that any JS engine supports, so there's no precedent from JS engines to guide the implementation. The way this has historically worked is via Babel which adds `__esModule` so it seems reasonable to me to add it here like this. This is the standard "branding check for ESM."

> I still don't understand the original issue - how is Webpack loading an ES module external in the first place? That is the mechanism that will be responsible for the real ES module interop in this case.

You don't need Webpack for this. Another example is TypeScript, which is often used in a one-file-at-a-time mode where all imports are external:

```ts
import x from 'y'
console.log(x)
```

That gets converted into this [by the TypeScript compiler](https://www.typescriptlang.org/play?module=1#code/JYWwDg9gTgLgBADzgMyhEcDkBPTAoAYwgDsBnCAGwFMA6CiAcwAoEBKPIA):

```js
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
Object.defineProperty(exports, "__esModule", { value: true });
const y_1 = __importDefault(require("y"));
console.log(y_1.default);
```

So if the file `'y'` is an ES module that is converted to CommonJS by the bundler, `y_1` will become `{ "default": y_exports }` and if `y_exports` has a `default` property, you will then need to use `y_1.default.default` to access it.

moltar:
Broke this package: https://github.com/jawj/zapatos/issues/92

Not sure where the problem is, just adding it for reference :)

guybedford:
@evanw that is fine as long as the flag remains fully internal and doesn't spill into ESM interfaces imported from native ESM, which was my only concern here.

bjornstar:
It also broke this package: https://github.com/Chevrotain/chevrotain/issues/1537

Both packages are using `gen-esm-wrapper` and hopefully it is fixed in `v1.1.2` which filters `__esModule` from exports.

PR #1342, opened 2021-06-02
fix(linker): order of css imported from js

dmitrage:
This change fixes CSS files appearing in bundle in reverse order when importing from JS. Test case from https://github.com/evanw/esbuild/issues/465 can also be used for reproduction.


evanw:
It's unfortunate that I didn't catch this regression. Thanks for the fix.

PR #1385, opened 2021-06-19
feat: mangle Infinity

Gusted:
This change will shave some bytes when the code is using `Infinity`. The ECMA definition of divide can return (-)Infinity based of the second value, specifically when it's zero. And based on the first value it will be positive or negative. See: https://tc39.es/ecma262/#sec-numeric-types-number-divide

It's a small change and shouldn't break any code, as any existing code should play along nice with these values.

Simplified
```js
Infinity**2
1/0**2
```
Hereby ** has priority over the `/` and could break the code, but now what's nice about `1` and `0` they won't change at all*, doesn't matter which expressions you try to use with which numbers.

* Technically the `1` could be set to 0 with
```js
0 * 1/0 // First 0 * 1 will be executed in the javascript engine. Making it `0/0`
0 * Infinity
```
But in both cases they should NaN and thus a non-issue.

Regards,
Gusted


CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # Changelog
 
+## Unreleased
+
+* Minify Syntax `Infinity` to `1/0`
+
+	`--minify-syntax` will now minify the `Infinity` keyword to `1/0` which is less in bytes.
+
+	```js
+	// Orginial code
```
bathos, line 10:
```suggestion
	// Original code
```

Gusted: Thanks!

Gusted:
Seems like the Uglify tests are failing because `1/0.toString`(and more cases like that) isn't the same as `Infinity.toString()` and all other methods. I assume wrapping parentheses around `1/0` should do it, or ESBuild allows for complicated computations that can check if DNumber is being used for a function like `.toString()`

evanw:
The lack of parentheses causes other issues too. For example, turning `0/Infinity` into `0/1/0` incorrectly changes the result from `0` to `NaN`. Don't worry though, I'll take it from here.

Gusted:
Sure thing 👍🏽, Il be interested in the final (golang) code for this one.

evanw:
It looks like you have perhaps [disallowed edits from maintainers](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) on your branch. At least I am unable to push to it (I get permission denied). Can that setting be changed on your end?

Gusted:
image
Should be enabled...

evanw:
Ah sorry. I always forget that GitHub's handy command line instructions rename the branch to something else, which breaks this workflow. I figured it out.

internal/js_printer/js_printer_test.go
```diff
@@ -932,13 +932,60 @@ func TestAvoidSlashScript(t *testing.T) {
...
+
 	expectPrintedMangleMinify(t, "x = Infinity", "x=1/0;")
-	expectPrintedMangleMinify(t, "x = -Infinity+2", "x=-1/0+2;")
-	expectPrintedMangleMinify(t, "x = (-Infinity+2) / 3", "x=(-1/0+2)/3;")
-	expectPrintedMangleMinify(t, "1 + -Infinity", "1+-1/0;")
-	expectPrintedMangleMinify(t, "1 - -Infinity", "1- -1/0;")
+	expectPrintedMangleMinify(t, "x = -Infinity", "x=-1/0;")
+	expectPrintedMangleMinify(t, "x = (Infinity).toString", "x=(1/0).toString;")
+	expectPrintedMangleMinify(t, "x = (-Infinity).toString", "x=(-1/0).toString;")
+	expectPrintedMangleMinify(t, "x = Infinity ** 2", "x=(1/0)**2;")
```
Gusted, line 982: I've might overlook this but `1/0 ** 2` will result in Infinity as well. Can the parentheses be "turned off" for different `level`'s? Or will that cause more problems 😅.

evanw: The expressions `(a / b) ** c` and `a / (b ** c)` have different ASTs. Having minification turn one into the other is not generally correct, even if it's correct in that specific case. For example, changing `console.log(0 ** Infinity ** 1);` into `console.log(0 ** 1 / 0 ** 1);` changes the output from `0` to `NaN`.

Gusted: Got it, thanks for the explanation 🎉

PR #1390, opened 2021-06-22
Add support for shorten transform/translate3d

steambap:
Add support for minify `transform: translate3d(0, 0, tz)` to `translateZ(tz)`


YurySolovyov:
can the same trick be done with `translateX` and/or `translateY` ?

evanw:
There are many other transform identities that can be used for minification here. I’m already working on the rest.

steambap:
Hi @YurySolovyov,
Yes, there are many tricks for transform including translateXY. You can take a look at some other tricks here: https://github.com/cssnano/cssnano/tree/master/packages/postcss-reduce-transforms

PR #1440, opened 2021-07-12
Also should be configurable

yisar:
I encountered an error and could not prompt to redefine, then I used replace and it worked well.

```shell
typeerror cannot redefine property default
```

```js
content.replace('{ get: all[name], enumerable: true }',`{ get: all[name], enumerable: true, configurable: true }`)
```

Actually I don’t like Object.defineproperty, but if you have to, add a parameter to cover more situations.

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -171,7 +171,7 @@ func code(isES6 bool) string {
 		export var __export = (target, all) => {
 			__markAsModule(target)
 			for (var name in all)
-				__defProp(target, name, { get: all[name], enumerable: true })
+				__defProp(target, name, { get: all[name], enumerable: true, configurable: true })
 		}
 		export var __reExport = (target, module, desc) => {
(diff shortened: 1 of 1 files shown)


evanw:
Doing this would violate the JavaScript specification: https://tc39.es/ecma262/#sec-module-namespace-exotic-objects. The specification says `configurable` should be `false`. So the change you are proposing is not valid.

yisar:
@evanw My project is indeed more complicated, because it is executed in a new function, I am not the reason for the error, but I do know that the problem can be solved.

https://github.com/berialjs/berial/blob/master/src/sandbox.ts#L124

Speaking of standards, I am very annoyed by the cjs format code generated by esbuild now, which contains redundant runtime dirty code. As an alternative, I suggest abandoning this packaging method.

I think it will be better to implement scope hoisted in cjs, just like what esbuild does for esm now, instead of using a mapping.

PR #1449, opened 2021-07-16
Abort esbuild if stdin is closed when serving

josevalim:
Most command line tools (cat, fsevents, etc) and JS CLI watchers (postcss,
webpack, etc) listen to stdin and abort when the stdin is closed (for example,
by sending Ctrl+D). This is important because there is no guarantee the
process who invokes esbuild will terminate it, but it is guaranteed the stdin
will be closed once the parent dies.

This pull requests watches the stdin when serving and exits accordingly,
in order to provide a user convenience and avoid zombie processes.

PS1: note I chose to always enable stdin watching (postcss does the same)
but you may want to gate this behaviour behind a flag (like webpack).

PS2: I noticed that you can communicate with esbuild via stdio but I am
not sure if this communication is enabled when serving. If it is, then this
needs to be revisited.

PS3: thanks for esbuild!

### DIFF
--- a/cmd/esbuild/main.go
+++ b/cmd/esbuild/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"runtime/debug"
 	"strings"
@@ -261,11 +262,17 @@ func main() {
 				}
 			}
 
-			// Disable the GC since we're just going to allocate a bunch of memory
-			// and then exit anyway. This speedup is not insignificant. Make sure to
-			// only do this here once we know that we're not going to be a long-lived
-			// process though.
-			if !isServe {
+			if isServe {
+				// Watch stdin and abort in case it is closed
+				go func() {
+					io.ReadAll(os.Stdin)
+					os.Exit(exitCode)
+				}()
+			} else {
+				// Disable the GC since we're just going to allocate a bunch of memory
+				// and then exit anyway. This speedup is not insignificant. Make sure to
+				// only do this here once we know that we're not going to be a long-lived
+				// process though.
 				debug.SetGCPercent(-1)
 			}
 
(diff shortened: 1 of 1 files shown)


evanw:
Thanks for the PR. I wanted to give you a heads up that I'm moving across the country this week and so will probably not have time to look at this (or anything related to esbuild) until some time after my move. Probably sometime next week instead. But I've seen this and I agree it will be good to fix. It's on my list of things to look into as soon as I get time after my move.

evanw:
I just looked into this a little. Here are some other issues for reference:

* Webpack
    * https://github.com/webpack/webpack/issues/1308
    * https://github.com/webpack/webpack/pull/1311
* Parcel (they decided to not do it since there's a workaround)
    * https://github.com/parcel-bundler/parcel/pull/1004
* Brunch (they added this and it broke something)
    * https://github.com/brunch/brunch/issues/920
    * https://github.com/brunch/brunch/pull/922
    * https://github.com/brunch/brunch/issues/998
    * https://github.com/brunch/brunch/pull/1003
* PostCSS
    * https://github.com/postcss/postcss-cli/pull/358

josevalim:
Good summary!

I had forgotten why Brunch introduced a separate flag for listening to stdin but one of your links resurfaced the reason: if you wait for stdin to be closed then it can no longer run as a background process. That sounds like a fair use case, so I would be glad to gate this feature behind an option if you prefer!

evanw:
I'd like to avoid adding a flag for this. Right now I'm thinking of trying to only monitor stdin when it's not a TTY. That should keep the ability to use esbuild as a background process while hopefully also enabling the Erlang VM child process scenario without a flag. The only thing it won't enable is letting you use Ctrl+D to stop watch mode but you can just do Ctrl+C instead so that doesn't seem important.

Can you confirm that the previous commit works with the Erlang VM in the way that you wanted?

josevalim:
@evanw nice solution! It worked as expected, thank you! :heart:

Issue #1414, opened 2021-07-02
emitFile like Rollup has


ianberdin:
Hi, 
Thank you, you did amazing work!

I'm begging you, can we add the emitFile function?
This is very important for the implementation of the plugin: html as entrypoint.

For an example:
```js 
let htmlEntryPlugin = {
  name: 'htmlEntry',
  setup( build ) {
    build.onLoad({ filter: /.html$/, namespace: 'html-entry' }, ( args ) => {

      const file = getFile(args.path)
      const scripts = getScripts(file)

      for ( const scriptSrc of sripts ) {
        build.emitFile({ path: scriptSrc })
      }

      return {
        contents: '',
        loader: 'js',
      }
    })

    build.onEnd(result => {
      result = injectScriptsToHtmlEnty(result)
      return result
    })
  },
}
```

ianberdin:
Sorry, is something there?

DavidArchibald:
I've tried to do my due diligence by looking around so hopefully this isn't necroing something with a known answer.

First I'd just like to add on my own thoughts. This would also be helpful for handling importing Handlebars files and transitive dependencies there (since they can themselves contain imports). Really it's important for custom entrypoints in general, something I've recently been looking at for, in essence, a custom manifest file (in specific it's for .

As I know Go and Typescript I'd be willing to put together a PR for this if that'd be helpful! From my vantage I _think_ it could be simple as well. I imagine that ideally `emitFile` (name can be bikeshedded) would function with the same input and output values as `build.resolve` but in addition actually put the file into the output.

Maybe with the metafile and the internals of ESBuild in general that would become very difficult? I am unsure. Though just to add another bike shedding comment, since they'd function with the same input and output maybe `emit` could just be added as an option to `build.resolve`?

PR #1460, opened 2021-07-19
Add "isEntryPoint" to OutputFile and "entryPoints" to metafile

jgoz:
The motivation for this PR is to support plugins and higher level tooling that operate on HTML.

- The `isEntryPoint` property on `OutputFile` will be `true` if the output corresponds to a user-defined entry point, but will be falsy otherwise (dynamic chunks, additional files, sourcemaps, etc.)
- The `entryPoints` array on `Metafile` contains the array of output files that correspond to user-defined entry points. This is different from the set of `metafile.outputs` where `entryPoint` is set because `entryPoints` will include CSS output files generated from JS entry points that import CSS

**Plugin case**
With the current Build API, it is not currently possible to reliably determine which output files should be referenced by `<script>` or `<link>` tags in an HTML file for all cases. The keys of `metafile.outputs` can be used to determine which output files were generated, but since the `entryPoint` property corresponds 1:1 with an input file, CSS outputs produced from JS imports cannot be determined. It's also not possible to map entry point basenames from the initial build options because the `entryNames` option might not include the name at all.

This is solved by the `entryPoints` array on metafile because it allows plugins to determine which output files correspond to application entry points without re-implementing internal logic or dependency graph traversal. If further details are needed about any of the contained entry points, they can be used as keys in the `metafile.outputs` object.

**Higher level tooling case**
While the `entryPoints` array on metafile would be sufficient for higher level tooling that wraps the Build API, adding `isEntryPoint` to OutputFile allows the same kind of introspection without enabling metafile generation.

With this option, it would be possible to parse an HTML file for entry points, pass them to the build API, and then reliably replace the paths in the HTML output.

### DIFF
--- a/cmd/esbuild/service.go
+++ b/cmd/esbuild/service.go
@@ -929,6 +929,9 @@ func encodeOutputFiles(outputFiles []api.OutputFile) []interface{} {
 		values[i] = value
 		value["path"] = outputFile.Path
 		value["contents"] = outputFile.Contents
+		if outputFile.IsEntryPoint {
+			value["isEntryPoint"] = true
+		}
 	}
 	return values
 }
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -2200,6 +2200,7 @@ func (b *Bundle) generateMetadataJSON(results []graph.OutputFile, allReachableFi
 	// Write outputs
 	isFirst = true
 	paths := make(map[string]bool)
+	var entryPoints []string
 	for _, result := range results {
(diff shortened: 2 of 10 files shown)


ianberdin:
Hi @jgoz!
Yea, I have the same problem.
It will be even easier using `emitFile()`. Rollup has it.

I made an issue: https://github.com/evanw/esbuild/issues/1414.

jgoz:
Not relevant after https://github.com/evanw/esbuild/commit/4bd03f3d1426acd58db7bed999c34687825d4251

PR #1468, opened 2021-07-24
feat: support module conditions when platform set to browser

hardfist:
when use esbuild to build [tslib](https://github.com/microsoft/tslib/blob/main/package.json#L31), esbuild seems not follow module field in conditions exports, which makes treeshaking not working

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -247,6 +247,7 @@ func NewResolver(fs fs.FS, log logger.Log, caches *cache.CacheSet, options confi
 	switch options.Platform {
 	case config.PlatformBrowser:
 		esmConditionsDefault["browser"] = true
+		esmConditionsDefault["module"] = true;
 	case config.PlatformNode:
 		esmConditionsDefault["node"] = true
 	}


evanw:
I assume it's similar to the `module` main field. In that case, it has nothing to do with the current platform so it shouldn't be configured via the platform setting. The `module` main field just means the source code is an ES module instead of a CommonJS module. Since esbuild supports both of those module formats, it can bundle code in either of those formats for either the browser or for node. From [esbuild's documentation](https://esbuild.github.io/api/#main-fields):

> For package authors: Some packages incorrectly use the `module` field for browser-specific code, leaving node-specific code for the `main` field. This is probably because node ignores the `module` field and people typically only use bundlers for browser-specific code. However, bundling node-specific code is valuable too (e.g. it decreases download and boot time) and packages that put browser-specific code in `module` prevent bundlers from being able to do tree shaking effectively. If you are trying to publish browser-specific code in a package, use the `browser` field instead.

If you want to use the `module` condition, can do that yourself without modifying esbuild by passing `--conditions=module` to esbuild (see https://esbuild.github.io/api/#conditions). I don't think it's a good idea to have esbuild enable this condition automatically since the `exports` field has [very specific behavior](https://nodejs.org/api/packages.html#packages_conditional_exports) about ES modules vs. CommonJS modules that respecting `module` by default would mess up. It seems more appropriate to require users to opt-in to that condition instead.

hardfist:
It seems that webpack supports module field in conditions field ，and treeshaking works

Andarist:
@evanw I think this is a stale PR that can be closed, what is requested here got essentially implemented in: https://github.com/evanw/esbuild/commit/5abe0715f9be662b182989d2f38a44c7c8b28a2d

PR #1482, opened 2021-07-30
refactor(deno): use denoflate instead of compress

lucacasonato:
deno.land/x/denoflate is about 10% smaller, and a lot more polished and
up to date than deno.land/x/compress.


evanw (approved):
Sounds good to me 👍

PR #1585, opened 2021-09-09
inline math.MaxInt constants (go 1.13 compatibility)

davezuko:
Since esbuild uses some newer Go API's it would be nice to accurately specify the minimum Go version for dependents. I ran into this while trying to upgrade esbuild to 0.12.26 while on Go 1.16, where Go would no longer compile with esbuild due to its [math.MaxInt usage](https://github.com/evanw/esbuild/blob/ab666a1d9697139ff9ad600c36f47592de016734/pkg/api/api_impl.go#L1758). I wish I could offer something more than this trivial change, but perhaps it will be helpful to others.

I'm far from a Go expert so if there are things to be careful with when updating this please let me know (i.e. I read through https://golang.org/doc/modules/gomod-ref and, weirdly, it sounds like Go _should_ have caught this...).

Thanks!

### DIFF
--- a/pkg/api/api_impl.go
+++ b/pkg/api/api_impl.go
@@ -3,7 +3,6 @@ package api
 import (
 	"fmt"
 	"io/ioutil"
-	"math"
 	"math/rand"
 	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


evanw:
It would be pretty straightforward to remove `math.MaxInt`. Should that be done instead?

davezuko:
@evanw absolutely, I thought about that as well. I proposed this change simply because I didn't know what else might be in esbuild that requires go > 1.13 and thought it might be worth the update regardless. The wise thing to do would probably be for me to downgrade go locally to 1.13 and see if any other issues crop up. I know that Go 1.16 was at least working until the addition of math.MaxInt.

davezuko:
Update the PR to just inline the math.MaxInt constants directly. That should be more palatable.

Issue #1579, opened 2021-09-07
Suggest another `__require`.


hyrious:
Original issue: egoist/tsup#369. All credits go to @mvrlin.

```js
// Before
var __require = (x) => {
  if (typeof require !== "undefined")
    return require(x);
  throw new Error('Dynamic require of "' + x + '" is not supported');
};

// After
var __require = (x) => {
  throw new Error('Dynamic require of "' + x + '" is not supported');
}

if (typeof require !== "undefined") {
  __require = require
}
```

### Motivation

The [tsup](https://github.com/egoist/tsup) implements code splitting for cjs by **bundling** the code to esm then **transform** them to cjs. In this approach some `import`s are transformed to `require`, but the original `require` in the code becomes `__require`, which prevents `require.resolve` (which becomes `__require.resolve`) to work properly.

Unless esbuild find out some way to do code splitting in cjs format, here is a simpler solution to replace the `__require`'s code without losing its purpose.

evanw:
Closing now that https://github.com/evanw/esbuild/issues/1558 has been merged.

AlexRMU:
What about
```js
import { createRequire } from "module";
const require = createRequire(import.meta.url);
```
?

Issue #946, opened 2021-03-10
WIP: ReferenceError: require is not defined is ESM Node projects


aral:
In my ESM Node project, [Place](https://github.com/small-tech/place) that also imports CommonJS modules (any any Node project these days probably does given ESM modules are not the norm yet),  I get the following error when I run the bundle:

```txt
ReferenceError: require is not defined
``` 

If I add the following line to the top of the bundle to create the require function, it runs:

```js
const require = createRequire(import.meta.url)
```

I don’t have a small, reproducible case yet but opening this issue early, as a work-in-progress, in case it helps without it.

zaydek:
I ran into something similar to this when playing with `const evaluator = new Function(code)`. I didn’t know you could do this but I found the technique here: https://github.com/digital-loukoum/esrun/blob/main/src/main.js#L63.

I don’t know how relevant this is for you but given I encountered the same error, I needed to do this: `const evaluator = new Function("require", code)` and then invoke the evaluator like this `evaluator(require)`. This seemed hacky to me so I’ve since changed my implementation, but I thought it was worth sharing.

So my guess is what `createRequire` is evaluating JavaScript in memory and therefore `require` becomes undefined as a side-effect. This is my best guess as to what’s going on.

aral:
@zaydek Thank you for the additional information… interesting. If that is the case, it doesn’t affect good ’ol `eval()` itself as [JSDB](https://github.com/small-tech/jsdb) uses it and that’s not creating a problem, so it must be limited to the `new Function()` method. Hmm.

Given that it’s a single line of code that needs to be added, I’m just going to inject that as part of my build process for the time being.

aral:
My hacky workaround for the moment is running this after the esbuild script in the npm build task:

```js
import fs from 'fs'
import path from 'path'
const __dirname = new URL('.', import.meta.url).pathname

const build = path.join(__dirname, '..', 'place.js')
const unpatchedBuild = fs.readFileSync(build, 'utf-8')
const patchedBuild = unpatchedBuild.replace('#!/usr/bin/env node', '#!/usr/bin/env node\n\nconst require = createRequire(import.meta.url)\n')
fs.writeFileSync(build, patchedBuild, 'utf-8')
```

popeindustries:
In case anyone else gets tripped up by this error when converting cjs to esm (where `require('native-node-module or module-marked-as-external')` preserves the `require` statement and throws a runtime error), the most elegant solution I've found is via use of the `banner` option:

[...]

PR #1604, opened 2021-09-14
Fix a bug in __require

lovasoa:
This reverts commit 1dea2db919c016bf0b11dcc6aa94a108a604591b,
which introduced an optimization based on the invalid assumption that the `require` function will always be either already defined when `__require` is defined, or never.

Fixes https://github.com/sveltejs/kit/issues/2400

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -117,7 +117,8 @@ func code(isES6 bool) string {
 		export var __name = (target, value) => __defProp(target, 'name', { value, configurable: true })
 
 		// This fallback "require" function exists so that "typeof require" can naturally be "function"
-		export var __require = typeof require !== 'undefined' ? require : x => {
+		export var __require = x => {
+			if (typeof require !== 'undefined') return require(x)
 			throw new Error('Dynamic require of "' + x + '" is not supported')
 		}
 


lovasoa:
This was an "optimization" that was introduced without any comment on the actual performance gains. Since `require` is in itself an expensive operation, I highly doubt removing an `if` from it leads to any measurable performance gain.

Conduitry:
For more context, when SvelteKit is producing a production build to run on a Node environment, it uses esbuild for the final bundling of the server-side app. If any of the bundled dependencies from `node_modules` are written in CJS and `require()` any Node builtins, esbuild leaves these as `require()`s in the final bundle, even when the whole app is being bundled to ESM. (https://github.com/sveltejs/kit/pull/1648#issuecomment-856160009)

To get around this, SvelteKit inserts code to define a global `require` function created using `createRequire`. This code is, apparently, ending up after esbuild's own runtime code, and so when `__require` now checks at define time (rather than at call time) whether `require` is defined, it isn't (yet), and so the global `require` shim SvelteKit defines is no longer used, and any (untranspiled) `require`s of Node globals in CJS dependencies of an ESM project end up as function calls that throw a runtime error.

The quickest way back to a working state would indeed probably be to revert this optimization in `__require`, but in the longer term it would be nice if this shim of SvelteKit's were not necessary and `require`s of Node builtins in CJS dependencies were transpiled into `import` statements when bundling to an ESM target. I haven't looked for whether there's already an issue here for that or any related discussion.

Conduitry:
#946 appears to be the issue to track the underlying thing that resulted in SvelteKit needing this `global.require` workaround at all. I haven't checked whether the other suggested workarounds in that issue (like using a `banner`) would help in this case without reverting the change this PR proposes to - i.e., does the banner end up before the `__require` definition.

Conduitry:
I should also note that the mechanism SvelteKit is using to define these globals has also since changed since my PR I linked above. As of https://github.com/sveltejs/kit/pull/1822 it is using the official 'inject' feature of esbuild, and it is this that is apparently not defining it in a way or in a place that the `__require` definition can see it.

hyrious:
A better solution would be not generating this require shim when bundling cjs dependencies in esm environment.

Currently this shim just act as a _side effect_.
The previous version can be safely omitted by rollup/terser because a function is not used.
However current version can not be removed, which causes these errors.

evanw:
This approach won't work. Just reverting the change will re-break #1579 which was trying to use `require.resolve`. That's the reason why the global `require` function is currently being preferred if it exists.

I suppose a solution that will work for both edge cases might be a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that forwards property accesses and calls to the global if it exists. That solution seems pretty complicated, however.

evanw:
Closing this in favor of the `Proxy` approach.

hyrious:
@evanw Hi, thank you for looking into this. Just be curious, is it possible to omit this `__require` when bundling cjs code that not using `require` in esm environment?
```js
// a.js, should be esm
import "./b"

// b.js, should be cjs
module.exports = 1

// esbuild a.js --bundle
var __require = ... // ← it is not in use, actually
var require_b = __commonJS({ ... })
var import_b = __toModule(require_b());
```
I don't know the implementation, but I guess a `__require` is always generated when linking a cjs module in esm env, even though it might not be used.

The very first implementation of `__require` is just a function, which will be safely removed by rollup/terser/esbuild.
The next 2 impls, actually depend on a runtime toplevel `require`'s value to work, which can not be removed, even though nothing use it. This tiny difference causes esbuild minify to include more useless codes.

- - -

Another possible solution would be not generating this shim at all, since you can always get an error when accessing a nonexist variable/function.

PR #1613, opened 2021-09-16
Add support for sha256 hashes

theodorton:
Esbuild will be one of the recommended bundlers to use with Rails 7 (through the [jsbundling-rails gem](https://github.com/rails/jsbundling-rails)), but due to how hashes are computed differently in esbuild and sprockets using code splitting is hard to configure right out of the box.

A patch has already been made to sprockets via https://github.com/rails/jsbundling-rails/issues/15 and https://github.com/rails/sprockets/pull/714. These changes make it possible for the asset pipeline in Rails to not re-digest files that already end with a SHA256 digests, and as Webpack already supports SHA256 it will make code splitting work in new Rails projects. This PR should add support for SHA256 in Esbuild as well (without breaking any backwards-compatibility). With a couple more changes to `jsbundling-rails`, it should make code splitting work for most users.

The new CLI parameter `--hash-function` and API/config option `HashFunction`, will let users determine whether to use the current Base32 hashes or SHA256.

A lot of the code changes are relevant for passing the options around internally in esbuild - so please let me know if it can be refactored or changed to fit the project guidelines.

The most important change is here:
https://github.com/onelittle/esbuild/blob/6d66286349245a3fbe2d6d2783a6d0bf4cb520d6/internal/bundler/linker.go#L377-L383

If anyone wants to use other hash functions like `md5` or `sha1`, these can now be added by changing the if to a switch statement.


theodorton:
@evanw Is there anything blocking from getting this PR merged? I'm happy to make any required changes. Integration with Rails 7 would benefit greatly from having a more predictable hash signature than base32.

nicbet:
I'd like to weigh in and voice my support for the proposed PR. Aside from Rails, having a flexible interface and a path forward for easily adding different fingerprinting methods for fragments and files is a great addition.

jtarchie:
It is not really clear from the PR. But why can't Rails change to support how esbuild does the hashing, rather than having esbuild change to support how Rails does it?

jtarchie:
Thanks for the explanation. I’m still trying to grasp though why do you want a tool to bend towards the hashing methods you have, rather than having the flexibility in your framework? There could be other tools that also don’t have the basing capability. Are you going to request those change too?

theodorton:
@jtarchie Base32 is not in the Ruby standard library, so it would require another dependency somewhere in the Rails stack to support esbuild. The current approach also relies on regex to detect hashes, which is simpler to do for hex characters. My motivation is to make the onboarding experience with esbuild in Rails as smooth as possible while making it possible to use modern bundling techniques like bundle splitting and asynchronous loading.  I know there has been development on the Rails side to make detecting hashes more flexible, I'll pursue an alternative approach that doesn't require this PR.

PR #1615, opened 2021-09-17
add --force-tree-shaking flag

### LINKED ISSUE #1610: minify: provide an option to enable treeshaking in single-file transform

yyx990803:
closes #1610

Please see committed changelog entry for details.


yyx990803:
Some details - I considered some other options without adding a new flag:

1. Add another enumerated value to `--tree-shake`, like `--tree-shake=force`. However, it would be ambiguous how it interacts with `--tree-shake=ignore-annotations`. Ideally, we want to keep the ability to force treeshaking but still ignore annotations.

2. Making it an implicit behavior when `--minify` is enabled, as in most cases minification implies removing unused code. However, this would technically be a breaking change so I tried to avoid it.

I'm open to any form of API adjustments as long as it allows us to minify a single file using `transform` with tree-shaking enabled.

yyx990803:
I just realized the current `--minify` behavior is already non-concatenation safe when processing a ESM input file.

Given this input:

```js
function foo() {}
export { foo }
```

Result of running `esbuild --minify`:

```js
function n(){}export{n as bar};
```

Notice `foo` has been renamed, which will break if another concatenated file expects to call `foo()`.

That is to say, **esbuild already assumes isolated scope when it detects ESM syntax**. With this in mind, I believe the correct behavior would be auto enabling tree-shaking when (1) `--minify` is enabled and (2) the input file is detected to be ESM. We can also add a separate `--minify-unused` flag if necessary.

yyx990803:
Closing in favor of #1617

PR #1617, opened 2021-09-18
enable tree-shaking when both --minify and --format flags are specified

### LINKED ISSUE #1610: minify: provide an option to enable treeshaking in single-file transform

yyx990803:
Also add dedicated --minify-unused flag. Please see committed changelog entry for more details.

closes #1610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+* Enable tree-shaking in transform mode when target format is specified ([#1610](https://github.com/evanw/esbuild/issues/1610))
+
+    Previously esbuild does not perform tree-shaking when transforming single files because esbuild cannot assume whether the file will be concatenated with other files, and thus it would be unsafe to remove unused bindings.
+    
+    However, when an output format is explicitly specified, it implies the file is a module so tree-shaking can be safely applied. In addition, the previous behavior when both `--minify` and `--format` are specified was already concatenation-unsafe, since esbuild renames top-level bindings in this case.
+
+    Before this release, there was no API that allows the user to force tree-shaking when transforming a single file. As of this release, esbuild will now enable tree-shaking when transforming a single file with both `--minify` and `--format` specified. This behavior can also be individually enabled with the `--minify-unused` flag.
+
+    This improvement is important for tools that intend to use esbuild as a non-bundling minifier to replace Terser - for example, a Vite or Rollup plugin that minifies ESM chunks produced by Rollup.
+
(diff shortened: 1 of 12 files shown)


evanw:
Thanks for the thoughtful issue and PRs. I do want to get this fixed and this PR looks like a good implementation of this specific approach. However, I'd still like to think through the alternate approach of augmenting `--tree-shaking=` instead of `--minify` myself first since that's where I'd expect people to look for this feature, and people may want this feature outside of the context of minification. You already thought through that in https://github.com/evanw/esbuild/pull/1615#issuecomment-921985853, copied below:

> I considered some other options without adding a new flag:
>
> 1. Add another enumerated value to `--tree-shake`, like `--tree-shake=force`. However, it would be ambiguous how it interacts with `--tree-shake=ignore-annotations`. Ideally, we want to keep the ability to force treeshaking but still ignore annotations.

I'm imagining that `--tree-shaking=` could be either unspecified, `false`, `true`, or `ignore-annotations`:

* Unspecified: the current behavior (`true` if bundling is enabled or if the format is `iife`, otherwise `false`)
* `false`: all unused code would be kept
* `true`: all unused code would be removed
* `ignore-annotations`: all unused code would be removed (ignoring annotations)

I don't think there's a conflict with `true` and `ignore-annotations` because "the ability to force treeshaking but still ignore annotations" could just be `ignore-annotations`. That is a subtle behavior change from the current behavior but I don't think it's an issue because `--tree-shaking=ignore-annotations` is currently meaningless if tree shaking isn't active, so having it activate tree shaking actually makes more sense to me. Let me know if you see any problems with this approach. If not, I think I may want to take that approach instead (although I'm still thinking about it).

yyx990803:
That sounds good to me! I avoided it mostly because I didn't want to change the behavior of ignore-annotations, but since you are ok with that then I think it's straightforward to just pass --tree-shaking=true.

Feel free to close this PR if you decide to go ahead with it, or I can modify this to reflect the change.

evanw:
I have now looked into this and I was wrong: `--tree-shaking=ignore-annotations` isn't actually currently meaningless when tree shaking is disabled. It still affects how [`/* @__PURE__ */` comments](https://esbuild.github.io/api/#pure) are interpreted when `--minify-syntax` is enabled, which can trigger dead-code removal within expressions. This is related to tree shaking but seems somewhat separate since "tree shaking" usually refers to declaration-level dead code removal and you may want dead-code removal within expressions without declaration-level dead code removal. So I think it's best to separate those two concepts. My PR for this change is here: #1625.

PR #1621, opened 2021-09-20
install using "optionalDependencies"

### LINKED ISSUE #286: Feature request: can config the registry url
https://github.com/evanw/esbuild/blob/2abe0243556294a5862c8c43c455e78f85d8a633/npm/esbuild/install.js#L20

idea：
config the `https://registry.npmjs.org` from `yarn config get registry` or `ENV`.

### LINKED ISSUE #291: Version 0.6.6 break the install with custom npm registry (e.g. cnpm)
0.6.6 support custom npm registry, but the `package/-/package-version.tgz` format is not support on [the cnpm registry](https://cnpmjs.org/).

* https://registry.npmjs.org/esbuild-linux-64/-/esbuild-linux-64-0.6.7.tgz (ok)
* https://r.cnpmjs.org/esbuild-linux-64/-/esbuild-linux-64-0.6.7.tgz (not found)

image

### LINKED ISSUE #319: npm install fails behind corporate proxy
After upgrading from 0.6.3 to 0.6.17 the install process of npm fails in post install step.
It turned out to be a proxy issue:

> set NODE_DEBUG=net,http,https
> npm install
> [...]
> HTTP 30356: SOCKET ERROR: connect ETIMEDOUT 104.16.18.35:443 Error: connect ETIMEDOUT 104.16.18.35:443
>     at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1144:16)
> Trying to download "https://registry.npmjs.org/esbuild-windows-64/-/esbuild-windows-64-0.6.17.tgz"
> Failed to download "https://registry.npmjs.org/esbuild-windows-64/-/esbuild-windows-64-0.6.17.tgz": connect ETIMEDOUT 104.16.18.35:443
> Install unsuccessful
> [...]
> npm ERR! Failed at the esbuild@0.6.17 postinstall script.
> [...]

The reason is that the fetch routine does direct http(s) connections.
https://github.com/evanw/esbuild/blob/master/lib/install.ts#L123

If there would be some way to use npm to download the dependencies that would be great.
Else you should provide a way to respect http_proxy and https_proxy env variables.

Right now I have to find a Windows machine that can do direct downloads and copy the files afterward.

### LINKED ISSUE #347: install.js should support HTTPS_PROXY environment variable
The [esbuild](https://www.npmjs.com/package/esbuild) module has below rudimentary `fetch` implementation that does not support fetching via HTTPS proxies which means the script will run for over 4 minutes waiting for the HTTPS request to timeout in environments without direct Internet connectivity.

I'd suggest switching to [make-fetch-happen](https://github.com/npm/make-fetch-happen) for this download which is a `fetch`-compatible module that `npm` itself uses and it has full proxy support.

````js
function fetch(url2) {
  return new Promise((resolve, reject) => {
    https.get(url2, (res) => {
      if ((res.statusCode === 301 || res.statusCode === 302) && res.headers.location)
        return fetch(res.headers.location).then(resolve, reject);
      if (res.statusCode !== 200)
        return reject(new Error(`Server responded with ${res.statusCode}`));
      let chunks = [];
      res.on("data", (chunk) => chunks.push(chunk));
      res.on("end", () => resolve(Buffer.concat(chunks)));
    }).on("error", reject);
  });
}
````

Related: https://github.com/evanw/esbuild/issues/319

I could probably send you a pull request but I could not find the source for this script in the repo.

### LINKED ISSUE #369: npm install esbuild got permission denied
```
[root@localhost ~]# npm install esbuild

> esbuild@0.6.30 postinstall /root/node_modules/esbuild
> node install.js

Trying to install "esbuild-linux-64" using npm
Failed to install "esbuild-linux-64" using npm: EACCES: permission denied, mkdir '/root/node_modules/esbuild/.install'
Trying to download "https://registry.npmjs.org/esbuild-linux-64/-/esbuild-linux-64-0.6.30.tgz"
/root/node_modules/esbuild/install.js:176
      throw e;
      ^

Error: EACCES: permission denied, open '/root/node_modules/esbuild/bin/esbuild'
    at Object.openSync (fs.js:465:3)
    at Object.writeFileSync (fs.js:1416:35)
    at /root/node_modules/esbuild/install.js:70:8
    at Generator.next (<anonymous>)
    at fulfilled (/root/node_modules/esbuild/install.js:6:24)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  errno: -13,
  syscall: 'open',
  code: 'EACCES',
  path: '/root/node_modules/esbuild/bin/esbuild'
}
npm WARN enoent ENOENT: no such file or directory, open '/root/package.json'
npm WARN root No description
npm WARN root No repository field.
npm WARN root No README data
npm WARN root No license field.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! esbuild@0.6.30 postinstall: `node install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the esbuild@0.6.30 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-09-05T15_42_39_991Z-debug.log
```

then i set env not download, the problem remains the same.

```
export ESBUILD_BIN_PATH_FOR_TESTS=/usr/local/bin/esbuild

[root@localhost ~]# npm install esbuild

> esbuild@0.6.30 postinstall /root/node_modules/esbuild
> node install.js

internal/fs/utils.js:269
    throw err;
    ^

Error: EACCES: permission denied, unlink '/root/node_modules/esbuild/bin/esbuild'
    at Object.unlinkSync (fs.js:1136:3)
    at installOnUnix (/root/node_modules/esbuild/install.js:172:8)
    at Object.<anonymous> (/root/node_modules/esbuild/install.js:213:3)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  errno: -13,
  syscall: 'unlink',
  code: 'EACCES',
  path: '/root/node_modules/esbuild/bin/esbuild'
}
npm WARN enoent ENOENT: no such file or directory, open '/root/package.json'
npm WARN root No description
npm WARN root No repository field.
npm WARN root No README data
npm WARN root No license field.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! esbuild@0.6.30 postinstall: `node install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the esbuild@0.6.30 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-09-05T15_58_04_212Z-debug.log
```

There is a very confused problem. Who can help me?

### LINKED ISSUE #547: Install using npm failed because can't remove install folder
```
Trying to install "esbuild-windows-64" using npm
Failed to install "esbuild-windows-64" using npm: EBUSY: resource busy or locked, rmdir 'C:\path_to_project\node_modules\.pnpm\esbuild@0.7.22\node_modules\esbuild\.install'
```

Maybe you can create install directory in system temp and ignore errors if can't remove?

### LINKED ISSUE #565: Postinstall ignore the project specific .npmrc - IMPORTANT FOR CI
It would be important that a postinstall consider the project specific configurations.

Inside a CI Pipeline we does not have root privileges - so we only can change the local .npmrc.

But the npm install of OS specific esbuild binary does not use this settings.

### LINKED ISSUE #789: Different strategy for installing platform-specific binaries

### LINKED ISSUE #921: Sometimes can't install esbuild with 3rd-party registry
~~~
events.js:292
      throw er; // Unhandled 'error' event
      ^

Error: spawn F:\projects\test\my-vue-app\node_modules\esbuild\esbuild.exe ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:269:19)
    at onErrorNT (internal/child_process.js:465:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
    at onErrorNT (internal/child_process.js:465:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn F:\\projects\\test\\my-vue-app\\node_modules\\esbuild\\esbuild.exe',
  path: 'F:\\projects\\test\\my-vue-app\\node_modules\\esbuild\\esbuild.exe',
  spawnargs: [ '--service=0.8.55', '--ping' ]
}
~~~

The problem was solved after I rolled back esbuild to 0.8.54

### LINKED ISSUE #1193: Configurable registry for binary downloads
We use an internal artifactory instance that contains all of the packages we we need to build and those machines don't have access to the internet. Currently when we try to use esbuild it fails in CI since the installer attempts to download the public registry.

Would you be open to a PR that added npmrc support for defining the registry? I was thinking something like the [`sass_binary_site`](https://www.npmjs.com/package/node-sass#binary-configuration-parameters) config.

### LINKED ISSUE #1270: support install without post-install
It seems that esbuild relys `post-install` to download the real bin for arch-spec bin file, which cause some troubles

*  npm7 problem
* not support offline install

I wonder whether is there better ways to handle this?

### LINKED ISSUE #1382: Fail to install last version on windows 64 behind proxy
last version
```
npm install esbuild
npm ERR! code 1
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c node install.js
npm ERR! Trying to install "esbuild-windows-64" using npm
npm ERR! Failed to install "esbuild-windows-64" using npm: Command failed: npm install --loglevel=error --prefer-offline --no-audit --progress=false esbuild-windows-64@0.12.9
npm ERR! npm ERR! code ETARGET
npm ERR! npm ERR! notarget No matching version found for esbuild-windows-64@0.12.9.
npm ERR! npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! npm ERR! notarget a package version that doesn't exist.
npm ERR!
npm ERR!
npm ERR! Trying to download "https://registry.npmjs.org/esbuild-windows-64/-/esbuild-windows-64-0.12.9.tgz"
npm ERR! Failed to download "https://registry.npmjs.org/esbuild-windows-64/-/esbuild-windows-64-0.12.9.tgz": unable to get local issuer certificate
npm ERR! Install unsuccessful
```
previous one
```
npm install esbuild@0.12.8
4 packages are looking for funding
run `npm fund` for details
found 0 vulnerabilities
```

### LINKED ISSUE #1422: Failed to download binaries from private registry in CI environment without internet access and yarn v1 as package manager
We use an internal nexus instance that contains all of the packages we we need to build and those machines don't have access to the internet. Currently when we try to use esbuild it fails in CI since the installer attempts to download the public registry. We are using `yarn v1` as package manager.

None of solutions below worked:

`export NPM_CONFIG_REGISTRY`
`export npm_config_registry`
`yarn config set registry`
`yarn config set registry -g`
`.yarnc`
`.npmrc`

Yarn issues yarnpkg/yarn#5854 and yarnpkg/yarn#5683

Only worked solution was pre-install packages via npm install with --registry option for cache warmup. Also in docker there seems to be problem with process.env.TMPDIR and that require('os').tmpdir() is returns cwd instead of temp directory path (npm/npm#4531).

Can install script be extended with support for custom env variable for registry and with `--unsafe-perm` flag for correct process.env.TMPDIR value?

### LINKED ISSUE #1450: Support repositories other than npmjs
We currently can't install esbuild on our CI platform. I suspect that the initial attempt to install from npm fails as it's trying to create a temporary folder which will fail on our jenkins build due to file system access permissions. Installing from npmjs then fails as access to that is blocked. We have our own private npm repository.

The way this is usually handled is that an environment variable is used to specify where compiled binaries can be downloaded from.

### LINKED ISSUE #1485: Install script uses wrong registry
I tried a fresh vite project which has esbuild as dep. When I try to run `npm install` i see that the install script is trying to download something from a different registry than npm is configured with.
Is there a way to override the url from the install.js so it is asking the correct registry for the package?
```
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c node install.js
npm ERR! Trying to install "esbuild-windows-64" using npm
npm ERR! Failed to install "esbuild-windows-64" using npm: Command failed: npm install --loglevel=error --prefer-offline --no-audit --progress=false esbuild-windows-64@0.12.17
npm ERR! npm ERR! code ETARGET
npm ERR! npm ERR! notarget No matching version found for esbuild-windows-64@0.12.17.
npm ERR! npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! npm ERR! notarget a package version that doesn't exist.
npm ERR!
npm ERR! npm ERR! A complete log of this run can be found in:
npm ERR! npm ERR!     C:\Users\<user>\AppData\Local\npm-cache\_logs\2021-08-01T20_58_36_415Z-debug.log
npm ERR!
npm ERR! Trying to download "https://registry.npmjs.org/esbuild-windows-64/-/esbuild-windows-64-0.12.17.tgz"
npm ERR! Failed to download "https://registry.npmjs.org/esbuild-windows-64/-/esbuild-windows-64-0.12.17.tgz": connect ETIMEDOUT
```

### LINKED ISSUE #1546: fs.writeFileSync options.mode not work when file already exists
```bash
echo 1 > a && ls -l a
# -rw-rw-r-- 1 admin admin 2 Aug 24 18:29 a

node -e "fs.writeFileSync('./a', 'echo world', {mode: 0o755})" && ls -l a
# -rw-rw-r-- 1 admin admin 10 Aug 24 18:29 a

./a
# bash: ./a: Permission denied
```

Even we specify the mode to be `0o755`, `fs.writeFileSync` still doesn't change to file mode which is `0o644`.

So the installer which uses `fs.writeFileSync` [L63](https://github.com/evanw/esbuild/blob/master/lib/npm/install.ts#L63) won't make `esbuild` executable.

This is a default behavior on POSIX when trying to write Buffer to an existing file.

[PR#1547](https://github.com/evanw/esbuild/pull/1547) shall fix it.

### LINKED ISSUE #1574: esbuild postinstall scripts hangs randomly on WSL
When doing a `npm install` inside my devcontainer (`FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:14`/`:16` esbuild always causes the entire terminal to freeze.

Nothing changes when I clear the npm cache, restart WSL, restart/reset docker.

This is all logs I have available:

```
node ➜ /app/backend (02ed521 ✗) $ npm cache clean --force
npm WARN using --force I sure hope you know what you are doing.
node ➜ /app/backend (02ed521 ✗) $ npm ci
npm WARN prepare removing existing node_modules/ before installation

> esbuild@0.12.25 postinstall /app/backend/node_modules/esbuild
> node install.js
```

After this I have to manually stop the script again as it simply never ends.

How could I debug this properly?

### LINKED ISSUE #1609: Installation fails when using a private npm registry that requires _authToken.
I'm using AWS CodeArtifact as a npm registry and it fails to install esbuild. More specifically, the install of esbuild itself succeeds, but the execution of `install.js` defined in 'postinstall' fails. In `install.js`, node's `child_process` executes another `npm install`, but at this time, '_authToken' defined in the `.npmrc` placed in the project does not take effect. (status 401 is returned.) On the other hand, the registry specification in the `.npmrc` is still enabled in `child_process`, so it seems that the `.npmrc` is not completely ignored.

My `.npmrc` is defined as follows:
```
registry=https://my_domain-111122223333.d.codeartifact.us-west-2.amazonaws.com/npm/my-cli-repo/
//my_domain-111122223333.d.codeartifact.us-west-2.amazonaws.com/npm/my_repo/:_authToken=eyJ2ZX...
//my_domain-111122223333.d.codeartifact.us-west-2.amazonaws.com/npm/my_repo/:always-auth=true
```

I think this is not a problem specific to AWS CodeArtifact, but a common problem with private registries that require 'authToken'.

Has anyone else experienced a similar case? 
Also, is there any solution to this problem?
Thanks in advance.

evanw:
This PR changes esbuild's installation strategy in an attempt to improve compatibility with edge cases such as custom registries, custom proxies, offline installations, read-only file systems, or when post-install scripts are disabled.

**The old installation strategy** manually downloaded the correct binary executable in a [post-install script](https://docs.npmjs.com/cli/v7/using-npm/scripts). The binary executable is hosted in a separate platform-specific npm package such as [`esbuild-darwin-64`](https://www.npmjs.com/package/esbuild-darwin-64). The install script first attempted to download the package via the `npm` command in case npm had custom network settings configured. If that didn't work, the install script attempted to download the package from https://registry.npmjs.org/ before giving up. This was problematic for many reasons including:

* Not all of npm's settings can be forwarded due to npm bugs such as https://github.com/npm/cli/issues/2284, and npm has said these bugs will never be fixed.
* Some people have configured their network environments such that downloading from https://registry.npmjs.org/ will hang instead of either succeeding or failing.
* The installed package was broken if you used `npm --ignore-scripts` because then the post-install script wasn't run. Some people enable this option so that malicious packages must be run first before being able to do malicious stuff.

**The new installation strategy** automatically downloads the correct binary executable using npm's `optionalDependencies` feature to depend on all esbuild packages for all platforms but only have the one for the current platform be installed. This is a built-in part of the package manager so my assumption is that it should work correctly in all of these edge cases that currently don't work. And if there's an issue with this, then the problem is with the package manager instead of with esbuild so this should hopefully reduce the maintenance burden on esbuild itself. Changing to this installation strategy has these drawbacks:

* Old versions of certain package managers (specifically npm and yarn) print lots of useless log messages during the installation, at least one for each platform other than the current one. These messages are harmless and can be ignored. However, they are annoying. There is nothing I can do about this. If you have this problem, one solution is to upgrade your package manager to a newer version.

* Installation will be significantly slower in old versions of npm, old versions of pnpm, and all versions of yarn. These package managers download all packages for all platforms even though they aren't needed and actually cannot be used. This problem has been fixed in npm and pnpm and the problem has been communicated to yarn: https://github.com/yarnpkg/berry/issues/3317. If you have this problem, one solution is to use a newer version of npm or pnpm as your package manager.

* This installation strategy does not work if you use `npm --no-optional` since then the package with the binary executable is not installed. If you have this problem, the solution is to not pass the `--no-optional` flag when installing packages.

* There is still a small post-install script but it's now optional in that the `esbuild` package should still function correctly if post-install scripts are disabled (such as with `npm --ignore-scripts`). This post-install script optimizes the installed package by replacing the `esbuild` JavaScript command shim with the actual binary executable at install time. This avoids the overhead of launching another `node` process when using the `esbuild` command. So keep in mind that installing with `--ignore-scripts` will result in a slower `esbuild` command.

Despite the drawbacks of the new installation strategy, I believe this change is overall a good thing to move forward with. It should fix edge case scenarios where installing esbuild currently doesn't work at all, and this only comes at the expense of the install script working in a less-optimal way (but still working) if you are using an old version of npm. So I'm going to switch installation strategies and see how it goes.

The platform-specific binary executables are still hosted on npm in the same way, so anyone who wrote code that downloads builds from npm using the instructions here should not have to change their code: https://esbuild.github.io/getting-started/#download-a-build. However, note that these platform-specific packages no longer specify the `bin` field in `package.json` so the `esbuild` command will no longer be automatically put on your path. The `bin` field had to be removed because of a collision with the `bin` field of the `esbuild` package (now that the `esbuild` package depends on all of these platform-specific packages as optional dependencies).

Related issues:

Fixes #286
Fixes #291
Fixes #319
Fixes #347
Fixes #369
Fixes #547
Fixes #565
Fixes #789
Fixes #921
Fixes #1193
Fixes #1270
Fixes #1382
Fixes #1422
Fixes #1450
Fixes #1485
Fixes #1546
Fixes #1547
Fixes #1574
Fixes #1609


lukeed:
Is there a way to test this? Are you looking for people to test this in their toolchains?

evanw:
Testing this out would be much appreciated. I have published a test package called `esbuild-experimental-optdeps@0.12.10` which should be an accurate representation of how the `esbuild` package will behave after this PR lands. I have tested this with npm, pnpm, and yarn and everything seems to work.

However, it would be good to know if this works with unusual configurations such as custom registries, HTTP proxies, offline installations, etc. Right now I'm assuming it'll work since `optionalDependencies` is a long-standing existing feature of the npm ecosystem, but there could always be package manager bugs around this feature. The tracking issue for this PR is #789 so you can read that thread for more details and discussion about this approach.

I'm assuming that asking people to test this out prior to it landing isn't sufficient to flush out any issues with the approach since I've already asked before and very few people have actually tried it out. So as long as there aren't any blocker problems, I'm planning on merging this as a breaking change release to get wider feedback. But testing this out would still be very much appreciated as it's obviously better to uncover blocker issues before it lands.

lukeed:
Worked for me! I'm running pnpm@5.11.1 and it worked flawlessly in a regular repo and a monorepo environment (with multiple packages depending on `esbuild-experimental-optdeps@0.12.10` directly).

My concern was that it was going to an experience similar to `@swc/core` installation, which downloads _all_ binaries and then only keeps the matching package, but that wasn't the case 🎉

hardfist:
> Worked for me! I'm running pnpm@5.11.1 and it worked flawlessly in a regular repo and a monorepo environment (with multiple packages depending on `esbuild-experimental-optdeps@0.12.10` directly).
> 
> My concern was that it was going to an experience similar to `@swc/core` installation, which downloads _all_ binaries and then only keeps the matching package, but that wasn't the case 🎉

pnpm has already fixes the bug since latest v6 verison and downloads only partial binaries of @swc/core

lukeed:
@hardfist right, but I wanted to see if the non-patched version treated this PR the same way

danvim:
Can confirm this (0.13.2) is working for me behind proxy. Coming from #319 🥳

silverwind:
Can also confirm this as working in cases where it previously failed with a network timeout trying to contact the registry directly. 🎉

Issue #1591, opened 2021-09-10
The export name "__esModule" is reserved and cannot be used


heypiotr:
Recently, I've been attempting to bundle modules from jspm.io (via a custom esbuild plugin which implements imports from URLs and import maps), and I've been running into this error a lot:

> error: The export name "__esModule" is reserved and cannot be used (it's needed as an export marker when converting ES module syntax to CommonJS)

Here's an example module where I've run into this for the first time: https://jspm.dev/npm:react-calendly@2.2.1!cjs

It all boils down to something like this:

```
// MRE.js

const exports = {}
Object.defineProperty(exports, "__esModule", { value: true })
export default exports

const __esModule = exports.__esModule
export { __esModule }
```

It's that named __esModule export that upsets esbuild -- if I comment it out, it builds fine.

The way I understood it from reading https://github.com/esnext/es6-module-transpiler/issues/85, the magic __esModule export was introduced so that when you transpile an ESM module to CJS, and do `const foo = require("esm-foo-transpiled-to-cjs")`, it'll know that you really mean `const foo = require("esm-foo-transpiled-to-cjs").default`. However, I belive JSPM's doing things the other way, i.e., it transpiles CJS modules to ESM. So I'm not sure if their usage of the __esModule export is correct.

We've had a brief conversation about this with @guybedford on the JSPM Discord, but we decided to move it here. I'm curious, what's the story behind this "__esModule is reserved" check/error in esbuild?

guybedford:
On the JSPM side, the issue here is that JSPM will convert CJS into ESM, and in doing so follow the Node.js semantics for what exports to expose.

Per the Node.js ESM CJS wrapper semantics, `__esModule` is a valid ESM export of a CJS module wrapper, and hence is reflected in the converted ESM of the CJS module, which then doesn't build with esbuild.

heypiotr:
>On the JSPM side, the issue here is that JSPM will convert CJS into ESM, and in doing so follow the Node.js semantics for what exports to expose.

Ah, I think I understand now. So e.g. react-calendly's dist/index.js does `Object.defineProperty(exports, '__esModule', { value: true });`. JSPM then iterates over all properties on `exports`, and converts them to ESM exports. So it sees `__esModule`, and treats it as any other export, hence generating the "offending" __esModule export. Which if course is not really "offending" by any standards.

[...]

PR #1622, opened 2021-09-20
Allow ES Modules to export __esModule

heypiotr:
This is my stab at fixing https://github.com/evanw/esbuild/issues/1591, as described in https://github.com/evanw/esbuild/issues/1591#issuecomment-920169259:

> Would it be at all feasible to make __export + __markAsModule a bit smarter => if there's already a correct __esModule marker in all, just let it be? If it's incorrect, I guess the only option would be to throw a runtime error, which would be a slight regression compared to the current compile-time error. But maybe that's good enough?

This PR removes the "__esModule is reserved" check, and updates `__export` + `__markAsModule` to only define `__esModule` if it doesn't exist yet.

I think this might only be problematic if the author of the script uses the `__esModule` export "incorrectly" (i.e., as anything other than an ESM marker). Originally, as stated in my comment, I was hoping to have `__markAsModule` detect that (`__esModule` exists, but evaluated to anything but `true`) and throw a runtime error... But turns out at the time of `__export`, we don't yet know the value of `__esModule`:

```
$ cat MRE.js
export const __esModule = "foo"

$ echo 'console.log(require("./MRE.js"))' | ./esbuild --bundle
(() => {
  var __defProp = Object.defineProperty;
  var __hasOwnProp = Object.prototype.hasOwnProperty;
  var __markAsModule = (target) => !__hasOwnProp.call(target, "__esModule") && __defProp(target, "__esModule", { value: true });
  var __require = typeof require !== "undefined" ? require : (x) => {
    throw new Error('Dynamic require of "' + x + '" is not supported');
  };
  var __esm = (fn, res) => function __init() {
    return fn && (res = (0, fn[Object.keys(fn)[0]])(fn = 0)), res;
  };
  var __export = (target, all) => {
    for (var name in all)
      __defProp(target, name, { get: all[name], enumerable: true });
    __markAsModule(target);
  };

  // MRE.js
  var MRE_exports = {};
  __export(MRE_exports, {
    __esModule: () => __esModule
  });
  var __esModule;
  var init_MRE = __esm({
    "MRE.js"() {
      __esModule = "foo";
    }
  });

  // <stdin>
  console.log((init_MRE(), MRE_exports));
})();
```

You can see `__export` happens before `init_MRE()`, and it's `init_MRE` that actually sets the value of `__esModule`.

So no runtime error for now, and if we still want it, I could use a hint/ideas on how to approach that.


vovacodes:
@evanw are there concerns preventing this PR from being merged, it helps tremendously in adoption of URL imports and Import Maps (dependencies provided by jspm.io) by allowing folks opting-in for bundling (using esbuild) while keeping the unbundled code valid in the browser. We are actively using this workflow at Framer and find it extremely compelling. 

I read the whole discussion in the original issue and while the topic is extremely controversial, it doesn't look like the solution proposed by @heypiotr in this PR is hurting any runtime compatibility.

WDYT?

PR #1624, opened 2021-09-22
Add NetBSD amd64 binary

gdt:
This is labeled amd64 in descriptive text rather than "64-bit" because
NetBSD supports a number of 64-bit CPU types, and NetBSD people will
be looking for "amd64" which is our word for that CPU and machine
type.  Otherwise, this is cut and paste from the FreeBSD x86_64
support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+* Experimental support for esbuild on NetBSD ([#1624](https://github.com/evanw/esbuild/pull/1624))
+
+    With this release, esbuild now has a published binary executable for [NetBSD](https://www.netbsd.org/) in the [`esbuild-netbsd-64`](http://npmjs.com/package/esbuild-netbsd-64) npm package, and esbuild's installer has been modified to attempt to use it when on NetBSD. Hopefully this makes installing esbuild via npm work on NetBSD. This change was contributed by [@gdt](https://github.com/gdt).
+
+    ⚠️ Note: NetBSD is not one of [Node's supported platforms](https://nodejs.org/api/process.html#process_process_platform), so installing esbuild may or may not work on NetBSD depending on how Node has been patched. This is not a problem with esbuild. ⚠️
+
 ## 0.13.5
 
 * Improve watch mode accuracy ([#1113](https://github.com/evanw/esbuild/issues/1113))


gdt:
I am a go and esbuild newbie, and am here because I need this for zwavejs.  I ran into a test issue that I think it is unrelated, so this hasn't been tested as well as it normally should be.   esbuild does build and run fine on NetBSD amd64.

npm/esbuild-netbsd-64/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "esbuild-netbsd-64",
+  "version": "0.13.0",
+  "description": "The NetBSD amd64 binary for esbuild, a JavaScript bundler.",
+  "repository": "https://github.com/evanw/esbuild",
+  "license": "MIT",
+  "os": [
+    "netbsd"
```
evanw, line 8: I'm not sure this will work because `netbsd` is not currently listed as a platform that node can return. From https://nodejs.org/api/process.html#process_process_platform:

> The `process.platform` property returns a string identifying the operating system platform on which the Node.js process is running.
>
> Currently possible values are:
>
> * `'aix'`
> * `'darwin'`
> * `'freebsd'`
> * `'linux'`
> * `'openbsd'`
> * `'sunos'`
> * `'win32'`
>
> The value `'android'` may also be returned if the Node.js is built on the Android operating system. However, Android support in Node.js [is experimental](https://github.com/nodejs/node/blob/HEAD/BUILDING.md#androidandroid-based-devices-eg-firefox-os).

This matters because esbuild's API uses `process.platform` to determine which package to take the binary executable from:

https://github.com/evanw/esbuild/blob/b8068370b3408679d53627b5e2aae7309c50239a/lib/npm/node-platform.ts#L35

This key is used to look up into the map of known packages here:

https://github.com/evanw/esbuild/blob/b8068370b3408679d53627b5e2aae7309c50239a/lib/npm/node-platform.ts#L10-L30

If `process.platform` can't return `netbsd` then running esbuild's JavaScript API on NetBSD won't work. You are welcome to [build esbuild from source](https://esbuild.github.io/getting-started/#build-from-source) from NetBSD because Go supports NetBSD, but I don't think it makes sense to publish a NetBSD binary executable to npm if node doesn't support NetBSD.

gdt:
Thank you for taking the time to review and the useful comments.   I will look into that with the NetBSD node people; we do have node in pkgsrc and people use it, so there could be some changes that ought to be upstreamed but haven't been.

If it's excess cognitive load  and you want to close the PR, and have me reopen if I figure out how to resolve all of this, that's totally fine with me.

gdt:
I am happy to rebase if someone with merge authority is willing to hit merge, or if it's otherwise helpful.

bsiegert:
The Net sad package collection has a node.js package but as far as I know,
the NetBSD support is maintained as a patch outside of the upstream node.js
tree.

This means that officially, there is no NetBSD support in node, however in
practice, there are a number of users on NetBSD.

Evan Wallace ***@***.***> schrieb am Di., 12. Okt. 2021,
17:51:

gdt:
Following up on what @bsiegert said, "supported" is a complicated word.  In practice, anyone who wants to run node on NetBSD will use it via pkgsrc (at least 95%+ of them), and when installed that way, it worked for me.    pkgsrc has patches for a vast number of things to deal with various portability issues, and while in theory people are supposed to file those upstream, there are always things that haven't been, because people don't get to it, or it turns out to be difficult, or something else.

I did review the patches in pkgsrc for node, and I didn't find any that I understood to change the platform list.  But I can't conclude from that there aren't any because it's complicated and I don't fully understand.

I don't know if there's a requirement for esbuild for there to be offiicial support, or just that when people try to use it in the usual way it works.   I can try to build node not under pkgsrc and see what happens.

evanw:
Ok, I will move forward and start publishing the package with a note that node doesn't support NetBSD. Based on this conversation, it sounds like npm's install mechanism could potentially be patched such that installing a package with `optionalDependencies` works even with packages containing `"os": ["netbsd"]` despite that being an invalid value, so it's worth a try.

evanw:
Version 0.13.6 of esbuild should theoretically have NetBSD support. Can someone on this thread try it out and let me know if it works or not?

PR #1625, opened 2021-09-22
separate "ignore annotations" from "tree shaking"

### LINKED ISSUE #1610: minify: provide an option to enable treeshaking in single-file transform

evanw:
This PR introduces a breaking change that gives you more control over when tree shaking happens ("tree shaking" here refers to declaration-level dead code removal). Previously esbuild's tree shaking was automatically enabled or disabled for you depending on the situation and there was no manual override to change this. Specifically, tree shaking was only enabled either when bundling was enabled or when the output format was set to `iife` (i.e. wrapped in an immediately-invoked function expression). This was done to avoid issues with people appending code to output files in the `cjs` and `esm` formats and expecting that code to be able to reference code in the output file that isn't otherwise referenced.

You now have the ability to explicitly force-enable or force-disable tree shaking to bypass this default behavior. This is a breaking change because there is already a setting for tree shaking that does something else, and it has been moved to a separate setting instead. The previous setting allowed you to control whether or not to ignore manual side-effect annotations, which is related to tree shaking since only side-effect free code can be removed as dead code. Specifically you can annotate function calls with `/* @__PURE__ */` to indicate that they can be removed if they are not used, and you can annotate packages with `"sideEffects": false` to indicate that imports of that package can be removed if they are not used. Being able to ignore these annotations is necessary because [they are sometimes incorrect](https://github.com/tensorflow/tfjs/issues/4248). This previous setting has been moved to a separate setting because it actually impacts dead-code removal within expressions, which also applies when minifying with tree-shaking disabled.

### Old behavior

* CLI
    * Ignore side-effect annotations: `--tree-shaking=ignore-annotations`
* JS
    * Ignore side-effect annotations: `treeShaking: 'ignore-annotations'`
* Go
    * Ignore side-effect annotations: `TreeShaking: api.TreeShakingIgnoreAnnotations`

### New behavior

* CLI
    * Ignore side-effect annotations: `--ignore-annotations`
    * Force-disable tree shaking: `--tree-shaking=false`
    * Force-enable tree shaking: `--tree-shaking=true`
* JS
    * Ignore side-effect annotations: `ignoreAnnotations: true`
    * Force-disable tree shaking: `treeShaking: false`
    * Force-enable tree shaking: `treeShaking: true`
* Go
    * Ignore side-effect annotations: `IgnoreAnnotations: true`
    * Force-disable tree shaking: `TreeShaking: api.TreeShakingFalse`
    * Force-enable tree shaking: `TreeShaking: api.TreeShakingTrue`

Fixes #1610
Closes #1617

### DIFF
--- a/cmd/esbuild/main.go
+++ b/cmd/esbuild/main.go
@@ -72,6 +72,8 @@ var helpText = func(colors logger.Colors) string {
   --footer:T=...            Text to be appended to each output file of type T
                             where T is one of: css | js
   --global-name=...         The name of the global for the IIFE format
+  --ignore-annotations      Enable this to work with packages that have
+                            incorrect tree-shaking annotations
   --inject:F                Import the file F into all input files and
                             automatically replace matching globals with imports
   --jsx-factory=...         What to use for JSX instead of React.createElement
@@ -105,8 +107,7 @@ var helpText = func(colors logger.Colors) string {
   --sourcemap=external      Do not link to the source map with a comment
   --sourcemap=inline        Emit the source map with an inline data URL
   --sources-content=false   Omit "sourcesContent" in generated source maps
-  --tree-shaking=...        Set to "ignore-annotations" to work with packages
-                            that have incorrect tree-shaking annotations
+  --tree-shaking=...        Force tree shaking on or off (false | true)
   --tsconfig=...            Use this tsconfig.json file instead of other ones
   --version                 Print the current version (` + esbuildVersion + `) and exit
 
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -2310,7 +2310,7 @@ func (cache *runtimeCache) parseRuntime(options *config.Options) (source logger.
 
 		// Always do tree shaking for the runtime because we never want to
 		// include unnecessary runtime code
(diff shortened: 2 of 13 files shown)


kzc:
> This previous setting has been moved to a separate setting because it actually impacts dead-code removal within expressions, which also applies when minifying with tree-shaking disabled.

Aren't the terms DCE and tree shaking synonymous?

privatenumber:
Seems @evanw is drawing a distinction between declaration vs expression-level DCE, where only the former is "tree-shaking".

> I have now looked into this and I was wrong: `--tree-shaking=ignore-annotations` isn't actually currently meaningless when tree shaking is disabled. It still affects how [`/* @__PURE__ */` comments](https://esbuild.github.io/api/#pure) are interpreted when `--minify-syntax` is enabled, which can trigger dead-code removal within expressions. This is related to tree shaking but seems somewhat separate since "tree shaking" usually refers to declaration-level dead code removal and you may want dead-code removal within expressions without declaration-level dead code removal. So I think it's best to separate those two concepts. My PR for this change is here: #1625.

- https://github.com/evanw/esbuild/pull/1617#issuecomment-925018103

PR #1630, opened 2021-09-23
fix cache condition in iswin_wasm

LeGEC:
unless I am mistaken, the boolean dance here does not match the intentions : in its current version, `checkedIfWindows` is never set to true to each individual call to `CheckIfWindows()` will trigger a call to `os.Stat()`


evanw:
Thanks for the fix!

PR #1660, opened 2021-10-05
feat: drop catch binding when optional catch binding is supported

sapphi-red:
This PR will slightly improve minimized result when optional catch binding is supported.

```js
try {
  throw 0
} catch (e) {}

// ----- into -----

try {
  throw 0
} catch {}
```
 
As far as I am aware, the pattern below has a side effect and cannot be dropped, but the code above will not have any. 
```js
try {
  throw {
    get a() {
      console.log(0)
      return 0;
    }
  }
} catch ({ a }) {} // prints 0
```

### DIFF
--- a/internal/js_parser/js_parser_test.go
+++ b/internal/js_parser/js_parser_test.go
@@ -4804,3 +4804,10 @@ func TestASCIIOnly(t *testing.T) {
 	expectPrintedTargetASCII(t, 5, "export var π", "export var \\u03C0;\n")
 	expectParseErrorTargetASCII(t, 5, "export var 𐀀", es5)
 }
+
+func TestDropUnusedCatchBinding(t *testing.T) {
+	expectPrintedMangle(t, "try { throw 0 } catch (e) { console.log(0) }", "try {\n  throw 0;\n} catch {\n  console.log(0);\n}\n")
+	expectPrintedMangle(t, "try { throw 0 } catch (e) { console.log(0, e) }", "try {\n  throw 0;\n} catch (e) {\n  console.log(0, e);\n}\n")
+	expectPrintedMangle(t, "try { thrower() } catch ({ a }) { console.log(0) }", "try {\n  thrower();\n} catch ({ a }) {\n  console.log(0);\n}\n")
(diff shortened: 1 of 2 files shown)


internal/js_parser/js_parser_test.go
```diff
@@ -4804,3 +4804,14 @@ func TestASCIIOnly(t *testing.T) {
...
+
+func TestDropUnusedCatchBinding(t *testing.T) {
+	expectPrintedMangle(t, "try { throw 0 } catch (e) { console.log(0) }", "try {\n  throw 0;\n} catch {\n  console.log(0);\n}\n")
+	expectPrintedMangle(t, "try { throw 0 } catch (e) { console.log(0, e) }", "try {\n  throw 0;\n} catch (e) {\n  console.log(0, e);\n}\n")
+	expectPrintedMangle(t, "try { thrower() } catch ({ a }) { console.log(0) }", "try {\n  thrower();\n} catch ({ a }) {\n  console.log(0);\n}\n")
+	expectPrintedMangleTarget(t, 2018, "try { throw 0 } catch (e) { console.log(0) }", "try {\n  throw 0;\n} catch (e) {\n  console.log(0);\n}\n")
+
+	expectPrintedMangle(t, "try {\n throw 1 } catch (x) { y.push(x); var x = 2; y.push(x) }", "try {\n  throw 1;\n} catch (x) {\n  y.push(x);\n  var x = 2;\n  y.push(x);\n}\n")
+	expectPrintedMangle(t, "try {\n throw 1 } catch (x) { var x = 2; y.push(x) }", "try {\n  throw 1;\n} catch (x) {\n  var x = 2;\n  y.push(x);\n}\n")
+	expectPrintedMangle(t, "try {\n throw 1 } catch (x) { var x = 2 }", "try {\n  throw 1;\n} catch (x) {\n  var x = 2;\n}\n")
```
sapphi-red, line 4816: This case can be minified like below.
But it will be complicated even though it's a very rare pattern.
So I did not implement it.

```js
try {
  throw 1;
} catch {}
```

PR #1678, opened 2021-10-12
Allow bundled esbuild with ESBUILD_BINARY_PATH

heypiotr:
In esbuild 0.12, there was this (unintended?) "feature" where you could bundle the esbuild JS wrapper, as long as you provided the ESBUILD_BINARY_PATH at runtime.

As of 0.13, when switching to optionalDependencies, [some code got shuffled](https://github.com/evanw/esbuild/commit/9e569c40f9f642139c59372e7321d90baeb507c1#diff-1453791842f22ef23bcd0b54684bc2ba6bdedc771b406796cd4d0de57e1d02f9L48), and this is no longer possible.

Now, I realize that this was never an official or intended behavior, but I think it's useful and it doesn't take a lot to bring it back, which is what this PR proposes.

Our specific use case is: we've got Node.js app which uses esbuild, which we run on an AWS Lambda. The easiest way for us to deploy it is to bundle the entire app into a single JS file. This way, we only need to upload two files (the JS bundle + the esbuild binary for the Lambda's architecture, which notably might be different from the architecture of the CI/build machine), set the ESBUILD_BINARY_PATH, and we're good to go.

Otherwise, we'll have to mark esbuild as external, then locate `node_modules/esbuild` in our Yarn-workspaces monorepo and upload it to the Lambda as well... Yet we'll still have to do the dance where we upload the binary for the Lambda's architecture, and use ESBUILD_BINARY_PATH to point to it. And if we're doing that anyway, then I think the esbuild JS wrapper could just as well be in a bundle?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 
     ⚠️ Note: NetBSD is not one of [Node's supported platforms](https://nodejs.org/api/process.html#process_process_platform), so installing esbuild may or may not work on NetBSD depending on how Node has been patched. This is not a problem with esbuild. ⚠️
 
+* Disable the "esbuild was bundled" warning if `ESBUILD_BINARY_PATH` is provided ([#1678](https://github.com/evanw/esbuild/pull/1678))
+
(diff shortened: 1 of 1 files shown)


heypiotr:
Thanks! ❤️

PR #1756, opened 2021-11-08
feat: minify `font`/`font-family`/`font-weight` CSS property

sapphi-red:
This PR will add the following minifications.

- `font-weight: bold` into `font-weight: 700`
- unquote `font-family` values if possible


CHANGELOG.md
```diff
@@ -37,6 +37,24 @@
...
+
+    ```css
+    /* Original code */
+    div {
+      font: bold 1rem / 1.2 "Segoe UI", sans-serif, "Segoe UI Emoji";
+    }
+
+    /* Output with "--minify" */
+    div{font:700 1rem/1.2 Segoe UI,sans-serif,"Segoe UI Emoji"}
+    ```
```
yisibl, line 52: Hi, @sapphi-red 

`font-family: "Segoe UI Emoji"`

What is the reason for not removing the quotes when compressing? As far as I know, it is safe to remove the quotes.

sapphi-red: Hi.

1. [`emoji` is a `<generic-family>`](https://www.w3.org/TR/css-fonts-4/#emoji-def).
1. if the font name includes `<generic-family>`, it must be quoted.

> Any identifier which could be misinterpreted as a pre-defined keyword in the font-family value definition, or the common-keywords|CSS-wide keyword values, is not allowed.
> Note: this means that if you really have a font whose name is the same as one of the [<generic-family>](https://www.w3.org/TR/css-fonts-4/#generic-family-value) names, or the common-keywords|CSS-wide keyword values, it must be quoted.
> https://www.w3.org/TR/css-fonts-4/#family-name-syntax

This is the reason.

yisibl: Thanks, I forgot that `emoji` is also `<generic-family>`.

PR #1758, opened 2021-11-08
feat: Shorten "top", "right"... properties into "inset" property

sapphi-red:
This PR will shorten `top`, `right` ... properties into `inset` property if it is supported.

```css
div {
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
}

/* ----- into ----- */

div {
  inset: 0;
}
```


evanw (approved):
Cool! Looks good to me. TIL about the `inset` property.

Issue #1803, opened 2021-11-25
panic after warning of missing semi-colon for css @import


sbfaulkner:
The following CSS causes a panic when esbuild tries to process it...

```
@import url("fonts.css")

.one {
	background: url("one.jpg");
}

.two {
	background: url("two.jpg");
}
```

### the panic…
```
$ bin/esbuild --loader=css --minify <style.css
 > style.css:9:1: warning: Expected ";"
    9 │ }
      ╵  ^

panic: runtime error: index out of range [1] with length 1

goroutine 10 [running]:
github.com/evanw/esbuild/internal/css_printer.(*printer).printTokens(0xc0000bf420, {0xc000222100, 0x4, 0x7301040000}, {0x10, 0x0})
	/Users/sbfaulkner/src/github.com/evanw/esbuild/internal/css_printer/css_printer.go:694 +0xcb9
github.com/evanw/esbuild/internal/css_printer.(*printer).printTokens(0xc0000bf420, {0xc000224000, 0x6, 0x0}, {0x0, 0x0})
	/Users/sbfaulkner/src/github.com/evanw/esbuild/internal/css_printer/css_printer.go:704 +0xa5e
github.com/evanw/esbuild/internal/css_printer.(*printer).printRule(0xc0000bf420, {{0x0}, {0x1512c88, 0xc000020840}}, 0x0, 0x0)
	/Users/sbfaulkner/src/github.com/evanw/esbuild/internal/css_printer/css_printer.go:104 +0x1058
github.com/evanw/esbuild/internal/css_printer.Print({{0xc00003c840, 0x1, 0x1}, {0xc00000e9a8, 0x1, 0x1}, {{0x0, 0x0}, {{0x0}, 0x0}}, ...}, ...)
	/Users/sbfaulkner/src/github.com/evanw/esbuild/internal/css_printer/css_printer.go:54 +0x305
github.com/evanw/esbuild/internal/bundler.(*linkerContext).generateChunkCSS(0xc00017b040, {0xc0000dd080, 0x1, 0x1}, 0x0, 0xc00002b850)
	/Users/sbfaulkner/src/github.com/evanw/esbuild/internal/bundler/linker.go:4882 +0xe38
created by github.com/evanw/esbuild/internal/bundler.(*linkerContext).generateChunksInParallel
	/Users/sbfaulkner/src/github.com/evanw/esbuild/internal/bundler/linker.go:356 +0x46f
```

[...]

PR #1808, opened 2021-11-27
recover from panic when generating chunks

sbfaulkner:
further to #1803 this PR recovers from any panics in the goroutines, converting them to errors (which can then be handled appropriately)

eg.

```
$ bin/esbuild --loader=css --minify <bad.css
 > <stdin>:9:1: warning: Expected ";"
    9 │ }
      ╵  ^

 > error: Internal error: runtime error: index out of range [1] with length 1

1 warning and 1 error
```


sbfaulkner:
thanks @evanw - the changes you made on top of this make it even better 🥂 

I'm not sure what your release cadence is, but I'd love to see a 0.14.1 when you get a chance ❤️

PR #1835, opened 2021-12-07
Allows watch mode even when build initially fails in javascript API

ncharalampidis:
The rebuild callback is ignored if the initial build fails, even if the error that causes the build to fail is being fixed afterwards.

```
require('esbuild').build({
  entryPoints: ['app.js'],
  watch: {
    onRebuild() {
      console.log('This is never called if app.js has an error initially')
    }
  }
})
```

This pull request fixes the issue by moving the error callback after the handling of the rebuildID and watchID and before the success callback.


evanw:
I'm closing this because the `onRebuild` callback has been removed, and the underlying problem now should be fixed. See #2816 for details.

PR #1864, opened 2021-12-17
Support s390x BE architecture

shahidhs-ibm:
This PR adds support for IBM Z 64-bit BE (s390x) Linux.


Makefile
```diff
@@ -29,7 +29,11 @@ check-go-version:
 ESBUILD_RACE ?= -race
 
 test-go:
+ifeq (s390x,$(shell uname -p))
+	go test ./internal/...
+else
```
evanw, line 34: Why is this necessary?

shahidhs-ibm: `-race` option is not supported on s390x arch. If we don't remove this flag for s390x, we get following error:
```
go test: -race is only supported on linux/amd64, linux/ppc64le, linux/arm64, freebsd/amd64, netbsd/amd64, darwin/amd64, darwin/arm64, and windows/amd64
```

evanw: Ok, good to know. I'll disable it by default in the future then.

lib/npm/node-platform.ts
```diff
@@ -24,6 +24,7 @@ export const knownUnixlikePackages: Record<string, string> = {
   'linux ia32 LE': 'esbuild-linux-32',
   'linux mips64el LE': 'esbuild-linux-mips64le',
   'linux ppc64 LE': 'esbuild-linux-ppc64le',
+  'linux s390x BE': 'esbuild-linux-s390x',
```
evanw, line 27: I just realized that this is our first big-endian platform! I think everything should work fine but we'll see. If something doesn't work about the published package or install script, we can fix it later.

evanw (approved):
This looks good. If all goes well, this should be supported starting with the next published version of esbuild whenever that goes out.

PR #1866, opened 2021-12-18
Added optional pixel values

wangdaoo:
Hi! 
Evan

I really admire your contribution, The emergence of esbuild has made the JavaScipt community even better.

Just now, I was looking at the source code and found that the CSS optional pixel values should also be `rem`, `vh`, `vw`, which are very common.

Hope you can see！

Thank you again for your contribution！

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@
 bin
 esbuild.exe
 node_modules/
+.vscode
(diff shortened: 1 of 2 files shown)


evanw:
It's true that they are very common, but it's not true that they are supported everywhere. That's why the comment above the code you changed references the compatibility table here: https://developer.mozilla.org/en-US/docs/Web/CSS/length. These units are not supported in IE8, which could still hypothetically be used by a user.

PR #1876, opened 2021-12-21
recover from unexpected brace at top-level

sbfaulkner:
When minifying, esbuild will currently truncate output when it encounters an unexpected `}` at the top-level of the css.

Given CSS like...

```
.red {
  color: red;
}
}
.blue {
  color: blue;
}
.green {
  color: green;
}
```

This produces...
```
bin/esbuild --loader=css --minify <rgb.css
▲ [WARNING] Expected end of file but found "}"

    <stdin>:4:0:
      4 │ }
        ╵ ^

1 warning
.red{color:red}
```

This PR attempts to recover from the situation (more like the best-effort recovery seen in other minification processors and/or css parsers), instead producing the following...
```
bin/esbuild --loader=css --minify <rgb.css
▲ [WARNING] Unexpected "}"

    <stdin>:4:0:
      4 │ }
        ╵ ^

1 warning
.red{color:red}.blue{color:#00f}.green{color:green}
```

### DIFF
--- a/internal/css_parser/css_parser.go
+++ b/internal/css_parser/css_parser.go
@@ -197,7 +197,15 @@ loop:
 		}
 
 		switch p.current().Kind {
-		case css_lexer.TEndOfFile, css_lexer.TCloseBrace:
+		case css_lexer.TEndOfFile:
+			break loop
+
+		case css_lexer.TCloseBrace:
+			if context.isTopLevel {
+				p.unexpected()
+				p.advance()
+				continue
+			}
 			break loop
 
 		case css_lexer.TWhitespace:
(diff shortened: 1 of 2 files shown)


sbfaulkner:
@evanw when you have time, is this the correct fix?

evanw:
Yeah, looks good to me. Thanks!

evanw:
Hmm actually when I run this in the browser, the result is different. Sorry for merging without checking first. In the browser, the `red` and `green` rules take effect but the `blue` rule doesn't. So esbuild should follow that behavior IMO. Still recover and continue but the rule starting with `}` is invalid. I'll change it to behave like that instead.

PR #1900, opened 2022-01-01
feat(wasm): use instantiateStreaming as much as possible

lbwa:
This PR aims to optimize `esbuild-wasm` loading process.

## Changes
  1. load wasm file in ~~main~~ worker thread
      - absolute path is required for resource fetching in worker thread. PR has covered scenario that wasmURL is a related path.
      - correct mime header is required. Otherwise, fallback to `instantiate` API.
  1. handle wasm loading errors and push them back to the main thread.
  1. ~~only use `WebAssembly.instantiate` to instantiate wasm code.~~ use `WebAssembly.instantiateStreaming` API by default. It will use `WebAssembly.instantiate` again if current browser doesn't support `instantiateStreaming`.

### DIFF
--- a/lib/npm/browser.ts
+++ b/lib/npm/browser.ts
@@ -104,7 +104,13 @@ const startRunningService = async (wasmURL: string, useWorker: boolean): Promise
     }
   }
 
-  worker.onmessage = ({ data }) => readFromStdout(data)
+  worker.onmessage = ({ data }) => {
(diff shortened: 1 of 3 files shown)


lbwa:
A reference from https://github.com/evanw/esbuild/pull/1036#issuecomment-806424401, or perhaps we have more available advice to optimize it? @evanw

PR #1974, opened 2022-01-31
Fix a mention of "EImportString" from "EImportCall"

lgarron:
Before, the comment on `LeadingInteriorComments` from "EImportCall" was referring to itself.


evanw (approved):
Whoops! Thanks.

Issue #1983, opened 2022-02-03
'browser' field in dependent package.json files is not respected


francisu:
If the `browser` field is specified in a `package.json` file of a dependency (not the top level), it's not respected.

This is different from the Vite behavior. (I'm using yarn 3 so I use the PR that addresses this issue: https://github.com/vitejs/vite/issues/1979). Vite out of the box does respect the `browser` field in dependencies for non yarn pnp environments.

Here is a repo that shows the problem: https://github.com/snapstrat/pino-browser

This issue appears to be related, but is not exactly the same: https://github.com/evanw/esbuild/issues/1324

One major library where this breaks is the `aws-sdk`. Resolving this would make esbuild work nicely in this case and help adoption.

Thanks so much for your hard work on esbuild! I'm so happy to be out from under create-react-app/webpack slowness and complexity.

evanw:
I think the problem is that you are using yarn, and `@yarnpkg/esbuild-plugin-pnp` doesn't support the `browser` field: https://github.com/yarnpkg/berry/issues/2987. It seems to work fine if you use npm instead of yarn:

```
npm install
npm run esbuild
```

This isn't something esbuild can fix itself because the `@yarnpkg/esbuild-plugin-pnp` plugin overrides path resolution (by design).

The easy way for you to fix this is to just use npm instead of yarn (or at least to disable Yarn PnP). The hard way would be to add support for the `browser` field to the `@yarnpkg/esbuild-plugin-pnp` plugin.

Closing as this is not a problem with esbuild itself.

francisu:
> Closing as this is not a problem with esbuild itself.

Thanks for the fast response!

I know that plugin well and am happy to fix it. I assume it would need to read the package.json file in the each directory in the loader (onLoad) to determine which actual files to get for the browser cases; this would make it have to be invoked for each module load, and I'm hoping that won't hurt performance too badly. I'm guessing esbuild handles the package.json interpretation directly only if it has direct access to the node_modules file system.

francisu:
BTW - not using yarn is not an option as I depend on some of the unique features of yarn 2+. So I will make this work for the yarn environment.

francisu:
[...]

PR #1985, opened 2022-02-04
add zip file support

evanw:
This PR enables esbuild to transparently traverse into `.zip` files as if they were directories. This is relevant to Yarn which stores packages in `.zip` files and then monkey-patches the JavaScript file system API so other JS code doesn't need to know about it. That doesn't work for esbuild because it's written in Go, so Yarn needs an esbuild plugin to do this which is single-threaded and also doesn't handle all of the cases esbuild handles. This change to add `.zip` file support to esbuild seems to work fine, and it's not a lot of code to maintain.

I do still need to figure out what the lifecycle of the zip file reader is though. Right now they aren't closed. I also need to check that this works on Windows, which is always needs special-casing when it comes to path separators. And ideally this would have some test coverage.

### DIFF
--- a/internal/fs/fs_real.go
+++ b/internal/fs/fs_real.go
@@ -110,12 +110,21 @@ func RealFS(options RealFSOptions) (FS, error) {
 		watchData = make(map[string]privateWatchData)
 	}
 
-	return &realFS{
+	var result FS = &realFS{
 		entries:           make(map[string]entriesOrErr),
 		fp:                fp,
 		watchData:         watchData,
 		doNotCacheEntries: options.DoNotCache,
-	}, nil
+	}
+
+	// Add a wrapper that lets us traverse into ".zip" files. This is what yarn
+	// uses as a package format when in yarn is in its "PnP" mode.
+	result = &zipFS{
+		inner:    result,
+		zipFiles: make(map[string]*zipFile),
+	}
+
+	return result, nil
 }
 
 func (fs *realFS) ReadDirectory(dir string) (entries DirEntries, canonicalError error, originalError error) {
--- a/internal/fs/fs_zip.go
+++ b/internal/fs/fs_zip.go
@@ -1,0 +1,292 @@
+package fs
+
+import (
+	"archive/zip"
+	"io/ioutil"
+	"strings"
+	"sync"
+	"syscall"
+)
+
+type zipFS struct {
+	inner FS
+
+	zipFilesMutex sync.Mutex
+	zipFiles      map[string]*zipFile
+}
+
+type zipFile struct {
+	reader *zip.ReadCloser
+	err    error
+
+	dirs  map[string]*compressedDir
+	files map[string]*compressedFile
+	wait  sync.WaitGroup
+}
+
+type compressedDir struct {
+	entries map[string]EntryKind
+	path    string
+
+	// Compatible entries are decoded lazily
+	mutex      sync.Mutex
+	dirEntries DirEntries
+}
+
+type compressedFile struct {
+	compressed *zip.File
+
+	// The file is decompressed lazily
+	mutex    sync.Mutex
+	contents string
+	err      error
+	wasRead  bool
+}
+
+func (fs *zipFS) checkForZip(path string, kind EntryKind) (*zipFile, string) {
+	var zipPath string
+	var pathTail string
+
+	// Do a quick check for a ".zip" in the path at all
+	path = strings.ReplaceAll(path, "\\", "/")
+	if i := strings.Index(path, ".zip/"); i != -1 {
+		zipPath = path[:i+len(".zip")]
+		pathTail = path[i+len(".zip/"):]
+	} else if kind == DirEntry && strings.HasSuffix(path, ".zip") {
+		zipPath = path
+	} else {
+		return nil, ""
+	}
+
+	// If there is one, then check whether it's a file on the file system or not
+	fs.zipFilesMutex.Lock()
+	archive := fs.zipFiles[zipPath]
+	if archive != nil {
+		fs.zipFilesMutex.Unlock()
+		archive.wait.Wait()
+	} else {
+		archive = &zipFile{}
+		archive.wait.Add(1)
+		fs.zipFiles[zipPath] = archive
+		fs.zipFilesMutex.Unlock()
+		defer archive.wait.Done()
+
+		// Try reading the zip archive if it's not in the cache
+		tryToReadZipArchive(zipPath, archive)
+	}
+
+	if archive.err != nil {
+		return nil, ""
+	}
+	return archive, pathTail
+}
+
+func tryToReadZipArchive(zipPath string, archive *zipFile) {
+	reader, err := zip.OpenReader(zipPath)
+	if err != nil {
+		archive.err = err
+		return
+	}
+
+	dirs := make(map[string]*compressedDir)
+	files := make(map[string]*compressedFile)
+
+	// Build an index of all files in the archive
+	for _, file := range reader.File {
+		baseName := file.Name
+		if strings.HasSuffix(baseName, "/") {
+			baseName = baseName[:len(baseName)-1]
+		}
+		dirPath := ""
+		if slash := strings.LastIndexByte(baseName, '/'); slash != -1 {
+			dirPath = baseName[:slash]
+			baseName = baseName[slash+1:]
+		}
+		if file.FileInfo().IsDir() {
+			// Handle a directory
+			lowerDir := strings.ToLower(dirPath)
+			if _, ok := dirs[lowerDir]; !ok {
+				dirs[lowerDir] = &compressedDir{
+					path:    dirPath,
+					entries: make(map[string]EntryKind),
+				}
+			}
+		} else {
+			// Handle a file
+			files[strings.ToLower(file.Name)] = &compressedFile{compressed: file}
+			lowerDir := strings.ToLower(dirPath)
+			dir, ok := dirs[lowerDir]
+			if !ok {
+				dir = &compressedDir{
+					path:    dirPath,
+					entries: make(map[string]EntryKind),
+				}
+				dirs[lowerDir] = dir
+			}
+			dir.entries[baseName] = FileEntry
+		}
+	}
+
+	// Populate child directories
+	seeds := make([]string, 0, len(dirs))
+	for dir := range dirs {
+		seeds = append(seeds, dir)
+	}
+	for _, baseName := range seeds {
+		for baseName != "" {
+			dirPath := ""
+			if slash := strings.LastIndexByte(baseName, '/'); slash != -1 {
+				dirPath = baseName[:slash]
+				baseName = baseName[slash+1:]
+			}
+			lowerDir := strings.ToLower(dirPath)
+			dir, ok := dirs[lowerDir]
+			if !ok {
+				dir = &compressedDir{
+					path:    dirPath,
+					entries: make(map[string]EntryKind),
+				}
+				dirs[lowerDir] = dir
+			}
+			dir.entries[baseName] = DirEntry
+			baseName = dirPath
+		}
+	}
+
+	archive.dirs = dirs
+	archive.files = files
+	archive.reader = reader
+}
+
+func (fs *zipFS) ReadDirectory(path string) (entries DirEntries, canonicalError error, originalError error) {
+	entries, canonicalError, originalError = fs.inner.ReadDirectory(path)
+	if canonicalError != syscall.ENOENT && canonicalError != syscall.ENOTDIR {
+		return
+	}
+
+	// If the directory doesn't exist, try reading from an enclosing zip archive
+	zip, pathTail := fs.checkForZip(path, DirEntry)
+	if zip == nil {
+		return
+	}
+
+	// Does the zip archive have this directory?
+	dir, ok := zip.dirs[strings.ToLower(pathTail)]
+	if !ok {
+		return DirEntries{}, syscall.ENOENT, syscall.ENOENT
+	}
+
+	// Check whether it has already been converted
+	dir.mutex.Lock()
+	defer dir.mutex.Unlock()
+	if dir.dirEntries.data != nil {
+		return dir.dirEntries, nil, nil
+	}
+
+	// Otherwise, fill in the entries
+	dir.dirEntries = DirEntries{dir: path, data: make(map[string]*Entry, len(dir.entries))}
+	for name, kind := range dir.entries {
+		dir.dirEntries.data[strings.ToLower(name)] = &Entry{
+			dir:  path,
+			base: name,
+			kind: kind,
+		}
+	}
+
+	return dir.dirEntries, nil, nil
+}
+
+func (fs *zipFS) ReadFile(path string) (contents string, canonicalError error, originalError error) {
+	contents, canonicalError, originalError = fs.inner.ReadFile(path)
+	if canonicalError != syscall.ENOENT {
+		return
+	}
+
+	// If the file doesn't exist, try reading from an enclosing zip archive
+	zip, pathTail := fs.checkForZip(path, FileEntry)
+	if zip == nil {
+		return
+	}
+
+	// Does the zip archive have this file?
+	file, ok := zip.files[strings.ToLower(pathTail)]
+	if !ok {
+		return "", syscall.ENOENT, syscall.ENOENT
+	}
+
+	// Check whether it has already been read
+	file.mutex.Lock()
+	defer file.mutex.Unlock()
+	if file.wasRead {
+		return file.contents, file.err, file.err
+	}
+	file.wasRead = true
+
+	// If not, try to open it
+	reader, err := file.compressed.Open()
+	if err != nil {
+		file.err = err
+		return "", err, err
+	}
+	defer reader.Close()
+
+	// Then try to read it
+	bytes, err := ioutil.ReadAll(reader)
+	if err != nil {
+		file.err = err
+		return "", err, err
+	}
+
+	file.contents = string(bytes)
+	return file.contents, nil, nil
+}
+
+func (fs *zipFS) OpenFile(path string) (result OpenedFile, canonicalError error, originalError error) {
+	result, canonicalError, originalError = fs.inner.OpenFile(path)
+	return
+}
+
+func (fs *zipFS) ModKey(path string) (modKey ModKey, err error) {
+	modKey, err = fs.inner.ModKey(path)
+	return
+}
+
+func (fs *zipFS) IsAbs(path string) bool {
+	return fs.inner.IsAbs(path)
+}
+
+func (fs *zipFS) Abs(path string) (string, bool) {
+	return fs.inner.Abs(path)
+}
+
+func (fs *zipFS) Dir(path string) string {
+	return fs.inner.Dir(path)
+}
+
+func (fs *zipFS) Base(path string) string {
+	return fs.inner.Base(path)
+}
+
+func (fs *zipFS) Ext(path string) string {
+	return fs.inner.Ext(path)
+}
+
+func (fs *zipFS) Join(parts ...string) string {
+	return fs.inner.Join(parts...)
+}
+
+func (fs *zipFS) Cwd() string {
+	return fs.inner.Cwd()
+}
+
+func (fs *zipFS) Rel(base string, target string) (string, bool) {
+	return fs.inner.Rel(base, target)
+}
+
+func (fs *zipFS) kind(dir string, base string) (symlink string, kind EntryKind) {
+	return fs.inner.kind(dir, base)
+}
+
+func (fs *zipFS) WatchData() WatchData {
+	return fs.inner.WatchData()
+}
--- a/internal/fs/fs_zip.go
+++ b/internal/fs/fs_zip.go
@@ -107,10 +107,14 @@ func tryToReadZipArchive(zipPath string, archive *zipFile) {
 			// Handle a directory
 			lowerDir := strings.ToLower(dirPath)
 			if _, ok := dirs[lowerDir]; !ok {
-				dirs[lowerDir] = &compressedDir{
+				dir := &compressedDir{
 					path:    dirPath,
 					entries: make(map[string]EntryKind),
 				}
+
+				// List the same directory both with and without the slash
+				dirs[lowerDir] = dir
+				dirs[lowerDir+"/"] = dir
 			}
 		} else {
 			// Handle a file
@@ -122,7 +126,10 @@ func tryToReadZipArchive(zipPath string, archive *zipFile) {
 					path:    dirPath,
 					entries: make(map[string]EntryKind),
 				}
+
+				// List the same directory both with and without the slash
 				dirs[lowerDir] = dir
+				dirs[lowerDir+"/"] = dir
 			}
 			dir.entries[baseName] = FileEntry
 		}
@@ -147,7 +154,10 @@ func tryToReadZipArchive(zipPath string, archive *zipFile) {
 					path:    dirPath,
 					entries: make(map[string]EntryKind),
 				}
+
+				// List the same directory both with and without the slash
 				dirs[lowerDir] = dir
+				dirs[lowerDir+"/"] = dir
 			}
 			dir.entries[baseName] = DirEntry
 			baseName = dirPath
--- a/internal/fs/fs_zip.go
+++ b/internal/fs/fs_zip.go
@@ -3,6 +3,7 @@ package fs
 import (
 	"archive/zip"
 	"io/ioutil"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -159,6 +160,8 @@ func tryToReadZipArchive(zipPath string, archive *zipFile) {
 }
 
 func (fs *zipFS) ReadDirectory(path string) (entries DirEntries, canonicalError error, originalError error) {
+	path = mangleYarnPnPVirtualPath(path)
+
 	entries, canonicalError, originalError = fs.inner.ReadDirectory(path)
 	if canonicalError != syscall.ENOENT && canonicalError != syscall.ENOTDIR {
 		return
@@ -197,6 +200,8 @@ func (fs *zipFS) ReadDirectory(path string) (entries DirEntries, canonicalError 
 }
 
 func (fs *zipFS) ReadFile(path string) (contents string, canonicalError error, originalError error) {
+	path = mangleYarnPnPVirtualPath(path)
+
 	contents, canonicalError, originalError = fs.inner.ReadFile(path)
 	if canonicalError != syscall.ENOENT {
 		return
@@ -242,11 +247,15 @@ func (fs *zipFS) ReadFile(path string) (contents string, canonicalError error, o
 }
 
 func (fs *zipFS) OpenFile(path string) (result OpenedFile, canonicalError error, originalError error) {
+	path = mangleYarnPnPVirtualPath(path)
+
 	result, canonicalError, originalError = fs.inner.OpenFile(path)
 	return
 }
 
 func (fs *zipFS) ModKey(path string) (modKey ModKey, err error) {
+	path = mangleYarnPnPVirtualPath(path)
+
 	modKey, err = fs.inner.ModKey(path)
 	return
 }
@@ -260,6 +269,9 @@ func (fs *zipFS) Abs(path string) (string, bool) {
 }
 
 func (fs *zipFS) Dir(path string) string {
+	if prefix, suffix, ok := parseYarnPnPVirtualPath(path); ok && suffix == "" {
+		return prefix
+	}
 	return fs.inner.Dir(path)
 }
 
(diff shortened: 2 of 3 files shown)


arcanis:
I don't know if you'd find that too specific to be in the core, but Yarn also makes a little change on the filesystem paths: if you have a path like `/path/to/project/__virtual__/<segment>/<n>/path/to/file.js`, we take `/path/to/project`, apply `N` times the `..` operation (usually 0), and append `path/to/file.js` (the segment is completely ignored). So for instance:

```
/project/.yarn/__virtual__/something/0/cache/lodash.zip
  -> /project/.yarn/cache/lodash.zip

/project/subfolder/__virtual__/something/1/hello/world.js
  -> /project/hello/world.js
```

We do this because, to allow that a single package is instantiated multiple times (required when a package has peer dependencies that are provided by multiple different ancestors), they must have different paths on the disk (otherwise Node will cache them). Symlinks don't work, because Node runs realpath on them, hence those virtual paths, where in practice we replace `something` by the hash of the package ancestor, thus providing a different path for each peer dependency set.

evanw:
> I don't know if you'd find that too specific to be in the core, but Yarn also makes a little change on the filesystem paths: if you have a path like `/path/to/project/__virtual__/<segment>/<n>/path/to/file.js`, we take `/path/to/project`, apply `N` times the `..` operation (usually 0), and append `path/to/file.js` (the segment is completely ignored).

Thanks for the extra context. I don't fully understand what you're saying because I don't know enough about how Yarn works. But I'm guessing you're saying this:

* The `onResolve` callback for `@yarnpkg/esbuild-plugin-pnp` might map an import of `lodash/add.js` into `/project/.yarn/__virtual__/something/0/cache/lodash.zip/add.js`

* The `onLoad` callback for `@yarnpkg/esbuild-plugin-pnp` then takes in `/project/.yarn/__virtual__/something/0/cache/lodash.zip/add.js` and turns it into `/project/.yarn/cache/lodash.zip/add.js` because the actual zip file is `/project/.yarn/cache/lodash.zip` not `/project/.yarn/__virtual__/something/0/cache/lodash.zip`

My idea was for esbuild to fully replace `onLoad` since that should be faster than using a plugin. It means files can be loaded in parallel because everything stays in Go. If my understanding above is correct then I think esbuild would have to interpret `__virtual__` as well when it replaces `onLoad`, otherwise it wouldn't find the `.zip` files.

If Yarn did this `__virtual__` substitution in `onResolve` instead of esbuild doing it, then that would break semantics because esbuild (like node) also needs modules to have separate paths for their identities to be separate. Otherwise those modules would be instantiated once instead of being instantiated multiple times, which might be desirable from a bundle size perspective but is incorrect from a bundling perspective because it could result in behavior changes to the code.

And if Yarn did this `__virtual__` substitution in its own `onLoad`, that would defeat the goal of having esbuild handle `onLoad` for performance reasons.

> ```
> /project/subfolder/__virtual__/something/1/hello/world.js
> ```

When does this form without `.yarn` come up? It seems the most "clean" to only do this path manipulation when `.yarn` is a parent directory.

evanw:
Another idea: Once esbuild understands paths inside `.zip` files, you could potentially use the new-ish [`resolve` API](https://esbuild.github.io/plugins/#resolve) to reuse esbuild's internal logic for path remapping using `package.json` and other sources of information. The `@yarnpkg/esbuild-plugin-pnp` plugin could just map the package prefix to a prefix inside a `.zip` file in `onResolve` and then call `resolve` for esbuild to take it from there. I think that should fix things like #1983 without needing Yarn's esbuild plugin to copy all of esbuild's behaviors.

So specifically an import path of `lodash/add` could be mapped into `project/.yarn/__virtual__/something/0/cache/lodash.zip/add` and then passed to esbuild's `resolve` function which would then figure out that it needs to be `project/.yarn/__virtual__/something/0/cache/lodash.zip/add.js`. The drawback of this would be that it would be slower because it involves more IPC traffic. But the benefit is that the plugin would automatically match all of esbuild's various behaviors. Obviously the optimal solution would just be for esbuild to do it all but it sounds like stabilizing the PnP data format is still a while away.

Not sure if this is a good idea or not because of the performance impact, but I'm putting it out there anyway. Actually I could potentially add an API to say "call resolve on the return value for me" for `onResolve` to eliminate the extra IPC overhead. Hmm.

arcanis:
> But I'm guessing you're saying this:
> 
> * The `onResolve` callback for `@yarnpkg/esbuild-plugin-pnp` might map an import of `lodash/add.js` into `/project/.yarn/__virtual__/something/0/cache/lodash.zip/add.js`
> * The `onLoad` callback for `@yarnpkg/esbuild-plugin-pnp` then takes in `/project/.yarn/__virtual__/something/0/cache/lodash.zip/add.js` and turns it into `/project/.yarn/cache/lodash.zip/add.js` because the actual zip file is `/project/.yarn/cache/lodash.zip` not `/project/.yarn/__virtual__/something/0/cache/lodash.zip`

Yep, exactly! We do this work in the `onLoad` so that Node/Esbuild always see the files as different entities.

> When does this form without .yarn come up? It seems the most "clean" to only do this path manipulation when .yarn is a parent directory.

In practice probably never, but we have a [`virtualFolder`](https://yarnpkg.com/configuration/yarnrc#virtualFolder) setting that lets users tweak the location. I imagine it's seldom used, but it's difficult to say...

> Another idea: Once esbuild understands paths inside `.zip` files, you could potentially use the new-ish [`resolve` API](https://esbuild.github.io/plugins/#resolve) to reuse esbuild's internal logic for path remapping using `package.json` and other sources of information.

I agree it'd be really nice! The PnP API even has a function ([`resolveToUnqualified`](https://yarnpkg.com/advanced/pnpapi#resolvetounqualified)) that does just that, which is what we try to use when possible (for example that's what Webpack's enhanced-resolve use).

Note however that some context has to be preserved for the resolution to work, in particular for the [`exports` field](https://nodejs.org/api/packages.html#subpath-exports) to be taken into account (`exports` only apply to bare identifier resolutions, not to relative/absolute paths; so if the PnP API was calling Esbuild's `resolve` API with an absolute path, then we'd need a way to explain that originally it was a package path, and thus that the `exports` field should be taken into account).

CarlosBellange:
Yo, I wonder, when I uploaded my content, how fats you where to distribute and move on. 

Be fair

bryanjtc:
Any progress on this?

evanw:
This is sort of blocked waiting on Yarn to expose the necessary data, which from what I understand is blocked on releasing another major version of Yarn. See https://github.com/yarnpkg/berry/issues/3591 specifically:

> Write a formal spec for the PnP data files, so that 3rd-party resolvers (in particular native resolvers) can implement their own

arcanis:
@evanw I've started writing a formal spec at https://github.com/yarnpkg/berry/pull/4671 - do you mind reviewing it whenever you have some time?

evanw:
Thanks for putting that together! Took a look. Some thoughts (keep in mind while reading that I don't use Yarn myself):

* The specification doesn't mention `__virtual__`. Does that mean that it can be implemented correctly without needing to do anything related to `__virtual__`? If not, should the specification mention `__virtual__` somewhere?

* It sounds like `pnpEnableInlining` defaults to off which means this stuff won't work by default. Would it be ok to try to extract this information from `.pnp.js`/`.pnp.cjs` in that case? I'm pretty tempted because it would make a lot of scenarios just start working without any configuration changes (assuming the data is in the same format as the one in the specification, which is what it looks like). I tried extracting the data and it was pretty easy. I'm fine using the data in the JavaScript file as long as the format is documented so I know what it means, and as long as it's unlikely to undergo a breaking change in the future. I'm guessing that the specification is just stabilizing Yarn's internal format, which means it's now unlikely to change?

* I'm guessing that this unusual tuple-style representation of maps is intended to make it easy to pass it to the `new Map` constructor from ES6. Things made more sense when I realized this. Before I realized this, it was pretty confusing to see lots of nested arrays everywhere. For example, I was wondering how many elements could be in each array and what they meant. TypeScript type declarations might help maybe? The examples are great too though, so thanks for those.

* Is the iteration order over these maps important? If so, are there examples of cases where this matters?

* What is the format of `fallbackExclusionList`? I have a guess, but it's probably good to put it in the specification. My guess is that it's unpacked via something like this:

    ```js
    for (const [ident, references] of manifest.fallbackExclusionList)
      for (const reference of references)
        fallbackExclusionList.push({ ident, reference })
    ```

    To be clear, I'm not saying the format should be changed. Just documented.

* > For each _registryPkg_ entry in _manifest.packageRegistryData_

    FWIW it took me a while to figure out what this means. After reading the whole thing a few times, I now think this means something like this:

    ```js
    for (const [packageIdent, x] of manifest.packageRegistryData) {
      for (const [packageReference, y] of x) {
        const registryPkg = { packageIdent, packageReference, ...y }
        ...
      }
    }
    ```

* > For each _fallbackLocator_ in _manifest.packageRegistryData_

    I think you mean in _manifest.fallbackPool_?

* > Let _reference_ be the entry from _parentPkg.packageDependencies_ referenced by _ident_
    > If _dependency_ is null, then

    I think you mean if _reference_ is null?

* The function **GET_PACKAGE** is never defined. After reading everything, I'm guessing it goes something like this:

    ```js
    function GET_PACKAGE(manifest, { ident, reference }) {
      for (const [i, [r, data]] of manifest.packageRegistryData)
        if (i === ident && r === reference)
          return data
      return null
    }
    ```

    Obviously it's more efficient in reality since I assume these would be map lookups.

* > If _url_ starts with _registryPkg_, then

    Starts with what? _registryPkg.ident_? _registryPkg.reference_? _registryPkg.packageLocation_? Something else? I'm guessing it's _registryPkg.ident_, but I'm flagging this because I found this confusing.

    It's also not clear what exactly "starts with means". For example, "react-dom" starts with "react". Is that supposed to match? Or is it supposed to be some sort of path-based "starts with" instead of string-based "starts with"?

* What is supposed to happen if these **GET_PACKAGE** calls fail to find the package?

    > Let _parentPkg_ be **GET_PACKAGE**(_manifest_, _parentLocator_)
    > Let _fallbackPkg_ be **GET_PACKAGE**(_manifest_, _fallbackLocator_)
    > Let _dependencyPkg_ be **GET_PACKAGE**(_manifest_, {_ident_, _reference_})

* What's supposed to happen if **RESOLVE_VIA_FALLBACK** returns null here?

    > Set _resolved_ to **RESOLVE_VIA_FALLBACK**(_manifest_, _specifier_) and return it

    Generally auditing everything to make sure failure paths are covered could be good. I might have missed some.

* > Let _ident_ be the package scope and name from _specifier_

    I think I understand what this is saying, but you might want to be more explicit here and/or show some examples. I'm guessing this goes something like how it goes in node (copied below):

    > 4. If _packageSpecifier_ does not start with _"@"_, then
    >    1. Set _packageName_ to the substring of _packageSpecifier_ until the first
    >       _"/"_ separator or the end of the string.
    > 5. Otherwise,
    >    1. If _packageSpecifier_ does not contain a _"/"_ separator, then
    >       1. Throw an _Invalid Module Specifier_ error.
    >    2. Set _packageName_ to the substring of _packageSpecifier_
    >       until the second _"/"_ separator or the end of the string.
    > 6. If _packageName_ starts with _"."_ or contains _"\\"_ or _"%"_, then
    >    1. Throw an _Invalid Module Specifier_ error.

    It's ok if it's different in Yarn. But would be good to know if there are any characters that might cause the package scope and name to be considered invalid, for example.

* This isn't strictly necessary, but it would make my life easier if there was either a big reference project that uses all of these features or some sort of easy-to-use test suite for this to help ensure that my implementation is correct.

I also tried implementing this specification on a branch here if it's interesting to you: https://github.com/evanw/esbuild/commits/yarn-pnp.

merceyz:
That's some great feedback @evanw but it would probably be better to make it in https://github.com/yarnpkg/berry/pull/4671 so readers of that PR can see it.

evanw:
My intent of posting a comment here was to provide one-off feedback instead of being drawn into the spec development process. You are welcome to do what you like with my feedback.

merceyz:
Ah, that's totally fine, feel free to hide these three comments 👍

arcanis:
Thanks for the review! I've updated the document and corrected a few other places. A couple of important mistakes I fixed:

- About `fallbackPool` - it's actually not the list of locators whose dependencies should act as fallbacks, but rather the fallback dependencies themselves (see the updated [`RESOLVE_VIA_FALLBACK`](https://deploy-preview-4671--yarn2.netlify.app/advanced/pnp-spec#resolve_via_fallbackmanifest-specifier)). The dependencies of the top-level package (ie the one referenced by `null, null` in `packageRegistryData`) are however implicitly added to `fallbackPool`, overriding them¹

- The dependencies listed in `packageDependencies` (and `fallbackPool`) are a map where the key is an ident, and the string is either a string (a reference), or a new `[ident, reference]` tuple. This structure only happens when a package is "aliased" to another (for example you write in your lockfile that `foo` resolves to `bar@1.0.0` - in this case, the mapping in the manifest would be `["foo", ["bar", "npm:1.0.0"]]`). Doesn't happen often, and probably could be deprecated, but I haven't investigated enough yet so I still preferred to write it into the spec.

- In your case (a tool part of a project) the [`FIND_PNP_MANIFEST`](https://deploy-preview-4671--yarn2.netlify.app/advanced/pnp-spec#find_pnp_manifesturl) function should probably only be called once for the whole process. I clarified the function definition to make that more obvious.

Regarding your detailed feedbacks:

> The specification doesn't mention __virtual__

I've added a [section](https://deploy-preview-4671--yarn2.netlify.app/advanced/pnp-spec#package-locations) about this.

> Would it be ok to try to extract this information from .pnp.js/.pnp.cjs in that case?

I think so, although we perhaps should add markers around the data string, to make sure it doesn't change? 🤔 

> I was wondering how many elements could be in each array and what they meant. TypeScript type declarations might help maybe?

As a workaround we have some types [here](https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-pnp/sources/types.ts#L38-L47)

> Is the iteration order over these maps important? If so, are there examples of cases where this matters?

I don't think so - we shouldn't have duplicate keys in those data, so order shouldn't matter. The only place where order matters is inside [`FIND_LOCATOR`](https://deploy-preview-4671--yarn2.netlify.app/advanced/pnp-spec#find_locatormanifest-moduleurl), hence the length check. While there can be two entries with the same `packageLocation`, they should have the exact same `packageDependencies` so it shouldn't have impact.

> Starts with what? registryPkg.ident? registryPkg.reference? registryPkg.packageLocation? Something else? I'm guessing it's registryPkg.ident, but I'm flagging this because I found this confusing.

Starts with `packageLocation` - basically we're looking for the deepest package location that contains the given module url. Note that `packageLocation` *always* ends with a trailing `/`, you wouldn't have `./react` but `./react/` (and thus it wouldn't match `./react-dom/`).

> What is supposed to happen if these GET_PACKAGE calls fail to find the package?

Clarified that the function cannot fail, because all packages referenced in the manifest are guaranteed to have a matching entry inside the `packageRegistryData` map.

> What's supposed to happen if RESOLVE_VIA_FALLBACK returns null here?

Clarified that it returns `undefined`, which bubbles into `RESOLVE_TO_UNQUALIFIED`, which throws an exception

> This isn't strictly necessary, but it would make my life easier if there was either a big reference project that uses all of these features or some sort of easy-to-use test suite for this to help ensure that my implementation is correct.

I've started writing tests [here](https://github.com/yarnpkg/berry/blob/mael/pnp-spec/packages/yarnpkg-pnp/tests/testExpectations.json). I wrote them in JSON so it's easier for you to integrate into your own test runner; in our case, the code that runs this JSON is here: [pnpStandardRunner.test.ts](https://github.com/yarnpkg/berry/blob/mael/pnp-spec/packages/yarnpkg-pnp/tests/pnpStandardRunner.test.ts).

---

¹ This behaviour was how the fallback worked, before `fallbackPool` even existed, but we probably could deprecate it to only use the fallback pool ... still, that's how it currently works, so I wrote it as such in the spec.

arcanis:
I just deployed the documentation to the website: https://yarnpkg.com/advanced/pnp-spec/

evanw:
Thank you for publishing the specification and the tests. I have updated my branch here, including incorporating the tests: https://github.com/evanw/esbuild/tree/yarn-pnp. Unfortunately many of the tests fail with the algorithm in the specification. I was able to get all tests to pass by deviating from the specification and following Yarn's implementation instead. I have the following additional feedback regarding the specification (sorry for another long post):

1. I believe `RESOLVE_VIA_FALLBACK(manifest, specifier)` should be `RESOLVE_VIA_FALLBACK(manifest, ident)` since the function body uses `ident` instead of `specifier`

2. The `dependencyTreeRoots` value is defined but never used. I'm assuming this means it can be ignored because it's irrelevant for implementing Yarn PnP?

3. I think the data format for `fallbackPool` is incorrect. The specification has this example data:

    ```js
    fallbackPool: [{
      name: "@app/monorepo",
      reference: "workspace:.",
    }],
    ```

    But the tests look like this:

    ```json
    "fallbackPool": [
      ["test-2", "npm:1.0.0"],
      ["alias", ["test-1", "npm:1.0.0"]]
    ],
    ```

    The type definitions in the implementation also use triple-nested arrays instead of arrays of objects. I changed esbuild to follow the data format used by the implementation to fix the failing fallback-related tests.

4. I suspect that the definition of `FIND_LOCATOR` is incorrect. Here is the definition:

    > ### FIND_LOCATOR(`manifest`, `moduleUrl`)
    >
    > Note: The algorithm described here is quite inefficient. You should make sure to prepare data structure more suited for this task when you read the manifest.
    > 1. Let `bestLength` be **0**
    > 2. Let `bestLocator` be **null**
    > 3. Let `relativeUrl` be the relative path between `manifest` and `moduleUrl`
    >     1. Note: Make sure it always starts with a `./` or `../`
    > 4. If `relativeUrl` matches `manifest.ignorePatternData`, then
    >     1. Return **null**
    > 5. For each `referenceMap` value in `manifest.packageRegistryData`
    >     1. For each `registryPkg` value in `referenceMap`
    >         1. If `registryPkg.discardFromLookup` **isn't true**, then
    >             1. If `registryPkg.packageLocation.length` is greater than `bestLength`, then
    >                 1. If `relativeUrl` starts with `registryPkg.packageLocation`, then
    >                     1. Set `bestLength` to `registryPkg.packageLocation.length`
    >                     2. Set `bestLocator` to the current `registryPkg` locator
    > 6. Return `bestLocator`

    However, here is the first manifest in the set of test cases:

    ```json
    {
      "__info": [],
      "dependencyTreeRoots": [{
        "name": "root",
        "reference": "workspace:."
      }],
      "ignorePatternData": null,
      "enableTopLevelFallback": false,
      "fallbackPool": [],
      "fallbackExclusionList": [],
      "packageRegistryData": [
        [null, [
          [null, {
            "packageLocation": "./",
            "packageDependencies": [],
            "linkType": "SOFT"
          }]
        ]],
        ["root", [
          ["workspace:.", {
            "packageLocation": "./",
            "packageDependencies": [["test", "npm:1.0.0"]],
            "linkType": "SOFT"
          }]
        ]],
        ...
      ]
    }
    ```

    Doing a lookup for `./` in this map, which is what the first test does, gives a locator of `[null, null]` via the algorithm in the specification but `["root", "workspace:."]` via real Yarn PnP. Thus the first test incorrectly fails because `test` is not found as a dependency.

    I tried digging around for the real implementation of `FIND_LOCATOR` and I'm assuming it's this:

    ```ts
    function findPackageLocator(location: PortablePath, {resolveIgnored = false, includeDiscardFromLookup = false}: {resolveIgnored?: boolean, includeDiscardFromLookup?: boolean} = {}): PhysicalPackageLocator | null {
      if (isPathIgnored(location) && !resolveIgnored) return null;
      let relativeLocation = ppath.relative(runtimeState.basePath, location);
      if (!relativeLocation.match(isStrictRegExp)) relativeLocation = `./${relativeLocation}` as PortablePath;
      if (!relativeLocation.endsWith(`/`)) relativeLocation = `${relativeLocation}/` as PortablePath;
      do {
        const entry = packageLocatorsByLocations.get(relativeLocation);
        if (typeof entry === `undefined` || (entry.discardFromLookup && !includeDiscardFromLookup)) {
          relativeLocation = relativeLocation.substring(0, relativeLocation.lastIndexOf(`/`, relativeLocation.length - 2) + 1) as PortablePath;
          continue;
        }
        return entry.locator;
      } while (relativeLocation !== ``);
      return null;
    }
    ```

    Specifically it uses the `packageLocatorsByLocations` map, which is populated like this:

    ```ts
    const packageLocatorsByLocations = new Map<PortablePath, {locator: PhysicalPackageLocator, discardFromLookup: boolean}>();
    data.packageRegistryData.forEach(([packageName, packageStoreData]) => packageStoreData.forEach(([packageReference, packageInformationData]) => {
      if ((packageName === null) !== (packageReference === null)) throw new Error(`Assertion failed: The name and reference should be null, or neither should`);
      const discardFromLookup = packageInformationData.discardFromLookup ?? false;
      const packageLocator: PhysicalPackageLocator = {name: packageName, reference: packageReference};
      const entry = packageLocatorsByLocations.get(packageInformationData.packageLocation);
      if (!entry) {
        packageLocatorsByLocations.set(packageInformationData.packageLocation, {locator: packageLocator, discardFromLookup});
      } else {
        entry.discardFromLookup = entry.discardFromLookup && discardFromLookup;
        if (!discardFromLookup) {
          entry.locator = packageLocator;
        }
      }
    }));
    ```

    This seems to indicate that a) iteration order matters after all for duplicate `packageLocation` values and b) that the "If `registryPkg.packageLocation.length` is greater than `bestLength`" condition isn't sufficient to capture what Yarn PnP is doing.

    Hopefully I'm correct that this is a deficiency of the specification and not a transcription error of mine. If it's a problem with the specification, then I leave it up to you how you want to resolve it. One suggestion could be to just specify the efficient algorithm Yarn actually uses instead of specifying the inefficient algorithm that Yarn doesn't use. I have made esbuild follow Yarn's implementation here (specifically `packageLocatorsByLocations`) and that fixed the test failures.

5. I think the part regarding `ignorePatternData` is also incorrect. Here's the specification:

    > 3. Let `relativeUrl` be the relative path between `manifest` and `moduleUrl`
    >
    >     1. Note: Make sure it always starts with a `./` or `../`
    >
    > 4. If `relativeUrl` matches `manifest.ignorePatternData`, then
    >
    >     1. Return **null**

    This causes the test `shouldn't go through PnP when trying to resolve dependencies from packages covered by ignorePatternData` to fail. The problem is that the specification matches with `ignorePatternData` after inserting the `./` prefix but the implementation tests for `ignorePatternData` before inserting the `./` prefix:

    ```ts
    function findPackageLocator(location: PortablePath, {resolveIgnored = false, includeDiscardFromLookup = false}: {resolveIgnored?: boolean, includeDiscardFromLookup?: boolean} = {}): PhysicalPackageLocator | null {
      if (isPathIgnored(location) && !resolveIgnored) return null;
      let relativeLocation = ppath.relative(runtimeState.basePath, location);
      if (!relativeLocation.match(isStrictRegExp)) relativeLocation = `./${relativeLocation}` as PortablePath;
      ...
      return null;
    }
    ```

    I changed esbuild to follow the implementation to fix this test.

6. I think the fallback logic in `RESOLVE_TO_UNQUALIFIED` is incorrect. Here's the specification:

    > 8. Let `referenceOrAlias` be the entry from `parentPkg.packageDependencies` referenced by `ident`
    >
    > 9. If `referenceOrAlias` is **undefined**, then
    >
    >     1. If `manifest.enableTopLevelFallback` is **true**, then
    >
    >         1. If `parentLocator` **isn't** in `manifest.fallbackExclusionList`, then
    >
    >             1. Set `referenceOrAlias` to **RESOLVE_VIA_FALLBACK**(`manifest`, `ident`)
    >
    > 10. If `referenceOrAlias` is still **undefined**, then
    >
    >     1. Throw a resolution error
    >
    > 11. If `referenceOrAlias` is **null**, then
    >
    >     1. Note: It means that `parentPkg` has an unfulfilled peer dependency on `ident`
    >
    >     2. Throw a resolution error

    However, the test called `should use the top-level fallback if a dependency is missing because of an unfulfilled peer dependency` doesn't work with this algorithm. The problem is that `parentPkg.packageDependencies` returns `null` for `referenceOrAlias` which by the specification means no fallback should be attempted (since that only happens if `referenceOrAlias` is `undefined`). But the implementation checks for `if (dependencyReference == null) {` to do the fallback, which means the fallback should happen if `dependencyReference` is either `null` or `undefined`. I changed esbuild to follow the implementation here.

arcanis:
I've opened https://github.com/yarnpkg/berry/pull/4724 to address the problems you pointed. A few notes:

> I have updated my branch here, including incorporating the tests: https://github.com/evanw/esbuild/tree/yarn-pnp

I tried it and it worked out of the box with `lodash` as dependency! However it failed with `clipanion` and `typanion`. Repro:

```
cd $(mktemp -d)
yarn init -2
yarn add typanion
echo "import 'typanion'" > index.js
~/esbuild/esbuild ./index.js --bundle
```

I think it has something to do with the zip filesystem - running `yarn unplug typanion` (which extracts the package inside a real folder in `.yarn/unplugged`) workarounds the issue.

I have also tried to run the build on the Yarn repo, but it failed to resolve the workspaces. If you want to try it:

```
~/esbuild/esbuild packages/yarnpkg-cli/sources/cli.ts --bundle
```

> Doing a lookup for ./ in this map, which is what the first test does, gives a locator of [null, null] via the algorithm in the specification but ["root", "workspace:."] via real Yarn PnP. Thus the first test incorrectly fails because test is not found as a dependency. This seems to indicate that iteration order matters after all for duplicate packageLocation values.

That's a bug in the test; the `null` / `null` and the `root@workspace:.` entries should have the same `packageDependencies`, because they are the same package (they have the same `packageLocation`, so it wouldn't make sense for them to have different dependency sets).

> I'm assuming this means it can be ignored because it's irrelevant for implementing Yarn PnP?

Correct - it's only useful at runtime for the [PnP API](https://yarnpkg.com/advanced/pnpapi#getdependencytreeroots) consumers.

> The problem is that the specification matches with ignorePatternData after inserting the ./ prefix but the implementation tests for ignorePatternData before inserting the ./ prefix

Indeed - I originally fixed that in the `ignorePatternData` description, but forgot to fix the function definition.

arcanis:
Regarding the workspaces not resolving thing, I dig a bit, and it seems to be caused by Esbuild not detecting the PnP file. Adding a log [here](https://github.com/evanw/esbuild/commit/50b47d779d7f33c9fb5c998ac5557f284458425e#diff-ff3593e6a6cb00f927118c829760e054ccca7686eae492f9f4f18b9b851a9a20R696), I get the following (the "Foo" output); notice how pnpData is `<nil>` even for the repository root, which definitely contains a `.pnp.cjs` file.

Digging more, it seems like it's because the code never gets into `dirInfoUncached`, and thus never populates `pnpData`, but I don't understand how that could happen since `dirInfoCached` seems the only place to mutate `dirCache` 🤔

```
● [DEBUG] Resolving import "@yarnpkg/core" in directory "/Users/mael.nison/berry/packages/yarnpkg-cli/sources" of type "import-statement"

  Checking dir (cached) "/Users/mael.nison/berry/packages/yarnpkg-cli/sources" %!q(bool=true)
  Read 7 entries for directory "/Users/mael.nison/berry/packages/yarnpkg-cli/sources"
  Foo "/Users/mael.nison/berry/packages/yarnpkg-cli/sources" %!q(*resolver.pnpData=<nil>)
  Foo "/Users/mael.nison/berry/packages/yarnpkg-cli" %!q(*resolver.pnpData=<nil>)
  Foo "/Users/mael.nison/berry/packages" %!q(*resolver.pnpData=<nil>)
  Foo "/Users/mael.nison/berry" %!q(*resolver.pnpData=<nil>)
  Foo "/Users/mael.nison" %!q(*resolver.pnpData=<nil>)
  Foo "/Users" %!q(*resolver.pnpData=<nil>)
  Foo "/" %!q(*resolver.pnpData=<nil>)
  Searching for "@yarnpkg/core" in "node_modules" directories starting from "/Users/mael.nison/berry/packages/yarnpkg-cli/sources"
```

arcanis:
Scratch my previous post - the reason is just that we changed the hook format in https://github.com/yarnpkg/berry/pull/4320; before:

```js
#!/usr/bin/env node
/* eslint-disable */

try {
  Object.freeze({}).detectStrictMode = true;
} catch (error) {
  throw new Error(`The whole PnP file got strict-mode-ified, which is known to break (Emscripten libraries aren't strict mode). This usually happens when the file goes through Babel.`);
}

function $$SETUP_STATE(hydrateRuntimeState, basePath) {
  return hydrateRuntimeState(JSON.parse('{\
    "__info": [\
      "This file is automatically generated. Do not touch it, or risk",\
      "your modifications being lost. We also recommend you not to read",\
      "it either without using the @yarnpkg/pnp package, as the data layout",\
      "is entirely unspecified and WILL change from a version to another."\
    ],\
    "dependencyTreeRoots": [\
      {\
```

After:

```js
#!/usr/bin/env node
/* eslint-disable */

try {
  Object.freeze({}).detectStrictMode = true;
} catch (error) {
  throw new Error(`The whole PnP file got strict-mode-ified, which is known to break (Emscripten libraries aren't strict mode). This usually happens when the file goes through Babel.`);
}

const RAW_RUNTIME_STATE =
'{\
  "__info": [\
    "This file is automatically generated. Do not touch it, or risk",\
    "your modifications being lost."\
  ],\
  "dependencyTreeRoots": [\
    {\
```

evanw:
Thanks! I updated my branch, which now appears to be working correctly in all of these cases. So it seems like it's ready for people to try out. I'm planning to move forward here by landing this zip PR, then landing my branch, then releasing these changes as a new minor release.

PR #2021, opened 2022-02-16
fix(ts): define type error

BlackHole1:
image


evanw:
This is wrong. The type annotations are correct. See the following paragraph for why:

> 

If things worked as you suggested, then esbuild would not be able to distinguish identifier replacement from string literal replacement.

Closing this PR as invalid.

PR #2022, opened 2022-02-16
chore: gitignore editor tmp dir & lib npm with private

ycjcl868:


lib/.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org
```
evanw, line 1: Why did you do this? Doesn't this break things for people that cannot access https://registry.npmjs.org?

ycjcl868: If using [nrm](https://github.com/Pana/nrm), running `npm i` will override `registry.npmjs.org` in `package-lock.json`

PR #2034, opened 2022-02-19
Allow `void` in `onResolve` and `onLoad` callbacks

hyrious:
So that we can use empty callbacks or `return;` in the callbacks, less bytes :).

### DIFF
--- a/lib/shared/types.ts
+++ b/lib/shared/types.ts
@@ -280,9 +280,9 @@ export interface PluginBuild {
   onEnd(callback: (result: BuildResult) =>
     (void | Promise<void>)): void;
   onResolve(options: OnResolveOptions, callback: (args: OnResolveArgs) =>
-    (OnResolveResult | null | undefined | Promise<OnResolveResult | null | undefined>)): void;
+    (OnResolveResult | null | void | Promise<OnResolveResult | null | void>)): void;
   onLoad(options: OnLoadOptions, callback: (args: OnLoadArgs) =>
-    (OnLoadResult | null | undefined | Promise<OnLoadResult | null | undefined>)): void;
+    (OnLoadResult | null | void | Promise<OnLoadResult | null | void>)): void;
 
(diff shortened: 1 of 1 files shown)


evanw:
I'm not sure about this. From what I understand about TypeScript, `void` means "the return value will not be observed and can be anything" but that's not the case here because the value is observed and used. For example, this is allowed in TypeScript:

```js
let foo = () => 123
let bar: () => void = foo
```

In this case the value returned from these callbacks really does have to be either null, undefined, or an object. Anything else should be detected as a type error because it will throw an exception at run-time. The TypeScript team appears to want to force you to have to say `return undefined;` instead of `return;` to return undefined, so it sounds like requiring verbose code when using TypeScript is intentional: https://github.com/Microsoft/TypeScript/issues/25481#issuecomment-702784121.

hyrious:
Sorry for being late to reply because I was asleep. 😴

You're right that `(...) => void` in typescript accepts anything as return value, it _does_ mean "I don't care about what it returns". However that's all of it. `void` is not `any`, and it will not conquer more typings when used in other ways.

```ts
let foo = () => 123;
let bar: () => string | void = foo;
//  ~~~
//     ^ error: cannot assign `number` to `string | void`
```

In short, `(...) => void` is the special case to let `void` act like `any`, in other cases `void` is just `void` (meaning `return;` or empty block).

hyrious:
One potential downside of `void` may be:

```ts
let foo: () => void = () => 123;
let bar: () => string | void = foo;
```

Because `() => A | B` is equal to `(() => A) | (() => B)`, therefore someone could pass in anything when the union type has `void` through this way.

Say it in esbuild, it's like:

```ts
const onLoadCallback: () => void = () => 123
onLoad({ ... }, onLoadCallback)
// or
onLoad({ ... }, <() => void>() => 123)
```

Which means people will have to write `() => void` to hack this kind of usage.

PR #2035, opened 2022-02-19
disable Go compiler optimization for riscv64

piggynl:
This PR adds a temporary workaround for golang/go#51101 and can be reverted as soon as Go 1.18 or 1.17.8 is released.


piggynl:
Thank you! This helps me a lot.

PR #2038, opened 2022-02-21
add support of TS 4.7 Instantiation Expression

g-plane:
This PR added support of TypeScript 4.7 [Instantiation Expression](https://github.com/microsoft/TypeScript/pull/47607).

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -3214,6 +3214,9 @@ func (p *parser) parsePrefix(level js_ast.L, errors *deferredErrors, flags exprF
 			})}
 		}
 
+		// Allow `const a = b<c>`
+		p.trySkipTypeScriptTypeArgumentsWithBacktracking()
+
 		ref := p.storeNameInRef(name)
 		return js_ast.Expr{Loc: loc, Data: &js_ast.EIdentifier{Ref: ref}}
 
--- a/internal/js_parser/ts_parser.go
+++ b/internal/js_parser/ts_parser.go
@@ -364,6 +364,7 @@ func (p *parser) skipTypeScriptTypeWithOpts(level js_ast.L, opts skipTypeOpts) {
 					}
 					p.lexer.Next()
 				}
+				p.skipTypeScriptTypeArguments(false /* isInsideJSXElement */)
 			}
 
 		case js_lexer.TOpenBracket:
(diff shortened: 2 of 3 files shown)


evanw:
Their example code doesn't appear to work on [TypeScript nightly builds](https://www.typescriptlang.org/play?ts=4.7.0-dev.20220215#code/GYVwdgxgLglg9mABAWwIYGsCmAhOAPAHgBUA+ACgDdUAbETALkSIEpEBvAWAChFfEAnTFBD8kbRFVqZEAXwDc3ed24QEAZygoMmAMpR+MMAHNceRAF4tWUwQ0HjJObwD0zxJRp1Gdw0dbmSdglPBkQfY1kVdU1wk3wLK119X1MyAHJUACMINOYnRFcgyS8w5IiZZS5VMA1EAFF+fjh+AFlUAAcEtvbbMqMAGnrG5scXNzBMAHd3f0Du3vsBoab+EiiazUxh1o6EiemGle6yPLHEedjBw5HuIA) because the compiler considers it a syntax error:

```ts
function makeBox<T>(value: T) {
    return { value };
};

const makeStringBox = makeBox<string>;  // (value: string) => { value: string }
const stringBox = makeStringBox('abc');  // { value: string }

const ErrorMap = Map<string, Error>;  // new () => Map<string, Error>
const errorMap = new ErrorMap();  // Map<string, Error>
```

I don't think we should land this before that happens. Sometimes the TypeScript team lands features and then reverts them. I wonder if that's what happened here?

AlCalzone:
It no longer is a syntax error in the [playground](https://www.typescriptlang.org/play?ts=4.7.0-dev.20220302#code/GYVwdgxgLglg9mABAWwIYGsCmAhOAPAHgBUA+ACgDdUAbETALkSIEpEBvAWAChFfEAnTFBD8kbRFVqZEAXwDc3ed24QEAZygoMmAMpR+MMAHNceRAF4tWUwQ0HjJObwD0zxJRp1Gdw0dbmSdglPBkQfY1kVdU1wk3wLK119X1MyAHJUACMINOYnRFcgyS8w5IiZZS5VMA1EAFF+fjh+AFlUAAcEtvbbMqMAGnrG5scXNzBMAHd3f0Du3vsBoab+EiiazUxh1o6EiemGle6yPLHEedjBw5HuIA) now.

Issue #1894, opened 2021-12-29
Import issue with circular re-exports


DavidZbarsky-at:
Hi, I found another behavior change when upgrading esbuild from 0.13.5 to latest. To repro:
- clone https://github.com/DavidZbarsky-at/esbuild-imports-repro-2
- run `npm install && ./test.sh`. Should see "done!" printed
- Upgrade esbuild to latest and repeat
- Should see "TypeError: Cannot read property 'f' of undefined"

This is the smallest reduction I could get, everything else I tried removing doesn't trigger the change anymore. The code is a bit convoluted so perhaps it's tripping up the analyzer?

AlCalzone:
I recently upgraded from 0.14.1 to 0.14.9 and observed the same. In particular, these imports:
https://github.com/zwave-js/node-zwave-js/blob/6ab67d68fa8dd6eb6b2a9be9a21b32aac15c4632/packages/zwave-js/src/lib/commandclass/API.ts#L19
https://github.com/zwave-js/node-zwave-js/blob/6ab67d68fa8dd6eb6b2a9be9a21b32aac15c4632/packages/zwave-js/src/lib/commandclass/CommandClass.ts#L35

cause `CCAPI` to be `undefined` in this line:
https://github.com/zwave-js/node-zwave-js/blob/6ab67d68fa8dd6eb6b2a9be9a21b32aac15c4632/packages/zwave-js/src/lib/commandclass/CommandClass.ts#L1519

mlrawlings:
We're seeing circular export issues in our project after upgrading as well.  I did a test of the intermediate versions and the last working version is `0.14.3`.

DavidZbarsky-at:
@evanw in case you had a chance to take a look, did the repro steps work for you?

indutny-signal:
I second this, this is our repro: https://github.com/indutny-signal/esbuild-bug-repro

Running `node main.js` would print "hello world", while running `node build/main.js` would error.

indutny-signal:
As @mlrawlings said the issue was introduced between 0.14.3 and 0.14.4 (even though it wasn't published), and it comes down to this generated code:

```js
var a_exports = {};
__export(a_exports, {
  a1: () => a1,
  a2: () => a2
});
```

vs previously generated
```js
__export(exports, {
  a1: () => a1,
  a2: () => a2
});
```

I'll see if I can come up with a fix for this.

indutny-signal:
[...]

PR #2059, opened 2022-02-26
Move `module.exports` assigment to the top

indutny:
When using cjs format make sure to move `module.exports` to the top of
the generated chunk, right after the `__export()` call:

    var exports = {};
    __export(exports, {
      // ...
    });
    module.exports = __toCommonJS(exports);

This guarantees that all circular requires following these calls would
see the correct `module.exports` object.

Additionally, add an extra argument for `__reExport` to copy the data to
the `module.exports`:

    __reExport(exports, module.exports, ...)

This is required to support `export * from '..'` since now
`module.exports` wouldn't automatically receive properties otherwise.

Fix: #1894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@
 
     Previously esbuild always created the forwarding `default` import in Babel mode, even if `module.exports` had no property called `default`. This was problematic because the getter named `default` still showed up as a property on the imported namespace object, and could potentially interfere with code that iterated over the properties of the imported namespace object. With this release the getter named `default` will now only be added in Babel mode if the `default` property exists at the time of the import.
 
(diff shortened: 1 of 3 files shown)


indutny:
cc @evanw

indutny-signal:
@evanw being a maintainer of a large project like esbuild is very stressful and I totally understand if you don't have time or desire to look into things. Let me know, however, if there is anything I could do to make it easier for you to review this and other PRs. Maybe I need to add test cases, write better description of the changes? Any feedback is welcome, and thanks for building and support this project!

evanw:
My personal life is really busy right now and I currently have extremely limited time to work on esbuild. I've got a ton of stuff going on all at once. Maintaining esbuild is one of the least stressful things about my life right now haha. I'm hoping to have more free time in a month or so but right now esbuild-related stuff is just going to go slowly.

indutny:
Totally understand. Thanks for sharing, and hope you will find a way to
reduce the stress!

PR #2062, opened 2022-02-27
Drop superfluous `__name()` calls

indutny:
When the output is generated with `--keep-names` flag -
`__name(fn, "name")` statements are inserted for every function/class
with a name. However, in many situations the renamer doesn't change
the name of the function/class, and thus the call to `__name` has no
effect other than imposing an extra runtime cost on the application.

This change adds `IsKeepName` flag to `ECall` struct and uses it in
js_printer to detect such situations and omit the `__name()` from being
printed. Sadly, it would be hard to drop it completely without
`--minify` flag, but at least we can drop the runtime cost associated
with it by just replacing the `__name()` with its first argument in such
situations.


indutny:
cc @evanw ?

indutny-signal:
This still leaves:
```js
          var memoized = /* @__PURE__ */ __name(function() {
            var args = arguments, key = resolver ? resolver.apply(this, args) : args[0], cache = memoized.cache;
            if (cache.has(key)) {
              return cache.get(key);
            }
            var result = func.apply(this, args);
            memoized.cache = cache.set(key, result) || cache;
            return result;
          }, "memoized");
```

which should probably dropped too, but it might be too heavy of an optimization for the compiler. Any thoughts?

indutny:
Oh wow! This PR looks much better now. Thanks @evanw !

evanw:
> which should probably dropped too, but it might be too heavy of an optimization for the compiler. Any thoughts?

This one is pretty trivial. I just added code for that too.

> Once could argue that keeping name in this case may not be necessary? I can see how V8 would report different stack traces in case name changed, but I'm not sure how the expectation for this behavior are set amongst users.

This is the same as the other cases. The name preservation is not necessary in this case either. It's true that just like the other cases, changing the name of the relevant identifier changes the `.name` property. But if you're running another tool after esbuild runs that changes identifier names, using esbuild's `--keep-names` feature alone isn't enough. You'd need to make sure the tool that has an equivalent of the `--keep-names` feature as well. It's not esbuild's responsibility to ensure that name is preserved through such a transformation.

indutny:
Thanks for merging! You are awesome :heart:

evanw:
FYI: I am reverting this PR to fix #2149. It turns out the assumption that the symbol initializer should get the correct name if the symbol has the correct name is incorrect in the case when the symbol initializer is moved into a temporary variable, which happens sometimes during esbuild's syntax transformations.

indutny-signal:
Oh, sorry for that! This is probably going to be tricky to fix correctly. :sob:

Issue #1921, opened 2022-01-07
How to fix "Dynamic require of "os" is not supported"


enricoschaaf:
How can we use dependencies that still use require while shipping esm? Is that possible?

Minimal reproduction:
https://github.com/enricoschaaf/esbuild-issue
Just run node index.mjs. This was generated from running build which executes esbuild to bundle this file.

Problem:
The problem is that node internal modules can of course not bet bundled so I expected the require syntax would get written to import syntax if the target is esm but they are still require.

Error message:
Error: Dynamic require of "os" is not supported
    at file:///home/enrico/work/esbuild-issue/index.mjs:1:450
    at file:///home/enrico/work/esbuild-issue/index.mjs:1:906
    at file:///home/enrico/work/esbuild-issue/index.mjs:1:530
    at file:///home/enrico/work/esbuild-issue/index.mjs:1:950
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)

kzc:
As far as I know, esbuild cannot convert external "static" commonjs require statements to static esm imports - not even for the side effect free node platform built in modules. This would be a useful feature for esbuild to have, as it's a fairly common use case for NodeJS applications and libraries. One could write an esbuild plugin for that, but a comprehensive solution would require reparsing each JS source file and the build would be considerably slower.

enricoschaaf:
Ok, I thought about writing one but I had the same reservations you had. But indeed that would be a super nice feature for node apps.

hronro:
I create a new issue #1927 and I think it may be related.

kzc:
Although this functionality would be better in esbuild itself, here's a super hacky script to transform node built-in `__require()`s in esbuild `esm` output to `import` statements. It only works with non-minified esbuild output.

```js
$ cat importify-esbuild-output.js 
var fs = require("fs");
var arg = process.argv[2];
var data = !arg || arg == "-" ? fs.readFileSync(0, "utf-8") : fs.readFileSync(arg, "utf-8");;
[...]

PR #2067, opened 2022-02-28
Rewrite `require()` calls of Node built-ins to import statements when emitting ESM for Node

eduardoboucas:
When outputting ESM, any `require` calls will be replaced with the `__require` shim, since `require` will not be available. However, since external paths will not be bundled, they will generate a runtime error.

This is particularly problematic when targeting Node, since any Node built-in modules are automatically marked as external and therefore requiring them will fail. This is described in https://github.com/evanw/esbuild/issues/1921.

That case is addressed in this PR. When targeting Node, any `require` calls for Node built-ins will be replaced by import statements, since we know those are available.

## Example

Take the example below:

```js
const { extname } = require("path");

console.log(extname("one.ts"));
```

And the following command to bundle it with esbuild:

```sh
esbuild --format=esm --bundle --platform=node --outfile=out.mjs
```

### Before 🔴 

Without this PR, esbuild produces the following bundle:

```js
var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
  get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
}) : x)(function(x) {
  if (typeof require !== "undefined")
    return require.apply(this, arguments);
  throw new Error('Dynamic require of "' + x + '" is not supported');
});

// ../../tests/esbuild-issue/index.js
var { extname } = __require("path");
console.log(extname("example.js"));
```

Running the bundle generates a runtime error:

```sh
$ node out.mjs
file:///Users/eduardoboucas/Sites/evanw/esbuild/test-out/index.mjs:6
  throw new Error('Dynamic require of "' + x + '" is not supported');
        ^

Error: Dynamic require of "path" is not supported
    at file:///Users/eduardoboucas/Sites/evanw/esbuild/test-out/index.mjs:6:9
    at file:///Users/eduardoboucas/Sites/evanw/esbuild/test-out/index.mjs:10:19
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

### After ✅ 

With this PR, esbuild produces the following bundle:

```js
var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
  get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
}) : x)(function(x) {
  if (typeof require !== "undefined")
    return require.apply(this, arguments);
  throw new Error('Dynamic require of "' + x + '" is not supported');
});

// ../../tests/esbuild-issue/index.js
var { extname } = import_path;
console.log(extname("example.js"));
import import_path from "path";
```

And running the bundle produces the expected output:

```sh
$ node out.mjs
.js
```

## Notes

This issue impacted Netlify customers, as reported in https://github.com/netlify/zip-it-and-ship-it/issues/1036. We would love to contribute a fix back to esbuild, and we'll be happy to accommodate any feedback from @evanw and the esbuild community into our PR.

If the maintainers decide against merging this functionality, we'll probably add it to [our esbuild fork](https://github.com/netlify/esbuild) (which you can read about [here](https://github.com/netlify/esbuild/blob/netlify/NETLIFY.md)). For this reason, I've tried to reduce the surface area of the changes to the absolute minimum, which reflects in a couple of implementation details:

1. Passing the list of Node built-in modules from the bundler to the parser feels a bit awkward. This is due to the fact that importing the resolver from the parser would lead to an import cycle. To avoid moving the list of Node built-ins to its own package and include from the resolver and the parser, I'm passing the list into the parser.

2. The `__require` runtime shim is not being dead-code-eliminated when it's not used, as shown in the example above. We can address this in different ways, depending on what the maintainers decide to do with this PR.

### DIFF
--- a/scripts/js-api-tests.js
+++ b/scripts/js-api-tests.js
@@ -4706,6 +4706,25 @@ ${path.relative(process.cwd(), input).replace(/\\/g, '/')}:1:2: ERROR: Unexpecte
    └ entry.js ── 25b ─── 25.0%
 `)
   },
+
+  async rewriteRequireToImport({ esbuild, testDir }) {
+    const outfile = path.join(testDir, 'out.mjs');
+    await esbuild.build({
+      stdin: {
+        contents: `const { extname } = require("path"); export { extname };`,
+      },
+      bundle: true,
+      bundle: true,
+      platform: 'node',
+      format: 'esm',
+      outfile
+    })
+
+    // This needs to use relative paths to avoid breaking on Windows.
+    // Importing by absolute path doesn't work on Windows in node.
+    const result = await import('./' + path.relative(__dirname, outfile))
+    assert.strictEqual(result.extname('one.ts'), '.ts')
+  },
 }
 
 async function assertSourceMap(jsSourceMap, source) {
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -181,6 +181,12 @@ func parseFile(args parseArgs) {
 		}
 	}()
 
+	// When building for Node, we mark all Node built-in modules as safe to rewrite
+	// as import statements if needed.
+	if args.options.Platform == config.PlatformNode {
+		args.options.RewriteRequireForPaths = resolver.BuiltInNodeModules
+	}
+
 	switch loader {
 	case config.LoaderJS:
 		ast, ok := args.caches.JSCache.Parse(args.log, source, js_parser.OptionsFromConfig(&args.options))
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -287,6 +287,8 @@ type Options struct {
 	NeedsMetafile           bool
 	SourceMap               SourceMap
 	ExcludeSourcesContent   bool
+
+	RewriteRequireForPaths map[string]bool
 }
 
 type TargetFromAPI uint8
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -322,6 +322,12 @@ type parser struct {
 	latestReturnHadSemicolon bool
 	warnedThisIsUndefined    bool
 	isControlFlowDead        bool
+
+	// In certain situations, we might want to rewrite require() calls to use an
+	// import statement rather than a runtime require. When this happens, we use
+	// this map to keep track of the import statement's default namespace symbol
+	// used for each import record.
+	rewrittenRequireSymbols map[uint32]js_ast.Ref
 }
 
 type injectedSymbolSource struct {
@@ -369,6 +375,8 @@ type Options struct {
 	// the name "optionsThatSupportStructuralEquality". This is only grouped like
 	// this to make the equality comparison easier and safer (and hopefully faster).
 	optionsThatSupportStructuralEquality
+
+	rewriteRequireForPaths map[string]bool
 }
 
 type optionsThatSupportStructuralEquality struct {
@@ -423,6 +431,8 @@ func OptionsFromConfig(options *config.Options) Options {
 			unusedImportsTS:         options.UnusedImportsTS,
 			useDefineForClassFields: options.UseDefineForClassFields,
 		},
+
+		rewriteRequireForPaths: options.RewriteRequireForPaths,
 	}
 }
 
@@ -13393,13 +13403,30 @@ func (p *parser) visitExprInOut(expr js_ast.Expr, in exprIn) (js_ast.Expr, exprO
 									return js_ast.Expr{Loc: expr.Loc, Data: js_ast.ENullShared}
 								}
 
-								importRecordIndex := p.addImportRecord(ast.ImportRequire, arg.Loc, helpers.UTF16ToString(str.Value), nil)
+								importRecordPath := helpers.UTF16ToString(str.Value)
+								importRecordIndex := p.addImportRecord(ast.ImportRequire, arg.Loc, importRecordPath, nil)
+
 								if p.fnOrArrowDataVisit.tryBodyCount != 0 {
 									record := &p.importRecords[importRecordIndex]
 									record.Flags |= ast.HandlesImportErrors
 									record.ErrorHandlerLoc = p.fnOrArrowDataVisit.tryCatchLoc
 								}
 								p.importRecordsForCurrentPart = append(p.importRecordsForCurrentPart, importRecordIndex)
+
+								// All require() calls will be replaced by a runtime require if the
+								// output format is ESM. For external imports, this will lead to a
+								// runtime error. There are some situations where we can avoid that,
+								// such as when dealing with a Node built-in, for which we know we
+								// can rewrite the require as an import statement. We do that here.
+								if p.shouldRewriteRequireToImport(importRecordPath) {
+									importRef := p.newSymbol(js_ast.SymbolOther, "import_"+js_ast.GenerateNonUniqueNameFromPath(importRecordPath))
+									p.moduleScope.Generated = append(p.moduleScope.Generated, importRef)
+									p.rewrittenRequireSymbols[importRecordIndex] = importRef
+
+									return js_ast.Expr{Loc: expr.Loc, Data: &js_ast.EIdentifier{
+										Ref: importRef,
+									}}
+								}
 
 								// Create a new expression to represent the operation
 								return js_ast.Expr{Loc: expr.Loc, Data: &js_ast.ERequireString{
@@ -13582,6 +13609,10 @@ func (p *parser) convertSymbolUseToCall(ref js_ast.Ref, isSingleArgCall bool) {
 		callUse.SingleArgCallCountEstimate++
 	}
 	p.symbolCallUses[ref] = callUse
+}
+
+func (p *parser) shouldRewriteRequireToImport(path string) bool {
+	return p.options.outputFormat == config.FormatESModule && p.options.rewriteRequireForPaths[path]
 }
 
 func (p *parser) warnAboutImportNamespaceCall(target js_ast.Expr, kind importNamespaceCallKind) {
@@ -14797,6 +14828,7 @@ func newParser(log logger.Log, source logger.Source, lexer js_lexer.Lexer, optio
 		isImportItem:            make(map[js_ast.Ref]bool),
 		namedImports:            make(map[js_ast.Ref]js_ast.NamedImport),
 		namedExports:            make(map[string]js_ast.NamedExport),
+		rewrittenRequireSymbols: make(map[uint32]js_ast.Ref),
 
 		suppressWarningsAboutWeirdCode: helpers.IsInsideNodeModules(source.KeyPath.Text),
 	}
@@ -14998,6 +15030,8 @@ func Parse(log logger.Log, source logger.Source, options Options) (result js_ast
 		}
 	}
 
+	parts = p.createImportsForRewrittenRequires(parts)
+
 	// Insert a variable for "import.meta" at the top of the file if it was used.
 	// We don't need to worry about "use strict" directives because this only
 	// happens when bundling, in which case we are flatting the module scopes of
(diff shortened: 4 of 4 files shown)


jrmyio:
This seems to work quite well, however when trying to use esbuild targeting esm with `prisma` in it, it still bugged out on: https://github.com/prisma/prisma/blob/cd0ec0a3d84bbc1ea0d81e7cf9c519f18e405bc0/packages/engine-core/src/library/LibraryEngine.ts#L502 .

I guess i will continue to not be possible to have code-splitting in a node application using esbuild as long as we have to rely on esm and tons of libraries not being ready for esm.

The plugin I was using to fix `__filename` and `__dirname`:
```ts
import type { Loader, Plugin } from 'esbuild';
import fs from 'fs-extra';

export const esmPlugin = (): Plugin => ({
    name: 'esmplugin',
    setup(build) {
        build.onLoad({ filter: /.\.(js|ts|jsx|tsx)$/, namespace: 'file' }, async (args) => {
            const globalsRegex = /__(?=(filename|dirname))/g;
            let fileContent = new TextDecoder().decode(await fs.readFile(args.path));
            let variables = fileContent.match(globalsRegex)
                ? `
                import { fileURLToPath as urlESMPluginFileURLToPath } from "url";
                import { dirname as pathESMPluginDirname} from "path";
                var __filename =urlESMPluginFileURLToPath(import.meta.url);
                var __dirname = pathESMPluginDirname(urlESMPluginFileURLToPath(import.meta.url));
            `
                : '';
           

            const contents = variables + '\n' + fileContent;
            const loader = args.path.split('.').pop() as Loader;

            return {
                contents,
                loader,
            };
        });
    },
});
```

now fails with this esbuild build, where it didn't do that before:
```
import import_path from "path";
SyntaxError: Identifier 'import_path' has already been declared
```

ghaedi1993 (approved):

PH4NTOMiki:
@ghaedi1993 can you merge it?

sinoon:
hi @evanw , plase merge is pr if it will helpful the issue

ctjlewis:
Excellent work! Hope to see this merged.

jensmeindertsma:
Great work! Any chance we can see this merged?

PH4NTOMiki:
@evanw please

loynoir:
Maybe something like below, when target `node-esm`?
```js
import {createRequire} from 'module'

var __require =  requireWrapper(createRequire(import.meta.url))
```

Fix all `Dynamic require of "X" is not supported`

ctjlewis:
**It is possible to work around this for the time being.**

The following approach lets us get working ESM bundles out.  It requires `--bundle` because all other imports besides Node builtins must be inlined, so only dynamic Node builtins remain.

In the emitted bundle, esbuild adds a polyfill to check if there's a `require` defined, and throws if not: `Dynamic require of [package] is not supported`.  We will ensure there is a fallback `require` defined for the runtime to load Node builtins that remain in the bundle.

This requires `esnext` target and top-level await available in the runtime environment, as that is the only way to dynamically import the modules needed to polyfill `require` at the top of the context.

### Approach

See the `BuildOptions` below:

```ts

const ESM_REQUIRE_SHIM = `
await (async () => {
  const { dirname } = await import("path");
  const { fileURLToPath } = await import("url");

  /**
   * Shim entry-point related paths.
   */
  if (typeof globalThis.__filename === "undefined") {
    globalThis.__filename = fileURLToPath(import.meta.url);
  }
  if (typeof globalThis.__dirname === "undefined") {
    globalThis.__dirname = dirname(globalThis.__filename);
  }
  /**
   * Shim require if needed.
   */
  if (typeof globalThis.require === "undefined") {
    const { default: module } = await import("module");
    globalThis.require = module.createRequire(import.meta.url);
  }
})();
`;

/**
 * ESNext + ESM, bundle: true, and require() shim in banner.
 */
const buildOptions: BuildOptions = {
  ...common,
  format: "esm",
  target: "esnext",
  bundle: true,
  banner: {
      "js": ESM_REQUIRE_SHIM,
    },
};

esbuild(buildOptions);
```

For a minified version of the shim, you can use the following (but, in general, you should not add minified code to the top of your bundle contexts because a stranger on GitHub tells you to, but feel free to verify it):

```ts
const ESM_REQUIRE_SHIM = `
await(async()=>{let{dirname:e}=await import("path"),{fileURLToPath:i}=await import("url");if(typeof globalThis.__filename>"u"&&(globalThis.__filename=i(import.meta.url)),typeof globalThis.__dirname>"u"&&(globalThis.__dirname=e(globalThis.__filename)),typeof globalThis.require>"u"){let{default:a}=await import("module");globalThis.require=a.createRequire(import.meta.url)}})();
`;
```

### End result

The end result of these build options are a single ESM bundle with all non-builtin modules inlined, and a global `require()` shimmed for any CJS imports that are left in the bundle (e.g. builtins like `events`).

jrmyio:
@ctjlewis This looks like an interesting approach that might even fix the issues I had with prisma. However, how does esbuild know the modules required with 'require' should be part of the bundle? Doesn't esbuild follow import statements to understand what is part of the bundle?

ctjlewis:
> @ctjlewis This looks like an interesting approach that might even fix the issues I had with prisma. However, how does esbuild know the modules required with 'require' should be part of the bundle? Doesn't esbuild follow import statements to understand what is part of the bundle?

You need `bundle: true`, all non-builtin imports and requires will be inlined. The only thing that will remain are (should be) builtins and the shim provide a `require` to load those with at runtime.

scripts/js-api-tests.js
```diff
@@ -4706,6 +4706,25 @@ ${path.relative(process.cwd(), input).replace(/\\/g, '/')}:1:2: ERROR: Unexpecte
...
   },
+
+  async rewriteRequireToImport({ esbuild, testDir }) {
+    const outfile = path.join(testDir, 'out.mjs');
+    await esbuild.build({
+      stdin: {
+        contents: `const { extname } = require("path"); export { extname };`,
+      },
+      bundle: true,
+      bundle: true,
```
zachleat, line 4717: Do you want two `bundle` properties here?

scalvert:
Anything we can do to help move this along? This is an issue that's also affecting a number of libraries we maintain, and as more and more of the ecosystem is moving to ESM this appears to be picking up steam.

scripts/js-api-tests.js
```diff
@@ -4706,6 +4706,25 @@ ${path.relative(process.cwd(), input).replace(/\\/g, '/')}:1:2: ERROR: Unexpecte
...
+      platform: 'node',
+      format: 'esm',
+      outfile
+    })
+
+    // This needs to use relative paths to avoid breaking on Windows.
+    // Importing by absolute path doesn't work on Windows in node.
+    const result = await import('./' + path.relative(__dirname, outfile))
```
bajtos, line 4725: Nitpick: have you considered using `url.pathToFileURL`? Something along the following lines:

```suggestion
    const result = await import(url.pathToFileURL(outfile))
```

Quoting from https://github.com/nodejs/node/issues/34765:

> This behavior isn't restricted to windows paths, the same is true for absolute Unix paths as well. The module import specifiers only support URLs, not file paths. To import file paths, they need to be converted to URLs first on all platforms (like you mentioned).
>
> The general problem with supporting both URLs and file paths is that some URLs are also valid (but different!) file paths and vice versa. We decided to go with URLs over file paths to be consistent with other runtimes (e.g. browsers) that are also using URL-based import specifiers.

josuevalrob:
> This work for me:
> 
> ```js
> const buildOptions: BuildOptions = {
>   ...common,
>   format: "esm",
>   target: "esnext",
>   platform: "node",
>   banner: 'import { createRequire } from 'module';const require = createRequire(import.meta.url);',
>   bundle,
> };
> 
> esbuild(buildOptions);`
> ```

[0] file:///Users/josuevalencia/dog/the-dog/ui/etc/esbuild/esbuild.mjs:62
[0]                     banner: 'import { createRequire } from 'module';const require = createRequire(import.meta.url);',
[0]                                                             ^^^^^^
[0] 
[0] SyntaxError: Unexpected identifier
[0]     at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:139:18)
[0]     at ESMLoader.moduleProvider (node:internal/modules/esm/loader:236:14)
[0]     at async link (node:internal/modules/esm/module_job:67:21)
[0] yarn bundle:dev:esbuild exited with code 1

Edweis:
> This work for me:
> 
> ```js
> const buildOptions: BuildOptions = {
>   ...common,
>   format: "esm",
>   target: "esnext",
>   platform: "node",
>   banner: 'import { createRequire } from 'module';const require = createRequire(import.meta.url);',
>   bundle,
> };
> 
> esbuild(buildOptions);`
> ```

Thank you this worked for me 🙏

revmischa:
> This work for me:

No I just add this to all bundled JS
```ts
{
  banner: "await(async()=>{let{dirname:e}=await import("path"),{fileURLToPath:i}=await import("url");if(typeof globalThis.__filename>"u"&&(globalThis.__filename=i(import.meta.url)),typeof globalThis.__dirname>"u"&&(globalThis.__dirname=e(globalThis.__filename)),typeof globalThis.require>"u"){let{default:a}=await import("module");globalThis.require=a.createRequire(import.meta.url)}})();
}
```
You are fake node developer unless you comprehend this, same goes for all wimps begging for this PR to be merged

arimgibson:
Maybe I'm missing something, but is there a reason this hasn't been merged?? Seems like a pretty substantial bug, has been around for months and months, already has the PR ready, and has community support. Does anyone have insight onto why this bug still exists?

ctjlewis:
> > This work for me:
> 
> No I just add this to all bundled JS
> 
> ```ts
> {
>   banner: "await(async()=>{let{dirname:e}=await import("path"),{fileURLToPath:i}=await import("url");if(typeof globalThis.__filename>"u"&&(globalThis.__filename=i(import.meta.url)),typeof globalThis.__dirname>"u"&&(globalThis.__dirname=e(globalThis.__filename)),typeof globalThis.require>"u"){let{default:a}=await import("module");globalThis.require=a.createRequire(import.meta.url)}})();
> }
> ```
> 
> You are fake node developer unless you comprehend this, same goes for all wimps begging for this PR to be merged

Lmao I wrote this shim, I literally posted it in this thread, and I've been posting in here asking for it to be merged in so that we don't have to do it by hand.

Your license to use that code is revoked, and I will DCMA codebases where you committed it in if you don't stay out of this thread until this issue is fixed.

https://github.com/tsmodule/tsmodule/blob/f25ad02c56c66de7d79834b88b8f3eff5eeb3473/src/specification/shim.ts#L9-L34

ctjlewis:
I talked with @revmischa and he says he's just having a goof and that we all want to see this fixed. Glad that's resolved.

eduardoboucas:
Closing, as it got stale and the community seems to have found different workarounds.

GeoffreyBooth:
The purpose of `banner` is not to polyfill missing initialization code. This bug should still be fixed.

evanw:
I’m reopening this because I still plan on doing this.

evanw:
Ah whoops sorry, didn’t see that this was a PR instead of an issue. I’ll close again. In any case, I still plan on doing this.

Jamie-BitFlight:
> If your program depends on `esbuild`, you will need to add it to your externs, i.e. `{ external: ["esnext"] }`.

@ctjlewis by depends on esbuild, do you mean you bundle with esbuild, or you use esbuild as a module in your app?

shellscape:
@evanw did you get around to this?

ctjlewis:
> @ctjlewis by depends on esbuild, do you mean you bundle with esbuild, or you use esbuild as a module in your app?

I mean, "if your program contains an `esbuild` import statement, such as `import { build } from "esbuild"`".

ctjlewis:
Re this issue: The burden of adding this polyfill by default would probably cause Evan more headaches than letting us polyfill ourselves, so it looks like it's been shelved.

To get this working, ensure there's a `require` identifier defined in the global scope for all emitted bundles by adding a `createRequire(import.meta.url)` statement to `{ banner: { js } }`.

casyalex:
Is it possible to add this shim by default when target to Node env and format is ESM at least ?

rreeves8:
> Is it possible to add this shim by default when target to Node env and format is ESM at least ?

100% it should be

LucienZhang:
Rather than fixing the issue, do we really want to use `banner` to add missing features or let people create their own forks?

PR #2085, opened 2022-03-05
avoid var declaration overwritten

### LINKED ISSUE #2080: Regression in v0.14.x
Hi,
thank you for great project it really moved community forward. I caught bug in v0.14 (using vite)
If it is was reported sorry i did not found it.

there is repro repo https://github.com/btd/esbuild-bug

magic-akari:
- fix #2080 regression introduced in 4fa3d7a

For the following input code

```javascript
console.log(1, f);

function f() {
  x();
}

console.log(2, f);

var f = 1;

console.log(3, f);

function f() {
  y();
}

console.log(4, f);

var f = 2;

console.log(5, f);

```

The following output will be generated

```
1 ƒ f() {
  y();
}
2 ƒ f() {
  y();
}
3 1
4 1
5 2
```

This means that input is equivalent to the following code

```javascript
var f;
function f() {
  x();
}
function f() {
  y();
}

console.log(1, f);
console.log(2, f);

f = 1;

console.log(3, f);
console.log(4, f);

f = 2;

console.log(5, f);
```

Functions can always be safely overwritten, whereas var declaration is not.

This PR tries to preserve the var declaration.

### DIFF
--- a/internal/js_ast/js_ast.go
+++ b/internal/js_ast/js_ast.go
@@ -1482,10 +1482,11 @@ type Scope struct {
 	// This will be non-nil if this is a TypeScript "namespace" or "enum"
 	TSNamespace *TSNamespaceScope
 
-	Parent    *Scope
-	Children  []*Scope
-	Members   map[string]ScopeMember
-	Generated []Ref
+	Parent     *Scope
+	Children   []*Scope
+	Members    map[string]ScopeMember
+	HoistFnRef map[string]*Ref
+	Generated  []Ref
 
 	// The location of the "use strict" directive for ExplicitStrictMode
 	UseStrictLoc logger.Loc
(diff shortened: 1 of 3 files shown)


magic-akari:
I've encountered another issue.

input
```javascript
let x = function foo(foo) {
  var foo;
  return foo;
}
```

output
```javascript
let x = function(foo) {
  var foo2;
  return foo2;
};
```

What do you think about it? @evanw

magic-akari:
It means we should keep all the vars and the last functions.
The equivalent code is as follows.

```javascript
console.log(1, f);

// Safely delete
// function f() {
//   x();
// }

console.log(2, f);

var f = 1;

console.log(3, f);

// It should be kept since hoisting
function f() {
  y();
}

console.log(4, f);

var f = 2;

console.log(5, f);
```

evanw:
Thanks for the PR! When I took a look at this, the bug had a simpler fix that also addressed the underlying issue, so I decided to make that change instead.

PR #2091, opened 2022-03-08
accurate bytesInOutput for css that includes urls

### LINKED ISSUE #2071: bytesInOutput from metafile is incorrect for css files since v0.12.12
UPDATE: `bytesInOutput` is actually incorrect whether you supply a `publicPath` or not. As long as you have a `url(...)` somewhere in your css that imports a path using file-loader.

---

When you supply a `publicPath`, the bytesInOutput value is incorrect for css files.  Repository with minimal reproducible example can be found here: https://github.com/somebee/esbuild-bytes-offset-bug.

Take the input:

```css
body {background-image: url(./image.svg);}
```
When specifying a publicPath `/some-long-public-path/`, esbuild generates the following:

```css
/* styles.css */
body {
  background-image: url(/some-long-public-path/image-WFRGLPG5.svg);
}
```
And metafile:
```json
"outputs": {
   ...
    "dist/index.css": {
      "imports": [],
      "inputs": {
        "styles.css": {
          "bytesInOutput": 61
        }
      },
      "bytes": 94
    }
  }
```

`bytesInOutput` of styles.css is shorter than the real output, as it does not include the `/some-long-public-path` in the calculation. The real bytesInOutput is 76

haikyuu:
This fixes #2071

### Problem statement

Bytes in output is based on `len(compileResult.CSS)` which is the result of the printer. And it includes the hash inside the url instead of the actual path.

Example: `div{ background: url(hashthats25characterslong); }`

That leads to a wrong value for `bytesInOutput`

### Solution

Right before adding `bytesInOutput` to the meta file, we loop over files that have parts in the current chunk (i.e that are imported from the current chunk):
- calculate the final import path 
- add the offset (id - path) to the total offset
- substract the offset from the length of the compiled result.

This is my first PR in esbuild, so I don't have enough knowledge about the internals to make a cleaner fix. But here are some thoughts/questions I had while working on this:

1. The length of the hash is 25 bytes, so I hardcoded that. I didn't find a reusable utility to build a hash in the codebase. But that makes sense since it's only used in the bundler and the linker like so `fmt.Sprintf("%sA%08d", args.uniqueKeyPrefix, args.sourceIndex)` with `A` for asset and 8 bytes int representing the index of the asset + the unique key prefix that's 16 bytes long. --> that's 25 bytes
2. Writing the metafile is done lazily in a couple of places using the `chunk.jsonMetadataChunkCallback` with a parameter that's the final output size. So it is possible to add a similar function for bytes in output. But it's useless unless we can avoid calculating the offset and defer calculating the bytes in output to a later stage


somebee:
@evanw Any way you could look at merging this?

PR #2107, opened 2022-03-17
Update other bundlers for benchmarks

nilshelmig:
I was curious how esbuild performed against the current versions of parcel, webpack and rollup.

This PR updates the dependencies of the benchmark projects. It also removes `bench-three-spack` from `bench-three` till spack  is ready.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -820,7 +820,8 @@ bench/three: | github/three
 	for i in 1 2 3 4 5 6 7 8 9 10; do echo "import * as copy$$i from './copy$$i/Three.js'; export {copy$$i}" >> bench/three/src/entry.js; done
 	echo 'Line count:' && find bench/three/src -name '*.js' | xargs wc -l | tail -n 1
 
-bench-three: bench-three-esbuild bench-three-spack bench-three-rollup bench-three-webpack5 bench-three-parcel2
+# Leaving spack out till it's ready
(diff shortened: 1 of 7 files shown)


haikyuu:
Curious what's the result?

evanw:
Yes, me too. Does this result in any significant changes?

nilshelmig:
My Laptop died so I had to set up a new system. Now I can present you my results.

| bundler | previous time [seconds] | new time [seconds] |
| -- | -- | -- |
| esbuild | 0.266 | 0.266 |
| rollup | 29.024 | 29.144 |
| webpack 5 | 35.164 | 35.146 |
| parcel 2 | 25.094 | 25.368 |

At first I was surprised, because it didn't change significantly. At second glance most optimization of the other bundlers have been for other assets types than javascript. Three.js will not cover those optimizations. Maybe a [real world example](https://github.com/gothinkster/realworld) would be the best fit to benchmark bundling with other assets.

Each benchmark was run 5 times. The bundle size did not change.
CPU: Ryzen 5 3600
NVMe SSD (PCIe 4.0): Sabrent Rocket 4.0 1TB
RAM: 16GB DDR4@3200MHz (8GB dual channel)

evanw:
Thanks for the update. In that case I'm going to close this issue because I'm not planning to update the benchmarks at this time. Redoing the benchmarks is a lot of work and I don't think it makes sense to do it if nothing significant has changed.

PR #2120, opened 2022-03-21
feat:  buildBinary according to node_env with gops arg

zhusjfaker:
When I wanted to develop esbuild, I found that the byte file did not carry args to debugging. There was no way to debug on an IDE such as vscode or goland. All the args to 'go build' were modified according to the NodeJs environment variables.


zhusjfaker:
When I wanted to develop esbuild, I found that the byte file did not carry args to debugging. There was no way to debug on an IDE such as vscode or goland. All the args to 'go build' were modified according to the NodeJs environment variables

PR #2174, opened 2022-04-11
feat: support typeof on #private Fields

magic-akari:
TypeScript 4.7 now allows us to perform `typeof` queries on private fields.

```typescript
class Container {
    #data = "hello!";

    get data(): typeof this.#data {
        return this.#data;
    }

    set data(value: typeof this.#data) {
        this.#data = value;
    }
}
```

- https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#typeof-on-private-fields
- https://github.com/microsoft/TypeScript/pull/47696

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 ## Unreleased
 
+* Add support for parsing `typeof` on #private fields from TypeScript 4.7 ([#2174](https://github.com/evanw/esbuild/pull/2174))
+
+    The upcoming version of TypeScript now lets you use `#private` fields in `typeof` type expressions:
+
+    https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#typeof-on-private-fields
+
+    ```ts
+    class Container {
+      #data = "hello!";
+
+      get data(): typeof this.#data {
+        return this.#data;
+      }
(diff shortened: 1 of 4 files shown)


jakebailey:
We're [reverting](https://github.com/microsoft/TypeScript/pull/48959) the linked change in TS, so you probably want to revert it too.

evanw:
I think it should be ok to leave it in. The only purpose of parsing these types is to strip them, not to validate them. It's harmless if esbuild strips types that TypeScript considers a syntax error. Developers will just see the error from TypeScript and fix their code. The only harm would come if esbuild's type parser somehow behaved differently than TypeScript's type parser regarding which text is considered to be part of the type, which could happen with different ASI rules. But I don't believe that can happen here. It also sounds like this feature may be added back in the future in which case esbuild wouldn't need any further updates to support it.

That said, this case is probably a good reason for esbuild to ignore PRs for features in beta versions of TypeScript in the future.

jakebailey:
That's true; I was just bulk notifying anyone who mentioned the PR. esbuild definitely wouldn't care about stuff after the colon for emit's sake.

Breaks like this should be rare, we just didn't notice it thanks to an inconvenient type cast.

PR #2188, opened 2022-04-18
Make TS InstantiationExpr parsing more permissive

g-plane:
This PR allows:

```ts
const x4 = f<true>
if (true) {}
```

Previously it's an error.

References:

- https://github.com/microsoft/TypeScript/issues/48654
- https://github.com/microsoft/TypeScript/pull/48659

### DIFF
--- a/internal/js_parser/ts_parser.go
+++ b/internal/js_parser/ts_parser.go
@@ -866,34 +866,36 @@ func (p *parser) canFollowTypeArgumentsInExpression() bool {
 		return true
 
 	case
-		// These tokens can't follow in a call expression, nor can they start an
-		// expression. So, consider the type argument list part of an instantiation
-		// expression.
-		js_lexer.TComma,                   // foo<x>,
-		js_lexer.TDot,                     // foo<x>.
-		js_lexer.TQuestionDot,             // foo<x>?.
-		js_lexer.TCloseParen,              // foo<x>)
-		js_lexer.TCloseBracket,            // foo<x>]
-		js_lexer.TColon,                   // foo<x>:
-		js_lexer.TSemicolon,               // foo<x>;
-		js_lexer.TQuestion,                // foo<x>?
-		js_lexer.TEqualsEquals,            // foo<x> ==
-		js_lexer.TEqualsEqualsEquals,      // foo<x> ===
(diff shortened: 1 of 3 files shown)


internal/js_parser/ts_parser.go
```diff
@@ -866,34 +866,36 @@ func (p *parser) canFollowTypeArgumentsInExpression() bool {
...
+		js_lexer.TStringLiteral,
+		js_lexer.TOpenBracket,
+		js_lexer.TOpenBrace,
+		js_lexer.TFunction,
+		js_lexer.TClass,
+		js_lexer.TNew,
+		js_lexer.TSlash,
+		js_lexer.TSlashEquals,
+		js_lexer.TIdentifier:
+		return false
```
evanw, line 895: Where did these come from? This doesn't appear to follow what Microsoft changed in their parser. For example, consider the following code:

```js
f<T>
import "x"

f<T>
import("x")
```

After their PR, [TypeScript parses it like this](https://www.typescriptlang.org/play?ts=4.7.0-dev.20220418#code/GYHgKgfAUAlgtgBwPYCcAuACARADy1KUSWRVNAClywEoog):

```js
f;
import "x";

f < T > import("x");
```

But with your PR, esbuild parses it like this:

```js
f;
import "x";
f;
import("x");
```

evanw (commented):
Thanks for giving me a heads up about this change in TypeScript!

internal/js_parser/ts_parser.go
g-plane: I forgot to handle `import` token, and I will add it.

They changed to use an existing function [`isStartOfExpression`](https://github.com/microsoft/TypeScript/blob/56a4a93718ce016f038bcbce425bd3f47c3bec78/src/compiler/parser.ts#L4235).

g-plane: How can I do something similar to `lookAhead` in esbuild?

evanw: It's kind of hacky. I stash a copy of `p.lexer` on the stack and then overwrite it again after backtracking has finished. I can take this PR from here and add the backtracking.

g-plane: To be honest, I hope we can have a more common method to do backtracking.

PR #2211, opened 2022-04-27
Code splitting for cjs target

goloveychuk:
if something like this could be merged, I can add tests


pkg/api/api_impl.go
```diff
@@ -1047,8 +1047,14 @@ func rebuildImpl(
...
 	// Code splitting is experimental and currently only enabled for ES6 modules
-	if options.CodeSplitting && options.OutputFormat != config.FormatESModule {
-		log.Add(logger.Error, nil, logger.Range{}, "Splitting currently only works with the \"esm\" format")
+	if options.CodeSplitting {
+		switch options.OutputFormat {
+		case config.FormatCommonJS:
+		case config.FormatESModule:
+
+		default:
+			log.Add(logger.Error, nil, logger.Range{}, "Splitting currently only works with the \"esm\" and \"cjs\" format")
```
RDIL, line 1056:
```suggestion
			log.Add(logger.Error, nil, logger.Range{}, "Splitting currently only works with the \"esm\" and \"cjs\" formats")
```

pkg/api/api_impl.go
```diff
@@ -1047,8 +1047,14 @@ func rebuildImpl(
 	}
 
 	// Code splitting is experimental and currently only enabled for ES6 modules
```
RDIL, line 1049: No longer accurate

goloveychuk: removed

pkg/api/api_impl.go
goloveychuk: changed

goloveychuk:
added one simple test

goloveychuk:
@evanw could you please check if this is smth that could be merged?

RDIL:
@evanw I would also like to see this get a review (and eventually land, ideally).  Sorry to ping, just want to make sure it's not forgotten about.

RDIL:
@goloveychuk can you rebase this against the master branch?

goloveychuk:
rebased

RDIL:
Thanks for the PR.  While I'm not a maintainer of the project (more of just an enthusiastic user), I like the direction is going in, but I have one concern.  While the splitting works, it still emits `await import("splitted-bundle-name.js")` - since the goal is to split CommonJS, shouldn't it emit `require("splitted-bundle-name.js")` instead, or am I just missing some context here as to why that *should* be the case?

goloveychuk:
Could you send a test case?

RDIL:
@goloveychuk https://github.com/RDIL/tiny-testing-ground/blob/56fb91329d01496187b7dab487c1419f4bd19d16/build/index.js#L15 for instance.

You can clone the repo and test the full case.  I committed node_modules since I swapped out the `esbuild.wasm` file (from `npm/esbuild-wasm` with `make platform-wasm`).

goloveychuk:
If I remember correctly, and looking at the docs, dynamic import expression is allowed in cjs modules. 

https://nodejs.org/api/esm.html#import-expressions

My target is to support cjs, do this is ok to have dynamic imports, even if it's part of esm spec.

RDIL:
Fair enough.  Given that is the case, I see no problems with this PR.

goloveychuk:
@evanw gentle ping.

RDIL:
It’s been months now with nothing from the maintainer on the matter - seriously considering publishing a community fork that implements some of the PRs that people want to see merged soon, e.g. this one.  If you’re interested in that, let me know.

goloveychuk:
In our case, we found another solution, using esm target.
Because esm become to have more and more support nowadays - there is no big sense to maintain fork for this feature. 
As a duty hack - esbuild results could be post-processed with script, replacing imports and exports

PR #2302, opened 2022-06-05
Fix arrow function lowering failing to emit var _this decl

nkeynes:
Fixes issue where the synthetic _this variable was being dropped, so that:
```
function foo(a) { 
  arr.forEach(e=>this.foo(e));
}
```
was incorrectly lowered to the nonfunctional code:
```
function foo(a) {
  arr.forEach(function(e) {
    return _this.foo(e);
  });
}
```


evanw:
Thanks for the fix. This code clearly should have been covered by tests. Sorry about that.

A note for myself: Looks like this was introduced in version 0.7.2 (specifically in [this commit](https://github.com/evanw/esbuild/commit/bee2b297f8578b08238283d55dac9261b1638d45)) and regressed recently in version 0.14.30 (specifically in [this commit](https://github.com/evanw/esbuild/commit/c9e2fdb5909373d57278bde39f83bf8eafa82637)).

PR #2348, opened 2022-06-24
fix: `esm` don't need `use strict`

### LINKED ISSUE #2347: "use strict" should not be emitted when format is ESM
#2264 added `"use strict"` when the tsconfig option `alwaysStrict` is true, but it emits it even when the format is ES modules. ES modules are always in strict mode, so the pragma is unnecessary. It should only be added it the format is not ESM.

susiwen8:
- Close #2347


curiousdannii:
Beautiful, thanks!

Issue #312, opened 2020-08-03
WebWorker support?


ebrensi:
Hi I just discovered esbuild, coming from Parcel. One thing I like abiut parcel is that if I instaniate a WebWorker with the string literal filename, like
```javascript
const ww = new WebWorker ('./my worker.js')
ww.sendmessage('work!')
```
Parcel will recognize the WebWorker constructor and create another bundle starting at `./myWorker.js`.  It also handles cache busting for the filename. So the bundle would be called `./myWorker8a68r8912q.js` or something  and that string would be updated in the code above.

Does ESBuild do something like that?  If not, where would I look to implement that?

evanw:
Oh, interesting. That's pretty cool. I didn't know about that feature of Parcel. I assume this also works for the `SharedWorker` constructor. You should be able to get this to work with the current version of esbuild by just adding `myWorker.js` as an additional entry point in the list of entry points passed to esbuild.

However, the cache busting part won't work yet. None of the entry points for esbuild are cache-busted right now. I'm working on enabling this as part of general code splitting support (see #16) but it's still a work in progress. Introducing cache busting in entry point names is also a breaking change and I've been waiting to do this until the next batch of breaking changes. Once that's in, I think it should be relatively straightforward to make something like this work.

ggoodman:
This sort of feature is sorely needed in Rollup and can be achieved in Webpack with something like [worker-plugin](https://github.com/GoogleChromeLabs/worker-plugin).

Having this built into esbuild as an opt-in would be absolutely fantastic! I think deferring to plugins would likely not do it 'right' as these would need to parse an AST, figure out bindings, etc.. (or just RegEx and yolo).

ggoodman:
There are other web apis that would benefit from this form of non `require`, non `import()`-based code splitting like [audio worklets](https://developers.google.com/web/updates/2017/12/audio-worklet).

rtsao:
I think this would be a really nice feature.

As a stopgap, I was thinking this could be implemented as an esbuild plugin. This would impose some rather unfortunate syntax, but in a pinch, something like this would work:

```js
import workerUrl from "workerUrl(./worker.js)";

const worker = new Worker(workerUrl)
```


```js
import path from "path";
import { build } from "esbuild";

let workerLoader = (plugin) => {
  plugin.setName("worker-loader");
  plugin.addResolver({ filter: /^workerUrl\((.+)\)/ }, (args) => {
    return { path: args.path, namespace: "workerUrl" };
[...]

PR #2349, opened 2022-06-25
fix #334: support automatic JSX runtime

### LINKED ISSUE #334: Support jsx automatic runtime
React 17 will be adding jsx automatic runtime. It's now available in the latest release candidate.

More info [here](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154).

i.e. react/jsx-runtime and react/jsx-dev-runtime

Thanks!

### LINKED ISSUE #718: Enable `jsxImportSource`
👋🏻  @evanw, thanks for your amazing project!

We were moving a group of tooling to use `esbuild` (from rollout + tsc) and we started having errors given that we use [jsxImportSource](https://www.typescriptlang.org/tsconfig#jsxImportSource) in our tsconfig.json in order to enable the use of  `@emotion/react`. This has allowed us to avoid adding the pragma on every file. 

We got this idea from [here](https://github.com/emotion-js/emotion/issues/2111#issuecomment-732763318)

in the meantime, is there a workaround to make it work?

We tried to add `--jsx-factory=jsxs` but the module is not there.
 
Thanks.

### LINKED ISSUE #1172: Add support for jsxImportSource
I know this has been requested once in issue #718, however the described use case in that issue won't explain why adding support for jsxImportSource would be awesome.

Auto-importing those libraries (React, Preact and so on) on jsx files is awesome, however, in some projects, people are interesting in mixing different libraries relying on the JSX syntax.

The way auto-importing is currently implemented in esbuild is on a per-build option from my understanding after reading the documentation. All injected files are added to all sources. It's currently not possible to state that only `.jsx` files should get some file injected, for example, from my current understanding.

If you're using React or Preact in parts of the application and would like to implement a new feature using the Callbag JSX library, which uses another factory for JSX, of course, it wouldn't be possible to do so with esbuild unless you use Callbag JSX without using the JSX syntax. If we could conditionally apply files injections based on patterns then we could have some sources based on Callbag JSX using another extension such as `.cbjsx` or something else.

Another way to handle cases like this would be to enable Babel's `jsxImportSource` pragma, that would allow some sources to opt for the corresponding source for JSX when using Callbag JSX.

Webpack is very slow, specially when compared to esbuild, which is the main reason why I'd love to be able to convert my project from Webpack to esbuild, however Webpack is very flexible. And I think some of that flexbility could be added to esbuild without impacting on the build performance in a noticeable way.

Webpack supports the possibility of chaining loaders and apply all transformations using a set of rules and this is very useful indeed.

Currently, from my understanding, if we enable JSX parsing in `.js` files, this will have an impact on the build time because even sources from node_modules will be parsed with a more expensive parser. At least I've measured the build time following the example in the Getting Started section of the documentation and noticed that if I process all JS files as JSX, the build gets a bit slower. However, I guess it wouldn't be the case if esbuild supported configurations such as `include` and `exclude` rule options for deciding when to apply a given configuration. For example, if we specify that only `.js` files under `./src` should support JSX, then I believe the build won't become noticeable slower.

But it would allow us to specify that a different file should be injected in `.jsx` files belonging to path `src/callbag-jsx/`.

These are all ideas that would enable esbuild to support projects mixing different libraries relying on the JSX syntax, but with different implementations. Supporting `jsxImportSource` is one of them, but allowing conditional rules based on the file path is another way to support it.

If any of this is confusing, please don't hesitate to ask for more information, as I'm very keen to get all features I need implemented in esbuild so that I could finally move my project from webpack to esbuild. My back-end loads instantly, but deploying a new version of the application to our servers currently could take several minutes, mostly because generating the final build with Webpack is really slow for our project. I've performed a few tests with esbuild and it's really orders of magnitude faster than webpack or any other bundler I'm aware of. Congratulations for such achievement. I really hope to be able to convert my project to esbuild as soon as possible. It would allow us to deploy new versions of the application much quicker.

### LINKED ISSUE #2318: JSX transform: add the debug source information
Most other compilers, namely Typescript and Babel, support a "dev" mode when the compiled code would include the 5th argument of the `react/jsx-dev-runtime` function call which provides the data on the original source file.

```diff
return jsx(
  'div',
  {
    children: [
      jsx(
        'h1',
        { children: 'Hello, world!' },
        void 0,
        false,
+       { fileName: _jsxFileName, lineNumber: 14, columnNumber: 5 },
        this,
      ),
      'text text',
    ],
  },
  void 0,
  true,
+ { fileName: _jsxFileName, lineNumber: 13, columnNumber: 9 },
  this,
)
```

Reference:
- https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#dev-only-transforms= for more details
- https://babeljs.io/docs/en/babel-plugin-transform-react-jsx-source

Some tools, such as [impulse](https://github.com/kirillrogovoy/impulse), leverage that data in dev runtime.

Once there's support for the new runtime API, it should be trivial to implement.

I'm ready to contribute, but I need help with higher-level design decisions.

Related to https://github.com/evanw/esbuild/issues/334

jgoz:
fixes #334
fixes #718 
fixes #1172
fixes #2318

This adds support for the [new (automatic) JSX runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) to esbuild for both the build and transform APIs.

**New CLI flags and API options**:
- `--jsx-runtime`, `jsxRuntime` &mdash; `"automatic"` or `"classic"` (default)
- `--jsx-development`, `jsxDevelopment` &mdash; toggles development mode for the automatic runtime
- `--jsx-import-source`, `jsxImportSource` &mdash; Overrides the root import for runtime functions (default `"react"`)

**New JSX pragmas**:
- `@jsx` &mdash; sets the runtime (`"automatic"` or `"classic"`)
- `@jsxImportSource` &mdash; sets the import source (only valid with automatic runtime)

`@jsxFragment` and `@jsxFactory` are only valid with classic runtime.

**Implementation details**:
Most of the work to support the automatic runtime happens in the second parsing pass. Here is a high level overview of the logic when `automatic` runtime is enabled:

- Determine whether `Fragment` needs to be imported (from the appropriate runtime import)
- Look for `key` prop and extract it from the props argument, since it's now passed directly to a jsx function
- Inspect children and determine whether they are considered static (generally, more than 1 child) or not
- Build up arguments list to pass to the appropriate jsx function
  - In development mode, this includes "source" and "self" args and "isStaticChildren" is an argument
- Determine which jsx function to use:
  - `jsxDEV` &mdash; development mode (from `{importSource}/jsx-dev-runtime`)
  - `jsxs` &mdash; production mode, static children (from `{importSource}/jsx-runtime`)
  - `jsx` &mdash; production mode, non-static children (from `{importSource}/jsx-runtime`)
  - `createElement` &mdash; fallback mode* (from `{importSource}`)
- Record an import of the appropriate function from the appropriate source
- Insert a function call with the assembled arguments

At the end of parsing, if any JSX symbols had recorded usage, the appropriate import statement is generated for only those symbols that were actually used within the file. These are generated as external imports, unlike esbuild's runtime.

\* Fallback mode &mdash; If a component includes spread props followed by a `key` prop, `{...props} key={1}`, [Babel triggers a fallback mode](https://github.com/babel/babel/blob/4c3365f6bd67385ddbaf44c86e975025e2f8ff2a/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts#L347-L351) to `createElement`. Apparently this is a temporary special case that will be removed once spreading "key" from props is no longer supported. It looks like the TypeScript compiler also implements this special case, so it seemed prudent to be compatible here.

**Development mode**:
Aside from importing a different function from a different source (`{ jsxDEV } from 'react/jsx-dev-runtime'`), development mode also passes more arguments, one of which is the "source" argument. This contains the current filename, line number, and column number mapping to the original source location of the element.

I used the LineColumnTracker to extract those details from the element's location, which will probably incur a minor performance cost that should be documented for the `jsxDevelopment` option. The alternative was to simply omit those details, but that would defeat the purpose of supporting development mode.

**TSConfig resolving**:
Along with accepting the new options directly via CLI or API, I implemented option inference from tsconfig compiler options.

- `"jsx": "preserve"` or `"jsx": "react-native"` &rarr; preserve
- `"jsx": "react"` &rarr; classic runtime
- `"jsx": "react-jsx"` &rarr; automatic runtime
- `"jsx": "react-jsxdev"` &rarr; automatic runtime and development mode enabled

It also reads the value of `"jsxImportSource"` from tsconfig if specified.

For `react-jsx` it's important to note that it doesn't implicitly set `jsxDevelopment` to false. This is to support the case where a user sets `react-jsx` in their tsconfig but then toggles development mode directly in esbuild, e.g., by environment variable or some other means.

**esbuild vs Babel vs TS vs...**

There are a few differences between the various technologies that implement automatic JSX runtimes. Though esbuild generally follows TypeScript's behavior when there is a disagreement, I chose to follow Babel in cases where it might help the user catch invalid or undesirable behavior.

Here are the notable differences:
- Element has `__source` or `__self` props:
  - Babel: print an error about a deprecated transform plugin
  - TypeScript: allow the props
  - swc: Hard crash
  - **esbuild**: print an error &mdash; I chose to follow Babel for this one because this might help people catch configuration issues where JSX files are being parsed by multiple tools

- Element has an "implicit true" key prop, e.g. `<a key />`:
  - Babel: print an error indicating that "key" props require an explicit value
  - TypeScript: silently omit the "key" prop
  - swc: Hard crash
  - **esbuild**: print an error like Babel &mdash; this might help catch legitimate programming mistakes

- Element has spread children, e.g. `<a>{...children}</a>`
  - Babel: print an error stating that React doesn't support spread children
  - TypeScript: use static jsx function and pass children as-is, including spread operator
  - swc: same as Babel
  - **esbuild**: same as TypeScript


jgoz:
I should note that I tested this successfully on a large internal codebase that uses the automatic runtime.

evanw:
Heads up that I'm currently traveling. Given that this is a big PR, I will not be able to look at this until I get back.

jgoz:
@evanw Thanks for the heads up &mdash; safe travels.

I thought about splitting the work into smaller chunks, but I couldn't think of a natural way of doing it. If there's anything I can do to make reviewing this easier when you get back, please let me know.

alexgorbatchev:
> This contains the current filename, line number, and column number mapping to the original source location of the element.

Thank you so much for this! 🙏

alexgorbatchev:
@evanw want to bump this PR, just in case you may be back from traveling!  🗺️

internal/js_parser/js_parser.go
```diff
@@ -1515,6 +1522,56 @@ func (p *parser) callRuntime(loc logger.Loc, name string, args []js_ast.Expr) js
...
+	if !ok {
+		ref = p.newSymbol(js_ast.SymbolOther, name)
+		p.moduleScope.Generated = append(p.moduleScope.Generated, ref)
+		p.isImportItem[ref] = true
+		symbols[name] = ref
+	}
+	p.recordUsage(ref)
+	return p.handleIdentifier(loc, &js_ast.EIdentifier{Ref: ref}, identifierOpts{
+		wasOriginallyIdentifier: true,
+		matchAgainstDefines:     true, // Allow defines to rewrite imported JSX symbols
```
jgoz, line 1571: I added this after resolving conflicts because it seems like if we support rewriting Fragment and createElement, we should support rewriting jsx, jsxs, and jsxDEV.

evanw:
Didn't get to this for this release cycle, sorry. Hoping to look at this soon (this week). Don't worry about resolving merge conflicts. I can do that when I process this PR.

evanw (approved):
I have gone through all of the code in this PR and made various changes. Most of them are minor bug fixes or tweaks. I will also take care of updating the documentation about all of this on esbuild's website.

One bigger change to call out is the removal of the `--jsx-runtime` setting. I merged it into the existing `--jsx` option, which was changed into a three-state value. This is similar to how TypeScript specifies this, and seemed simpler conceptually. As a bonus, it's also shorter to specify it on the command line. So the `--jsx` setting now has three values:

* `--jsx=transform` (the default)
* `--jsx=preserve`
* `--jsx=automatic`

I also wanted to say that this PR was really well done! It's obvious that a lot of thought and care went into this. I really appreciate your clear documentation about the changes in the PR description, the clean and straightforward implementation, and all of the added tests.

pleunv:
Hi @evanw, really appreciate you taking the time to get this merged. I think this was for many people the last missing puzzle piece to make a switch to esbuild, and it's great to see this land!

mbrevda:
Yup - great to see this! https://github.com/evanw/esbuild/issues/312 next?

internal/js_parser/js_parser.go
```diff
@@ -1573,7 +1573,6 @@ func (p *parser) importJSXSymbol(loc logger.Loc, jsx JSXImport) js_ast.Expr {
 	p.recordUsage(ref)
 	return p.handleIdentifier(loc, &js_ast.EIdentifier{Ref: ref}, identifierOpts{
 		wasOriginallyIdentifier: true,
-		matchAgainstDefines:     true, // Allow defines to rewrite imported JSX symbols
```
jgoz: Makes sense, I was clearly missing context here.

jgoz (commented):
> I will also take care of updating the documentation about all of this on esbuild's website.

Thank you 🙇 

> One bigger change to call out is the removal of the `--jsx-runtime` setting.

This makes complete sense. I was trying to avoid changes to existing features/flags as much as possible, but this is clearly better.

> I also wanted to say that this PR was really well done! It's obvious that a lot of thought and care went into this. I really appreciate your clear documentation about the changes in the PR description, the clean and straightforward implementation, and all of the added tests.

Thank you for being receptive to this change and for taking the time to review this PR so thoroughly. And, more broadly, thank you for esbuild itself &mdash; it is a significant achievement in web tooling. I've learned as much about the web platform just from your release notes as I have from any other source.

httpete:
It’s marvelous. I set it to automatic and it did everything perfectly even with jsxImportSource to emotion/ MUI.

PR #2354, opened 2022-06-29
Replace deprecated theme fragment syntax with prefers-color-scheme

rtsao:
Per https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to:

> The old method of specifying images based on the theme, by using a fragment appended to the URL (`#gh-dark-mode-only` or `#gh-light-mode-only`), is deprecated and will be removed in favor of the new method described above.


evanw:
Looks like it works. Thanks!

PR #2360, opened 2022-06-30
feat: drop unused local variables

sapphi-red:
This PR partially implements dropping unused local variables.
Multi-level dropping (for example: `a` is unused and `b` depends on `a`) is not implemented.

This is useful for some cases like minifying IIFE wrapped output.

### Example
**Input**
```js
(function () {
  'use strict';

  const const1 = 0;
  let let1 = 0;
  var var1 = 0;

  const const2 = 0;
  let let2 = 0;
  var var2 = 0;
  console.log(const2, let2, var2);

})();
```

**Current Output**
```js
(function() {
  "use strict";
  let let1 = 0;
  var var1 = 0;
  const const2 = 0;
  let let2 = 0;
  var var2 = 0;
  console.log(const2, let2, var2);
})();
```

**New Output**
```js
(function() {
  "use strict";
  const const2 = 0;
  let let2 = 0;
  var var2 = 0;
  console.log(const2, let2, var2);
})();
```

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -7702,6 +7702,36 @@ func (p *parser) mangleStmts(stmts []js_ast.Stmt, kind stmtsKind) []js_ast.Stmt 
 			}
 
 		case *js_ast.SLocal:
+			// Drop unused local variables
+			//
+			// The declaration must not be exported. We can't just check for the
+			// "export" keyword because something might do "export {id};" later on.
+			// Instead we just ignore all top-level declarations for now. That means
+			// this optimization currently only applies in nested scopes.
+			//
+			// Ignore declarations if the scope is shadowed by a direct "eval" call.
+			// The eval'd code may indirectly reference this symbol and the actual
+			// use count may be greater than 0.
+			if p.currentScope != p.moduleScope && !p.currentScope.ContainsDirectEval {
+				// Ignore "var" declarations unless we are running in function body.
+				// Because we may not have visited all of their uses yet by this point.
+				if s.Kind != js_ast.LocalVar || kind == stmtsFnBody {
+					newDecls := make([]js_ast.Decl, 0)
+					for _, decl := range s.Decls {
+						if id, ok := decl.Binding.Data.(*js_ast.BIdentifier); ok {
+							if !p.isSymbolUsed(id.Ref, s.Kind == js_ast.LocalVar) && (decl.ValueOrNil.Data == nil || p.exprCanBeRemovedIfUnused(decl.ValueOrNil)) {
+								continue
+							}
+						}
+						newDecls = append(newDecls, decl)
+					}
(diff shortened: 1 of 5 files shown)


internal/js_parser/js_parser.go
```diff
@@ -7702,6 +7702,36 @@ func (p *parser) mangleStmts(stmts []js_ast.Stmt, kind stmtsKind) []js_ast.Stmt
...
+			//
+			// The declaration must not be exported. We can't just check for the
+			// "export" keyword because something might do "export {id};" later on.
+			// Instead we just ignore all top-level declarations for now. That means
+			// this optimization currently only applies in nested scopes.
+			//
+			// Ignore declarations if the scope is shadowed by a direct "eval" call.
+			// The eval'd code may indirectly reference this symbol and the actual
+			// use count may be greater than 0.
+			if p.currentScope != p.moduleScope && !p.currentScope.ContainsDirectEval {
```
sapphi-red, line 7715: If this costs too much, I think this condition could be `p.currentScope.Parent == p.moduleScope && !p.currentScope.ContainsDirectEval` instead.

internal/js_parser/js_parser_test.go
```diff
@@ -5285,3 +5285,23 @@ func TestAutoPureForWeakMap(t *testing.T) {
...
+	expectPrintedMangle(t, "const v = 0", "const v = 0;\n")
+	expectPrintedMangle(t, "{ var v = 0 }", "var v = 0;\n")
+	expectPrintedMangle(t, "{ let v = 0 }", "")
+	expectPrintedMangle(t, "{ let v = 3; let v2 = v + 5 }", "{\n  let v2 = 3 + 5;\n}\n") // does not support multi-level
+	expectPrintedMangle(t, "{ const v = 3; const v2 = v + 5 }", "")
+	expectPrintedMangle(t, "{ const v = 0 }", "")
+	expectPrintedMangle(t, "{ const v = (() => 'foo')() }", "{\n  const v = (() => \"foo\")();\n}\n")
+	expectPrintedMangle(t, "{ const v = /* #__PURE__ */ (() => 'foo')() }", "")
+	expectPrintedMangle(t, "{ const v = 0; console.log(v) }", "console.log(0);\n")
+	// expectPrintedMangle(t, "function f() { const v = 0; eval('v') }", "function f() {\n  const v = 0;\n  eval(\"v\");\n}\n")
```
sapphi-red, line 5301: This test fails without this PR.
`const v = 0;` is dropped at here.
https://github.com/evanw/esbuild/blob/698898f86615a69f52d2e00681a13fe2b8635e88/internal/js_parser/js_parser.go#L9312-L9317

sapphi-red:
@unbyte That's a good catch.
But I will not campion this PR. This is because this PR is no longer essential for me.
Feel free to create a PR that supersedes this one. 😃

PR #2372, opened 2022-07-04
fix #2365: Hermes does not fully support `ConstAndLet`

### LINKED ISSUE #2365: Does Hermes really support `ConstAndLet`?
First of all, thanks for adding the Hermes target in 4e631f5643. We're maintaining a small plugin that targets Hermes, and this makes our lives so much easier.

It looks like the target is listed to support `ConstAndLet`, and I'm wondering what that means in terms of transpilation output. I know that `let` and `const` are listed under [In Progress](https://hermesengine.dev/docs/language-features#in-progress) on their website, and the engine does parse them just fine, but I also know that it basically just interprets them as `var`. There is no support for block-scoped variables (https://github.com/facebook/hermes/issues/575#issuecomment-902048116), and you can still overwrite the values. If `ConstAndLet` assumes either of these, then it's probably a bug and should be removed.

tido64:
Remove Hermes from `ConstAndLet` as it does not fully support it.

Resolves #2365.

### DIFF
--- a/internal/compat/js_table.go
+++ b/internal/compat/js_table.go
@@ -392,7 +392,6 @@ var jsTable = map[JSFeature]map[Engine][]versionRange{
 		Edge:    {{start: v{14, 0, 0}}},
 		ES:      {{start: v{2015, 0, 0}}},
 		Firefox: {{start: v{51, 0, 0}}},
-		Hermes:  {{start: v{0, 7, 0}}},
 		IE:      {{start: v{11, 0, 0}}},
 		IOS:     {{start: v{11, 0, 0}}},
 		Node:    {{start: v{6, 0, 0}}},


internal/compat/js_table.go
```diff
@@ -392,7 +392,6 @@ var jsTable = map[JSFeature]map[Engine][]versionRange{
 		Edge:    {{start: v{14, 0, 0}}},
 		ES:      {{start: v{2015, 0, 0}}},
 		Firefox: {{start: v{51, 0, 0}}},
-		Hermes:  {{start: v{0, 7, 0}}},
```
ecreeth, line 395: This file is generated by "compat-table.js". 

I think that the way to go is add `hermes: false` here
https://github.com/evanw/esbuild/blob/71be8bc24e70609ab50a80e90a17a1f5770c89b5/scripts/compat-table.js#L145

zloirock:
Maybe it should be fixed on the `compat-table` side?

evanw:
It turned out that this was a bug with esbuild’s table generation script, not with `compat-table`: https://github.com/evanw/esbuild/issues/2365#issuecomment-1175762051

PR #2373, opened 2022-07-04
Support Loongarch LE architecture 

beyond-1234:
This PR adds support for Loongarch 64-bit LE Linux.

Please be informed:
- For now, only `Go@1.18.2` and newer suppoort Loongarch.
- The version of `golang.org/x/sys` package must be above `v0.0.0-20220624` . Details in this [commit](https://github.com/golang/sys/commit/003f7fa1e2ed999af7526a2163157549a7f4c1b6)


npm/esbuild-linux-loong64/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "esbuild-linux-loong64",
```
evanw, line 2: Note that this package name doesn't work because it's already in use, and already has content published to it: https://www.npmjs.com/package/esbuild-linux-loong64. I will have to come up with another package name instead.

Makefile
```diff
@@ -339,6 +340,9 @@ platform-linux-riscv64:
 platform-linux-s390x:
 	@$(MAKE) --no-print-directory GOOS=linux GOARCH=s390x NPMDIR=npm/esbuild-linux-s390x platform-unixlike
 
+platform-linux-loong64:
+	@$(MAKE) --no-print-directory GOOS=linux GOARCH=loong64 NPMDIR=npm/esbuild-linux-loong64 platform-unixlike
```
evanw, line 344: > * For now, only `Go@1.18.2` and newer suppoort Loongarch.

When I try to run this command, I get this error:

```
go: unsupported GOOS/GOARCH pair linux/loong64
```

My `go version`:

```
go version go1.18.3 darwin/arm64
```

That implies that Go 1.18.2 and newer do not support this architecture. This makes sense because it's [supposed to only be supported in Go 1.19 and up](https://tip.golang.org/doc/go1.19):

>  Go 1.19 adds support for the Loongson 64-bit architecture LoongArch on Linux (`GOOS=linux`, `GOARCH=loong64`).

npm/esbuild-linux-loong64/package.json
shipujin: Hi,

"esbuild-linux-loong64" sounds like a good name, Whether can contact [npmjs/esbuild-linux-loong64 package](https://www.npmjs.com/package/esbuild-linux-loong64) ,
 ask him to help deal with this package

cheers~

cheungxi: > Note that this package name doesn't work because it's already in use, and already has content published to it: https://www.npmjs.com/package/esbuild-linux-loong64. I will have to come up with another package name instead.

I have deleted this package.

PR #2412, opened 2022-07-24
test(tsconfig): reproduce extends with colliding directory name

JounQin:


evanw:
Closing as the underlying issue has been fixed. An end-to-end test is more appropriate than a snapshot test for this since that tests the real file system instead of the mock file system. Testing the real file system is important because Windows does most things differently than POSIX file systems, which typically causes lots of problems unless it has explicit test coverage.

JounQin:
Thanks for fixing this, I'm not so familiar with the source/test codes. 😂

Issue #2319, opened 2022-06-15
Uncaught TypeError: Cannot set property Sidebar of #<Object> which has only a getter     at ./src/components/Sidebar.js (Sidebar.js:9:1)


gcizman:
Hi team

i am getting this issue on Sidebar.js

// Sidebar.js
/**
 * Sidebar for shipping information details
 */

import { Sidebar } from '@coreui/coreui';
import React,{Component, useEffect, useState} from 'react';
import {httpHelper} from "./helpers/httpHelper"
  
export default Sidebar =() => {
// const products
  const Products=()=>{
  const[product, setProducts] = useState(null)

  // url to pass values from the json format.

  const url = "http://localhost:3000/product"
  const api = httpHelper()

  // fetching data as json
  useEffect(()=>{
    getProducts()
  }, [])

  const postProducts = product=>{
    api 

      .post('${url}', {body:product})
      .then(response=>getProducts())
      .catch(err=>console.log(err))
  }

  // get all products information here.

  const getProducts =() =>{
    api
    .get('${url}?_expand=shipping')
    .then(response=> {
      getProducts(response)
    }).catch(err=>console.log(err))
  }

  // our products returns when nothing on json if empty.
  if(!product) return null
  }
   

}

hyrious:
The main problem here is you're assigning things to names imported from other package. i.e.

```js
import {A} from 'a'
A = 1
```

Which is not allowed in ESM context (where you can treat the `A` as if it was defined by `const A = ...`, assigning to a constant value is not allowed), and is not _safe_ to use it in CJS conext.

When you transform this code to CJS, it becomes:

```js
var import_a = require("a");
import_a.A = 1;
```

It is not _**SAFE**_ when the package `a` export things like:

```js
Object.defineProperty(exports, "a", { get: () => 42, enumerable: true })
```

You can try this code to reproduce your error message:

```js
var import_a = {}
Object.defineProperty(import_a, "a", { get: () => 42, enumerable: true })
import_a.a = 1 // error!
```

evanw:
This code looks strange to me:

```js
export default Sidebar =() => {
```

There's the issue of it crashing when run due to how JavaScript imports work (which has nothing to do with esbuild). But it also just looks wrong because even if it didn't crash, why would you want to overwrite an import like this? Are you perhaps confused about how to export something in JavaScript?

Did you mean to use a `default` export like this instead?

```js
export default () => {
```

[...]

Issue #2183, opened 2022-04-15
dyld: Symbol not found: _SecTrustEvaluateWithError


mwang0:
env
os:   mac 10.12.6
node: v14.19.1
npm:  6.14.16
pnpm: 6.32.7

```sh
pnpm dev


> admin-web@1.0.0 dev /Users/meimei/workspace/admin-web
> vite --host

dyld: Symbol not found: _SecTrustEvaluateWithError
  Referenced from: /Users/meimei/workspace/admin-web/node_modules/.pnpm/esbuild-darwin-64@0.14.29/node_modules/esbuild-darwin-64/bin/esbuild
  Expected in: flat namespace

failed to load config from /Users/meimei/workspace/admin-web/vite.config.ts
error when starting dev server:
Error: The service was stopped
    at /Users/meimei/workspace/admin-web/node_modules/.pnpm/esbuild@0.14.29/node_modules/esbuild/lib/main.js:1332:25
    at /Users/meimei/workspace/admin-web/node_modules/.pnpm/esbuild@0.14.29/node_modules/esbuild/lib/main.js:663:9
    at Socket.afterClose (/Users/meimei/workspace/admin-web/node_modules/.pnpm/esbuild@0.14.29/node_modules/esbuild/lib/main.js:641:7)
    at Socket.emit (events.js:412:35)
    at endReadableNT (internal/streams/readable.js:1334:12)
    at processTicksAndRejections (internal/process/task_queues.js:82:21)
 ELIFECYCLE  Command failed with exit code 1.

```

evanw:
[Go's official website](https://go.dev/dl/) says that Go requires macOS 10.13 or later, and you are using macOS 10.12. So you are using an unsupported platform. Your options are to either update your OS, try to build esbuild for yourself using an older version of the Go compiler, use the WebAssembly version of esbuild, or to use another tool instead.

evanw:
I'm closing this because there's nothing to do here on esbuild's end. You'll have to use one of the workarounds described in my previous comment.

siliushi:
i have resolved this porbloms
mac: 10.12.6

XuBin63:
Unhandled Rejection (Error): register failed, invalid key getInitialState from plugin

这个报错是啥情况也搜不到解决方案。

murongGitHub:
env
os: win11
node: v16.20.2
npm: 8.19.4

`
node:internal/modules/cjs/loader:1028
  const err = new Error(message);
              ^

[...]

PR #2431, opened 2022-08-02
mod: update go version in `go mod`

likzn:
# Main Change
For now, we use Go version 1.18.4 to build in `makefile`.
I think we can update `go version` in `go mod` to make it more clearer.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
 module github.com/evanw/esbuild
 
-go 1.13
+go 1.18
 
 require golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8


likzn:
@evanw Hi, PTAL~

evanw:
No. Go 1.13 support [is deliberate](https://github.com/evanw/esbuild/blob/7bd490073081c0001f213e82eb83163c952c1211/.github/workflows/ci.yml#L134-L152). It's for [people that need it](https://github.com/evanw/esbuild/issues/2183).

likzn:
image
@evanw But i find go 1.18  still says that it requires macOS 10.13 or later. So maybe it will not break about `old OS versions`?

evanw:
There are versions of macOS older than macOS 10.13. See the link I already posted above: https://github.com/evanw/esbuild/issues/2183

>  os: mac 10.12.6

likzn:
@evanw Yes, I mean `go1.18` and `go1.13` are required the same version for macOS, both require macOS 10.13 or later. And i find you added a note. So for go1.18, this doesn't  break the os version.
```
// Support for Go 1.13 is deliberate so people can build esbuild
// themselves for old OS versions. Please do not change this.
```

evanw:
> both require macOS 10.13 or later

No. From https://go.dev/doc/go1.13:

> As [announced](https://go.dev/doc/go1.12#darwin) in the Go 1.12 release notes, Go 1.13 now requires macOS 10.11 El Capitan or later; support for previous versions has been discontinued.

likzn:
> As [announced](https://go.dev/doc/go1.12#darwin) in the Go 1.12 release notes, Go 1.13 now requires macOS 10.11 El Capitan or later; support for previous versions has been discontinued.

Thanks. But in #2183 , his macOS is 10.12.6 , it is newer than macOS 10.11. But he build fails which is strange.

likzn:
> go directive https://go.dev/ref/mod#go-mod-file-go
A go directive indicates that a module was written assuming the semantics of a given version of Go

Hi,  I think `go directive` in go mod only has an effect on syntax. For the used version of macOS, I think it is related to the version of the real `go build`. So for #2319, he will fail to compile, just because we are currently using `go1.18 compiler` to build and compile, so the version of macOS must be after 10.13. What do you think

evanw:
> `go directive` in go mod only has an effect on syntax

That's the point. If you use new syntax from Go 1.18 (e.g. generics), then esbuild will fail to compile with Go 1.13. That's bad. Enforcing Go 1.13 syntax ensures esbuild can be compiled with Go 1.13.

> So for https://github.com/evanw/esbuild/issues/2319, he will fail to compile, just because we are currently using `go1.18 compiler` to build and compile, so the version of macOS must be after 10.13.

Official esbuild releases are done with Go 1.18 (now Go 1.19) but you can build it yourself using an older Go compiler if you have an older OS. You obviously wouldn't use Go 1.18 to build for macOS 10.12 because Go 1.18 doesn't support macOS 10.12.

PR #2442, opened 2022-08-05
feat: use user customized npm registry first

gemwuu:
In some districts, China for example, downloading tarballs from the official npm registry could be very slow, we use the registry.npmmirror.com for higher speed.

### DIFF
--- a/lib/npm/node-install.ts
+++ b/lib/npm/node-install.ts
@@ -195,9 +195,16 @@ function maybeOptimizePackage(binPath: string): void {
 }
 
 async function downloadDirectlyFromNPM(pkg: string, subpath: string, binPath: string): Promise<void> {
+  // Try to use user customized registry, use official registry by default;
+  const registryFromUser = child_process.execSync(`npm config get registry `, {
+    stdio: "pipe",
+    encoding: "utf8",
+  });
+  const registry = registryFromUser.trim() || 'https://registry.npmjs.org';
+
   // If that fails, the user could have npm configured incorrectly or could not
   // have npm installed. Try downloading directly from npm as a last resort.
-  const url = `https://registry.npmjs.org/${pkg}/-/${pkg}-${ESBUILD_VERSION}.tgz`;
+  const url = `${registry}/${pkg}/-/${pkg}-${ESBUILD_VERSION}.tgz`;
   console.error(`[esbuild] Trying to download ${JSON.stringify(url)}`);
   try {
     fs.writeFileSync(binPath, extractFileFromTarGzip(await fetch(url), subpath));


mbrevda:
Might make sense to (also?) read the [configured registry](https://docs.npmjs.com/cli/v8/using-npm/config#registry) with something like [`npm config`](https://docs.npmjs.com/cli/v8/commands/npm-config)` get registry` (or it's API equivalent), as the environmental variables are [no longer populated](https://github.com/npm/rfcs/blob/main/implemented/0021-reduce-lifecycle-script-environment.md) as of [npm v7](https://github.com/npm/cli/blob/v7.24.2/CHANGELOG.md#v700-beta0-2020-08-04)

gemwuu:
> Might make sense to (also?) read the [configured registry](https://docs.npmjs.com/cli/v8/using-npm/config#registry) with something like [`npm config`](https://docs.npmjs.com/cli/v8/commands/npm-config)` get registry` (or it's API equivalent), as the environmental variables are [no longer populated](https://github.com/npm/rfcs/blob/main/implemented/0021-reduce-lifecycle-script-environment.md) as of [npm v7](https://github.com/npm/cli/blob/v7.24.2/CHANGELOG.md#v700-beta0-2020-08-04)

Good point, fixed as suggested.

mbrevda:
Now it _only_ works with npm v7+ 😄

gemwuu:
> Now it _only_ works with npm v7+ 😄
No, the `npm config` API is implemented at [v6](https://docs.npmjs.com/cli/v6/commands/npm-config), too.

gemwuu:
Gently ping @evanw

lib/npm/node-install.ts
```diff
@@ -195,9 +195,16 @@ function maybeOptimizePackage(binPath: string): void {
 }
 
 async function downloadDirectlyFromNPM(pkg: string, subpath: string, binPath: string): Promise<void> {
+  // Try to use user customized registry, use official registry by default;
+  const registryFromUser = child_process.execSync(`npm config get registry `, {
```
killagu, line 199: registry config can read from env `npm_config_regsitry`.

killagu:
LGTM

fengmk2:
Any progress?

evanw:
I don't wish to reimplement npm like this. This is supposed to already be working because esbuild installs the package using the `npm` command, which should already handle this:

https://github.com/evanw/esbuild/blob/e848de20d505e7bd99afa4d9ff6f595dde708563/lib/npm/node-install.ts#L92-L98

Can you describe your problem? Why isn't the `npm` command using its own configured registry in your case?

gemwuu:
> I don't wish to reimplement npm like this. This is supposed to already be working because esbuild installs the package using the `npm` command, which should already handle this:
> 
> https://github.com/evanw/esbuild/blob/e848de20d505e7bd99afa4d9ff6f595dde708563/lib/npm/node-install.ts#L92-L98
> 
> Can you describe your problem? Why isn't the `npm` command using its own configured registry in your case?

Oh, I see what going on here, we didn't pass the right `env` when running postinstall scripts， so npm wasn't able to read the `process.env.npm_config_registry` variable to download from our customized registry.

We'll fix this internally, and this PR shoud be closed.

mbrevda:
oh, cool. Didnt know that!

PR #2457, opened 2022-08-11
fix: don't write pnpapi binary into own pkg

jonaskuske:
583569e8f5a16b64e8e1ca23e224f9092ad4712d changed the PnP workaround so that the binary is copied into the `esbuild` package location instead of a global cache dir. However, packages should never write into their own folder as it (usually) is read-only: https://yarnpkg.com/advanced/rulebook/#packages-should-never-write-inside-their-own-folder-outside-of-postinstall

In most cases this worked because `esbuild` is automatically unplugged due to its postinstall script and therefore its folder becomes writeable. However, as soon as you disable scripts for the package or globally in `.yarnrc.yml`, `esbuild` won't be unplugged any longer, its folder becomes read-only and everything breaks. So instead, this PR writes the PnP binary to `node_modules/.esbuild/<version>/`.

### DIFF
--- a/lib/npm/node-platform.ts
+++ b/lib/npm/node-platform.ts
@@ -2,6 +2,8 @@ import fs = require('fs');
 import os = require('os');
 import path = require('path');
 
+declare const ESBUILD_VERSION: string;
+
 // This feature was added to give external code a way to modify the binary
 // path without modifying the code itself. Do not remove this because
 // external code relies on this.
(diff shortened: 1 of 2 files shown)


jonaskuske:
Used a dotfolder inside `node_modules` as that's what most tools are doing, but any location is fine really.
On that note: my PR is pretty similar to the fix committed immediately before the location was switched to esbuild's own folder: https://github.com/evanw/esbuild/commit/98b0640424b9edea65a1acc67ab4aa8e8f4fbc3c#diff-c3519ab69ed278185e3afc76a8c10dcecad25c041002933b708fff3a064c8608R105-R114. Any reason the `mkdirSync` solution wasn't enough and the binary location was changed one commit later?

jonaskuske:
After applying this patch, usage from Node works but calling the binary directly still tries to write to `esbuild/lib/pnpapi-esbuild-linux-64-esbuild` and fails because it's read-only. So `yarn node -e require('esbuild').build()` works but a simple `yarn esbuild` still fails. However I did not find the equivalent part in the Go code where the PnP workaround is implemented 🤔

evanw:
Thanks for the proposed fix.

> Any reason the `mkdirSync` solution wasn't enough and the binary location was changed one commit later?

IIRC I removed the cache logic in 583569e8f5a16b64e8e1ca23e224f9092ad4712d both because it has been a source some of maintenance overhead (cache locations outside of the project directory sometimes don't work for certain people) and because caching the binary inside the project means the memory space is automatically reclaimed when the project is directory deleted, which was not previously the case.

evanw:
Just looked into this more. This should have been caught by CI here:

https://github.com/evanw/esbuild/blob/7cae76971b37033b76c50eb665522b98d448a1a2/Makefile#L181-L183

Unfortunately I don't use Yarn myself and so I ended up configuring Yarn incorrectly when writing the test. I'll fix that as well.

Issue #795, opened 2021-02-12
URL assets bundling


edoardocavazza:
Hello!

Is there any plan to support assets bundling using the URL constructor and import.meta?

Something like this:
```js
const myImg = new URL('./assets/my-img.png', import.meta.url);
```
would be equivalent to:
```js
import myImg from './assets/my-img.png';
```
but with native browser support.

This is how we can do that with rollup https://modern-web.dev/docs/building/rollup-plugin-import-meta-assets/. I think it could fit with the "convention over configuration" policy of esbuild.

ggoodman:
@edoardocavazza you might want to take a look at #312.

evanw:
Does any bundler currently do this? What would be the semantics specifically? Would this require the loader to be `dataurl` or `file`? The syntax `import myImg` currently returns a string, not a URL object. Would this be equivalent to `import string from './assets/my-img.png'; const myImg = new URL(string, import.meta.url)` instead?

ggoodman:
Not the OP, but I'd love to see something like Webpack@5 [Asset Modules](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#asset-modules). The specific behaviour that I think would be a great fit is the following:

1. When in a JavaScript file, a call to the _global_ `URL` constructor whose first argument can be evaluated to a _path_ at build time (I'd be fine w/ only relative paths) and whose second argument is a reference to `import.meta.url` will trigger a chunk boundary in the dependency graph similar to an `import()` expression.
2. This new edge will have a new _type_ and would run through `onResolve` plugins. Plugins could do custom resolution if they want, or set a custom loader. The default behaviour would be to select the same loader as if the resolved path were an entrypoint.
3. The edge will also go through `onLoad` plugins.
4. The `new URL('./relative/path', import.meta.url)` expression will be replaced with a constant string as if it were an asset reference. This behaviour would then be consistent with how the same code would behave in a modern browser.
5. The resulting chunk would be treated like another entrypoint or `async` chunk.

Benefits:

- Parity with how the same code behaves in the browser
- Extensible across use-cases for static assets, Workers, ServiceWorkers, SharedWorkers, Worklets, ...etc.
- Clean integration with existing plugin hooks

Really excited to see where you land on this @evanw.

evanw:
> The `new URL('./relative/path', import.meta.url)` expression will be replaced with a constant string as if it were an asset reference.

[...]

PR #2470, opened 2022-08-15
Include `new URL("...", import.meta.url)` references in the import graph.

lgarron:
There are important use cases when it is valuable to get the URL of an entry point, such as for web worker instantiation: https://github.com/evanw/esbuild/issues/312

This allows `esbuild` to handle the most portable way to instantiate a web worker (which works natively in browsers/`node`, and is understood by a lot of other bundlers already):

```js
const workerURL = new URL("./worker.js", import.meta.url);

// Either construct directly:
new Worker(workerURL, { type: "module" });

// Or construct using a trampoline (needed if the worker code is on a different origin than the current page, e.g. on a CDN):
const blob = new Blob([`import "${workerURL}";`], { type: "text/javascript" });
new Worker(URL.createObjectURL(blob), { type: "module" });
```

This PR implements, thanks in part to @firien [here](https://github.com/evanw/esbuild/issues/795#issuecomment-1214399891). However, it still has issues:

- It's currently hardcoded to detect `.ts` and `.js` suffixes so I can test it. But:
  - In TypeScript, it should be possible to use import paths without a file extension.
  - Other file types (e.g. `.tsx` and `.jsx`) should be included by default, possibly any file types as long as the file exists at the referenced relative path (to fully address #795).
- I haven't added tests yet.

@evanw Could I ask if you think the approach in this PR will work, and the best way to address the issues listed above?

### DIFF
--- a/internal/js_ast/js_ast.go
+++ b/internal/js_ast/js_ast.go
@@ -440,7 +440,7 @@ func (*EIf) isExpr()                   {}
 func (*ERequireString) isExpr()        {}
 func (*ERequireResolveString) isExpr() {}
 func (*EImportString) isExpr()         {}
-func (*ERelativeURL) isExpr()         {}
+func (*ERelativeURL) isExpr()          {}
 func (*EImportCall) isExpr()           {}
 
 type EArray struct {
--- a/internal/js_ast/js_ast.go
+++ b/internal/js_ast/js_ast.go
@@ -440,6 +440,7 @@ func (*EIf) isExpr()                   {}
 func (*ERequireString) isExpr()        {}
 func (*ERequireResolveString) isExpr() {}
 func (*EImportString) isExpr()         {}
+func (*ERelativeURL) isExpr()         {}
 func (*EImportCall) isExpr()           {}
 
 type EArray struct {
@@ -731,6 +732,10 @@ type ERequireString struct {
 }
 
 type ERequireResolveString struct {
+	ImportRecordIndex uint32
+}
+
+type ERelativeURL struct {
 	ImportRecordIndex uint32
 }
 
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -14359,35 +14359,40 @@ func (p *parser) visitExprInOut(expr js_ast.Expr, in exprIn) (js_ast.Expr, exprO
 
 	case *js_ast.ENew:
 		hasSpread := false
-		relativeJS := false
+		isImportGraphURL := false
 
 		e.Target = p.visitExpr(e.Target)
-		// new url?
+		// Test for `new URL("./relative/path.js", import.meta.url)` (https://github.com/evanw/esbuild/issues/795)
 		if id, ok := e.Target.Data.(*js_ast.EIdentifier); ok {
 			symbol := p.symbols[id.Ref.InnerIndex]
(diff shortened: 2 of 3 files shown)


lgarron:
> * I haven't added tests yet.

I've added a test that covers the most important behaviour. Hopefully I've added it in the appropriate place.

lgarron:
@evanw Anything I can do to help move this forward?

This implementation matches what I believe you support as the best approach: https://github.com/evanw/esbuild/issues/312#issuecomment-800758225

evanw:
I'm planning on releasing a form of this sometime soon. You can see my work in progress version here: #2508. This will initially require code splitting to be enabled (as well as bundling and ESM output of course) as esbuild's internals currently mean each input module only shows up in one output file. It will also require paths to be explicitly relative (i.e. start with `./` or `../`) to avoid ambiguity. Otherwise it will work very similar to how `import()` already works. So for example if you do `new URL('./foo.bar', import.meta.url)` and `.bar` files are loaded with the `copy` loader then it will be copied to the output directory. But if `.bar` files are loaded with the `json` loader then it will be parsed as JSON and converted to a JavaScript file with that JSON data as the `default` export.

> * Import paths should work, e.g. `@some/package/worker` if `@some/package` exports `./worker`.

This doesn't seem like something that should be done because `new URL('foo/bar', 'http://xyz.com/abc/def')` is the same as `new URL('./foo/bar', 'http://xyz.com/abc/def')` and results in the URL `http://xyz.com/abc/foo/bar`. It seems strange to me that esbuild would intercept this syntax and interpret it in some way that's not a relative URL when it looks like it would behave as one.

lgarron:
> 'm planning on releasing a form of this sometime soon. You can see my work in progress version here: #2508.

Ooooh, thanks, I'm really glad to hear it! 😍

> This will initially require code splitting to be enabled (as well as bundling and ESM output of course) as esbuild's internals currently mean each input module only shows up in one output file. It will also require paths to be explicitly relative (i.e. start with `./` or `../`) to avoid ambiguity. Otherwise it will work very similar to how `import()` already works. So for example if you do `new URL('./foo.bar', import.meta.url)` and `.bar` files are loaded with the `copy` loader then it will be copied to the output directory. But if `.bar` files are loaded with the `json` loader then it will be parsed as JSON and converted to a JavaScript file with that JSON data as the `default` export.

That sounds pretty good to me!

Does this also meant that `new URL('./file', import.meta.url)` will pick up a relative `./file.ts`?
I just tried to test, but I'll comment on the PR with my issues.

> This doesn't seem like something that should be done because `new URL('foo/bar', 'http://xyz.com/abc/def')` is the same as `new URL('./foo/bar', 'http://xyz.com/abc/def')` and results in the URL `http://xyz.com/abc/foo/bar`. It seems strange to me that esbuild would intercept this syntax and interpret it in some way that's not a relative URL when it looks like it would behave as one.

That seems fair. I don't personally need this, but I do know some others will need this use case. I *think* it could always be handled by a simple relative import that wraps the module import.

lgarron:
> Does this also meant that `new URL('./file', import.meta.url)` will pick up a relative `./file.ts`?
> I just tried to test, but I'll comment on the PR with my issues.

Okay, I was able to test by running a production build instead of watch mode.
Looks like it works! :-D

PR #2508, opened 2022-08-31
Batch of several breaking changes

evanw:
This is a batch of several breaking changes. I'm planning to release them together in an upcoming release. TL;DR: `new URL(...)`, glob `import()`, `@esbuild/` package scope, and `master` => `main`. In detail:

* Fix some parameter names for the Go API

    This release changes some parameter names for the Go API to be consistent with the JavaScript and CLI APIs:

    * `OutExtensions` => `OutExtension`
    * `JSXMode` => `JSX`

* Add additional validation of API parameters

    The JavaScript API now does some additional validation of API parameters to catch incorrect uses of esbuild's API. The biggest impact of this is likely that esbuild now strictly only accepts strings with the `define` parameter. This would already have been a type error with esbuild's TypeScript type definitions, but it was previously not enforced for people using esbuild's API JavaScript without TypeScript.

    The `define` parameter appears at first glance to take a JSON object if you aren't paying close attention, but this actually isn't true. Values for `define` are instead strings of JavaScript code. This means you have to use `define: { foo: '"bar"' }` to replace `foo` with the string `"bar"`. Using `define: { foo: 'bar' }` actually replaces `foo` with the identifier `bar`. Previously esbuild allowed you to pass `define: { foo: false }` and `false` was automatically converted into a string, which made it more confusing to understand what `define` actually represents. Starting with this release, passing non-string values such as with `define: { foo: false }` will no longer be allowed. You will now have to write `define: { foo: 'false' }` instead.

* Interpret and rewrite `new URL(..., import.meta.url)` expressions when bundling

    Some other bundlers have adopted a convention where the syntax `new URL('./file.js', import.meta.url)` causes `file.js` to be included in the current bundling operation as an additional entry point. The `'./file.js'` string is rewritten in the bundler's output to point to the resulting generated file for that entry point in the output directory (relative to the generated file containing the `new URL(...)` syntax). This is somewhat similar to how `import('./file.js')` works except that this reference just returns a `URL` object without importing the module. That lets you pass the URL of a module to other APIs such as `new Worker(...)` that take a script URL as input.

    Previously this pattern didn't work at all with esbuild, but it will now work in esbuild starting with this release. To use it you must ensure that bundling is enabled, that the output format is set to `esm`, and that code splitting is enabled (so you need to use `--bundle --format=esm --splitting`). In addition, the path must be a relative path (i.e. it must start with either `./` or `../`). Here's what using this feature looks like:

    ```ts
    const url = new URL('./worker.ts', import.meta.url)
    const worker = new Worker(url, { type: 'module' })

    worker.onmessage = (event: MessageEvent) => {
      console.log(event.data)
    }
    ```

* Handle import paths containing wildcards

    This release introduces wildcards in import paths in two places:

    * **Entry points**

        You can now pass a string containing glob-style wildcards such as `./src/*.ts` as an entry point and esbuild will search the file system for files that match the pattern. This can be used to easily pass esbuild all files with a certain extension on the command line in a cross-platform way. Previously you had to rely on the shell to perform glob expansion, but that is obviously shell-dependent and didn't work at all on Windows. Note that to use this feature on the command line you will have to quote the pattern so it's passed verbatim to esbuild without any expansion by the shell. Here's an example:

        ```sh
        esbuild --minify "./src/*.ts" --outdir=out
        ```

        Specifically the `*` character will match any character except for the `/` character, and the `/**/` character sequence will match a path separator followed by zero or more path elements. Other wildcard operators found in glob patterns such as `?` and `[...]` are not supported.

    * **Run-time import paths**

        Import paths that are evaluated at run-time can now be bundled in certain limited situations. The import path expression must be a form of string concatenation and must start with either `./` or `../`. Each non-string expression in the string concatenation chain becomes a wildcard. The `*` wildcard is chosen unless the previous character is a `/`, in which case the `/**/*` character sequence is used.

        ```js
        // These two forms are equivalent
        const json1 = await import('./data/' + kind + '.json')
        const json2 = await import(`./data/${kind}.json`)
        ```

        This feature works with `require(...)`, `import(...)`, and `new URL(..., import.meta.url)` because these can all accept run-time expressions. It does not work with `import` and `export` statements because these cannot accept run-time expressions.

* Move all binary executable packages to the `@esbuild/` scope

    Binary package executables for esbuild are published as individual packages separate from the main `esbuild` package so you only have to download the relevant one for the current platform when you install esbuild. This release moves all of these packages under the `@esbuild/` scope to avoid collisions with 3rd-party packages. It also changes them to a consistent naming scheme that uses the `os` and `cpu` names from node.

    The package name changes are as follows:

    * `@esbuild/linux-loong64` => `@esbuild/linux-loong64` (no change)
    * `esbuild-android-64` => `@esbuild/android-x64`
    * `esbuild-android-arm64` => `@esbuild/android-arm64`
    * `esbuild-darwin-64` => `@esbuild/darwin-x64`
    * `esbuild-darwin-arm64` => `@esbuild/darwin-arm64`
    * `esbuild-freebsd-64` => `@esbuild/freebsd-x64`
    * `esbuild-freebsd-arm64` => `@esbuild/freebsd-arm64`
    * `esbuild-linux-32` => `@esbuild/linux-ia32`
    * `esbuild-linux-64` => `@esbuild/linux-x64`
    * `esbuild-linux-arm` => `@esbuild/linux-arm`
    * `esbuild-linux-arm64` => `@esbuild/linux-arm64`
    * `esbuild-linux-mips64le` => `@esbuild/linux-mips64el`
    * `esbuild-linux-ppc64le` => `@esbuild/linux-ppc64`
    * `esbuild-linux-riscv64` => `@esbuild/linux-riscv64`
    * `esbuild-linux-s390x` => `@esbuild/linux-s390x`
    * `esbuild-netbsd-64` => `@esbuild/netbsd-x64`
    * `esbuild-openbsd-64` => `@esbuild/openbsd-x64`
    * `esbuild-sunos-64` => `@esbuild/sunos-x64`
    * `esbuild-wasm` => `@esbuild/wasm`
    * `esbuild-windows-32` => `@esbuild/win32-ia32`
    * `esbuild-windows-64` => `@esbuild/win32-x64`
    * `esbuild-windows-arm64` => `@esbuild/win32-arm64`
    * `esbuild` => `esbuild` (no change)

    Normal usage of the `esbuild` package should not be affected. These name changes should only affect users of the `esbuild-wasm` package as well as tools that hard-coded the individual binary executable package names into custom esbuild downloader scripts.

* Avoid marking entry points as external

    Previously you couldn't specify `--external:*` to mark all import paths as external because that also ended up making the entry point itself external, which caused the build to fail. With this release, esbuild's `external` API parameter no longer applies to entry points so using `--external:*` is now possible.

    One additional consequence of this change is that the `kind` parameter is now required when calling the `resolve()` function in esbuild's plugin API. Previously the `kind` parameter defaulted to `entry-point`, but that no longer interacts with `external` so it didn't seem wise for this to continue to be the default. You now have to specify `kind` so that the path resolution mode is explicit.

* Disallow non-`default` imports when `assert { type: 'json' }` is present

    There is now standard behavior for importing a JSON file into an ES module using an `import` statement. However, it requires you to place the `assert { type: 'json' }` import assertion after the import path. This import assertion tells the JavaScript runtime to throw an error if the import does not end up resolving to a JSON file. On the web, the type of a file is determined by the `Content-Type` HTTP header instead of by the file extension. The import assertion prevents security problems on the web where a `.json` file may actually resolve to a JavaScript file containing malicious code, which is likely not expected for an import that is supposed to only contain pure side-effect free data.

    By default, esbuild uses the file extension to determine the type of a file, so this import assertion is unnecessary with esbuild. However, esbuild's JSON import feature has a non-standard extension that allows you to import top-level properties of the JSON object as named imports. For example, esbuild lets you do this:

    ```js
    import { version } from './package.json'
    ```

    This is useful for tree-shaking when bundling because it means esbuild will only include the the `version` field of `package.json` in your bundle. This is non-standard behavior though and doesn't match the behavior of what happens when you import JSON in a real JavaScript runtime (after adding `assert { type: 'json' }`). In a real JavaScript runtime the only thing you can import is the `default` import. So with this release, esbuild will now prevent you from importing non-`default` import names if `assert { type: 'json' }` is present. This ensures that code containing `assert { type: 'json' }` isn't relying on non-standard behavior that won't work everywhere. So the following code is now an error with esbuild when bundling:

    ```js
    import { version } from './package.json' assert { type: 'json' }
    ```

* Rename the `master` branch to `main`

    The primary branch for this repository was previously called `master` but is now called `main`. This change mirrors a similar change in many other projects.


lgarron:
I'm really looking forward to this!

I've tried testing this locally by doing `make` build in `esbuild` repo and linking the binary into `node_modules/esbuild-darwin-arm64/bin/esbuild` and `node_modules/esbuild/bin/esbuild` for a project. Unfortunately, that fails for me:

<details><pre>
/Users/lgarron/Code/git/github.com/cubing/cubing.js/node_modules/esbuild/lib/main.js:1624
  let error = new Error(`${text}${summary}`);
              ^

Error: Build failed with 3 errors:
error: panic: Internal error (while parsing "src/sites/experiments.cubing.net/cubing.js/twisty/simultaneous.ts")
error: panic: Internal error (while parsing "src/sites/experiments.cubing.net/cubing.js/play/index.ts")
error: panic: Internal error (while parsing "src/cubing/search/instantiator.ts")
    at failureErrorWithLog (/Users/lgarron/Code/git/github.com/cubing/cubing.js/node_modules/esbuild/lib/main.js:1624:15)
    at /Users/lgarron/Code/git/github.com/cubing/cubing.js/node_modules/esbuild/lib/main.js:1266:28
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  errors: [
    {
      detail: undefined,
      id: '',
      location: null,
      notes: [
        {
          location: null,
          text: 'debug.Stack (runtime/debug/stack.go:24)\n' +
            'helpers.PrettyPrintedStack (internal/helpers/stack.go:9)\n' +
            'bundler.parseFile.func1 (internal/bundler/bundler.go:186)\n' +
            'panic (runtime/panic.go:838)\n' +
            'api.importKindToResolveKind (pkg/api/api_impl.go:1595)\n' +
            'api.(*pluginImpl).onResolve.func1 (pkg/api/api_impl.go:1637)\n' +
            'bundler.RunOnResolvePlugins (internal/bundler/bundler.go:782)\n' +
            'bundler.parseFile (internal/bundler/bundler.go:442)\n' +
            'bundler.(*scanner).maybeParseFile (internal/bundler/bundler.go:1299)'
        }
      ],
      pluginName: '',
      text: 'panic: Internal error (while parsing "src/sites/experiments.cubing.net/cubing.js/twisty/simultaneous.ts")'
    },
    {
      detail: undefined,
      id: '',
      location: null,
      notes: [
        {
          location: null,
          text: 'debug.Stack (runtime/debug/stack.go:24)\n' +
            'helpers.PrettyPrintedStack (internal/helpers/stack.go:9)\n' +
            'bundler.parseFile.func1 (internal/bundler/bundler.go:186)\n' +
            'panic (runtime/panic.go:838)\n' +
            'api.importKindToResolveKind (pkg/api/api_impl.go:1595)\n' +
            'api.(*pluginImpl).onResolve.func1 (pkg/api/api_impl.go:1637)\n' +
            'bundler.RunOnResolvePlugins (internal/bundler/bundler.go:782)\n' +
            'bundler.parseFile (internal/bundler/bundler.go:442)\n' +
            'bundler.(*scanner).maybeParseFile (internal/bundler/bundler.go:1299)'
        }
      ],
      pluginName: '',
      text: 'panic: Internal error (while parsing "src/sites/experiments.cubing.net/cubing.js/play/index.ts")'
    },
    {
      detail: undefined,
      id: '',
      location: null,
      notes: [
        {
          location: null,
          text: 'debug.Stack (runtime/debug/stack.go:24)\n' +
            'helpers.PrettyPrintedStack (internal/helpers/stack.go:9)\n' +
            'bundler.parseFile.func1 (internal/bundler/bundler.go:186)\n' +
            'panic (runtime/panic.go:838)\n' +
            'api.importKindToResolveKind (pkg/api/api_impl.go:1595)\n' +
            'api.(*pluginImpl).onResolve.func1 (pkg/api/api_impl.go:1637)\n' +
            'bundler.RunOnResolvePlugins (internal/bundler/bundler.go:782)\n' +
            'bundler.parseFile (internal/bundler/bundler.go:442)\n' +
            'bundler.(*scanner).maybeParseFile (internal/bundler/bundler.go:1299)'
        }
      ],
      pluginName: '',
      text: 'panic: Internal error (while parsing "src/cubing/search/instantiator.ts")'
    }
  ],
  warnings: []
}
</pre></details>

Do you have a recommended way one could test out these changes?

evanw:
> Do you have a recommended way one could test out these changes?

If you want to try out the command line you just need to type `make` and then `./esbuild <args>`. If you want to try out the JS API you can do `const esbuild = require('./scripts/esbuild.js').installForTests()` to get the current JS API. This is [what esbuild's tests do](https://github.com/evanw/esbuild/blob/28ccdc97029acbfc985d3d87d2ec98a5946ce234/scripts/node-unref-tests.js#L10).

lgarron:
> > Do you have a recommended way one could test out these changes?
> 
> If you want to try out the command line you just need to type `make` and then `./esbuild <args>`. If you want to try out the JS API you can do `const esbuild = require('./scripts/esbuild.js').installForTests()` to get the current JS API. This is [what esbuild's tests do](https://github.com/evanw/esbuild/blob/28ccdc97029acbfc985d3d87d2ec98a5946ce234/scripts/node-unref-tests.js#L10).

I see. Unfortunately, that doesn't quite work if `esbuild` is a transitive dependency. Is there a way to replace `node_modules/esbuild` with something like this, or ideally even an `npm link` approach that would work? 🤓

evanw:
The returned object has a `ESBUILD_PACKAGE_PATH` property with the package location. You can also just look at the source for `installForTests` and see what it's doing. It just does `npm pack` and `npm install`. You could also try `npm link` if you'd like.

arcanis:
Will variadic imports be eager or lazy? It'd be quite nice if we could configure that on a case-by-case basis. Webpack does this via [comment annotations](https://webpack.js.org/api/module-methods/#magic-comments):

```ts
await import(/* webpackMode: eager */ `./icons/${name}.png`)
```

evanw:
> Will variadic imports be eager or lazy? It'd be quite nice if we could configure that on a case-by-case basis.

That would be a separate feature, and isn't addressed here. What you are calling "variadic imports" are handled the same way normal dynamic imports are handled. Right now the equivalent behavior for esbuild (in the same chunk vs. a new chunk) is configured with the [splitting](https://esbuild.github.io/api/#splitting) parameter. In the future I hope to make it possible for plugins to reconfigure which chunk(s) a given module ends up in, but that will require rewriting how esbuild's linker works. I plan to do that in a separate release.

lgarron:
> The returned object has a `ESBUILD_PACKAGE_PATH` property with the package location. You can also just look at the source for `installForTests` and see what it's doing. It just does `npm pack` and `npm install`. You could also try `npm link` if you'd like.

Thanks!

> Unfortunately, that fails for me:

Okay, I figured out what broke — it seems to be the use of a plugin together with the copy loader:

```js
// build.js
import { build } from "esbuild";

build({
  format: "esm",
  target: "es2020",
  splitting: true,
  bundle: true,
  outdir: "/tmp/esbuild-test",
  loader: { ".jpg": "copy",  },
  entryPoints: ["./entry.ts"],
  plugins: [{
    name: "refresh",
    setup(build) { console.log(build); }, // Do almost nothing
  }],
});

```

```ts
// entry.ts
const url = new URL("./cubing.js.jpg", import.meta.url).toString();
console.log(url);
```

```
// cubing.js.jpg
(A JPEG file)
```

evanw:
> Unfortunately, that fails for me:

This should be fixed now. Thanks for reporting it.

arcanis:
I tried to use this branch on our codebase, but it crashes with "The service was stopped" after running for some time. The problem seems to have started in 55f6415, since commit 84b0d07 works fine. Our codebase doesn't have workers (or rather we have some, but I disabled them while trying to pinpoint the source of the issue), nor glob patterns.

The closest we have are imports like this:

```
import lastUpdated from 'last-updated!#./**/*.ts';
```

But those are intended to go into the `last-updated` loader, which does the resolution for them (it generates a virtual file exporting the last modification time for files matching the glob pattern). The files that contain those imports seem to build fine when I set them as entry points, so I'm not sure the problem comes from there.

arcanis:
I found a reproduction! It seems to be an interaction with the zip reader, when a dynamic import is found within a dependency. In my case a minimal repro is:

```
cd $(mktemp -d)
yarn init -2
yarn add esbuild moment-mini
echo "import 'moment-mini';" > index.js

# Eventually crashes with sigkill
ESBUILD_BINARY_PATH=~/esbuild/esbuild yarn esbuild --bundle index.js

# Materializing the zip content on disk works
yarn unplug moment-mini
ESBUILD_BINARY_PATH=~/esbuild/esbuild yarn esbuild --bundle index.js
```

evanw:
> I found a reproduction! It seems to be an interaction with the zip reader, when a dynamic import is found within a dependency

Thank you very much for narrowing this down. This was exactly the problem case. It turns out the way I was generating the directory entry maps in zip files meant that one of the entries was an empty string. This was fine until I started trying to recursively visit everything, which then created paths with an infinite number of slashes at the end like `file.zip/////...` which caused the hang. This should now be fixed.

eduardoboucas:
😍 Really looking forward to these changes! Is there a rough timeline for when this might ship?

lgarron:
> 😍 Really looking forward to these changes! Is there a rough timeline for when this might ship?

Also still really looking forward to this! It will unblock a bunch of my work once I know anyone using the latest `esbuild` (and therefore any major bundler) can use code published with the `new URL(..., import.meta.url) syntax. 😀

nijikon:
@evanw are you able to share any eta on this?

lgarron:
Firefox Nightly is now able to instantiate module workers, which hopefully means that `new URL(..., import.meta.url)` will soon be valuable for web worker code splitting in all modern browsers! 😄
https://bugzilla.mozilla.org/show_bug.cgi

lgarron:
@evanw: If these breaking changes are too much to land in the short term, could I ask/suggest support for `import.meta.resolve(...)` support as an alternative to `new URL(..., import.meta.url)`? This is becoming a fairly urgent point of friction for `esbuild` for any libraries with worker code.

Since the time this PR was originally created, `import.meta.resolve(...)` has now been available in all modern browsers for several months. Unfortunately, code published using `import.meta.resolve(...)` (e.g. `npm` packages) will break when used in projects that use `esbuild`, unless `esbuild` recognizes the imported path argument as a node in the import graph.

- [Chrome](https://chromestatus.com/feature/5086507990777856)
- [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1775424)
- [Safari](https://github.com/WebKit/WebKit/pull/4507)

As long as the path argument is handled as part of the import graph, it can be:

- emitted as `import.meta.resolve(...)` for `esnext` targets, or
- it can be transpiled to `new URL(..., import.meta.url)` for older ESM browsers.
  - (Or into direct dynamic imports, in some cases.)

I'd be happy to send an updated version of [this PR](https://github.com/evanw/esbuild/pull/2470) if that would help.

probablyup:
@evanw sorry to pester but would love to get the dynamic import feature in. Is there anything I can do to help?

guilhermesimoes:
If I may, I'd like to just comment on the usage of `new URL`.

We at [PeacockTV](https://www.peacocktv.com/) and [SkyShowtime](https://www.skyshowtime.com/) need to support all kinds of TVs and set-top boxes, some of which were launched more than 10 years ago. Some of these are based on older versions of Chromium, and lack `URL` support.
I know we could add a polyfill but in order to keep our bundle size small we usually try to rewrite our code before adding one. If it would be possible to emit _something_ for older ESM targets that does not rely on `URL` it would be most appreciated.

dhdaines:
Hi, unfortunately the requirement that the paths be explicitly relative with `./` or `../` will mean that Emscripten modules compiled with `-sEXPORT_ES6=1` won't be able to find the associated WebAssembly file, since, for some reason that is hard to understand, Emscripten doesn't add that leading `./`, e.g.:
```js
  // Use bundler-friendly `new URL(..., import.meta.url)` pattern; works in browsers too.
  wasmBinaryFile = new URL('soundswallower.web.wasm', import.meta.url).href;
```
Also, if you don't add `-sENVIRIONMENT=web`, then they will just fail to bundle, because they contain [this line](https://github.com/emscripten-core/emscripten/blob/6026829bd7f5fc6b6645c2a9c85de96bbbc63675/src/shell.js#L213):
```js
scriptDirectory = require('url').fileURLToPath(new URL('./', import.meta.url));
```
which esbuild dislikes because that's not a file.  Example verbose output:

```
● [DEBUG] Resolving import "./" in directory "/home/dhd/work/kissfft/jsbuild" of type "new-url"

  Checking "./" against the external pattern "module/*"
  Read 126 entries for directory "/home/dhd/work"
  The file "/home/dhd/work/kissfft/package.json" exists
  Read 41 entries for directory "/home/dhd/work/kissfft"
  Read 16 entries for directory "/home/dhd/work/kissfft/jsbuild"
  Read 41 entries for directory "/home/dhd/work/kissfft"
  No "browser" map found in directory "/home/dhd/work/kissfft"
  Attempting to load "/home/dhd/work/kissfft/jsbuild" as a file
    Checking for file "jsbuild"
    Checking for file "jsbuild.tsx"
    Checking for file "jsbuild.ts"
    Checking for file "jsbuild.jsx"
    Checking for file "jsbuild.js"
    Checking for file "jsbuild.css"
    Checking for file "jsbuild.json"
    Failed to find file "jsbuild"
  Attempting to load "/home/dhd/work/kissfft/jsbuild" as a directory
    Read 16 entries for directory "/home/dhd/work/kissfft/jsbuild"
    No "browser" map found in directory "/home/dhd/work/kissfft/jsbuild"
    Failed to find file "/home/dhd/work/kissfft/jsbuild/index.tsx"
    Failed to find file "/home/dhd/work/kissfft/jsbuild/index.ts"
    Failed to find file "/home/dhd/work/kissfft/jsbuild/index.jsx"
    Failed to find file "/home/dhd/work/kissfft/jsbuild/index.js"
    Failed to find file "/home/dhd/work/kissfft/jsbuild/index.css"
    Failed to find file "/home/dhd/work/kissfft/jsbuild/index.json"

✘ [ERROR] Could not resolve "./"

    ../kissfft/jsbuild/kissfft.esm.mjs:109:59:
      109 │     scriptDirectory = require('url').fileURLToPath(new URL('./', import.meta.url)); // includes trailing slash
          ╵                                                            ~~~~
```
I'm not really sure if this is esbuild's problem or Emscripten's, but these cases do work with webpack...

dhdaines:
Hi! I deleted my previous comment because there were ... a lot of different things going on there.

But, in any case, even after working around some problems in Emscripten, this change does *not* work for loading WebAssembly files that its ES6 module support outputs.  You can see a minimal example here: https://github.com/dhdaines/webpack-wasm-test

by running (assuming the `breaking` branch of `esbuild` is in your PATH):
```
make
npm install
perl -i -pe s,wasmtest.wasm,./wasmtest.wasm, wasmtest.mjs
npm run build:esbuild
```

If you look in `dist-esbuild/index.js` you will see that it has output this line:
```
      wasmBinaryFile = new URL("./wasmtest-N3TKMPSX.js", import.meta.url).href;
```
where the input was:
```
  wasmBinaryFile = new URL('./wasmtest.wasm', import.meta.url).href;
```
This is obviously wrong!  It will result in the code trying to load the JavaScript wrapper file as WebAssembly.

By contrast, webpack will output the `.wasm` file with a hash as a name and then load it (depending on your devtool of course) like this:
```
  wasmBinaryFile = new URL(/* asset import */ __webpack_require__(/*! wasmtest.wasm */ "./wasmtest.wasm"), __webpack_require__.b).href;
```

There is also an issue with Emscripten's wrapper code doing `new URL('./', import.meta.url)` which sometimes breaks because both esbuild and webpack try to resolve that as a module rather than just a directory, but that's Emscripten's problem, really.

dhdaines:
Basically I think the only reasonable way for a bundler to support `new URL(..., import.meta.url)` is to try to replicate as closely as possible what the browser would do.  It should return a URL that points to the asset that you want to access, *not* an ES6 module wrapping that URL, unless of course the asset in question happens to be an ES6 module (in which case why would you need to wrap it in a module?)

dhdaines:
Should also note that in the case that the target of `new URL()` is a directory, a bundler should just pass the code through unchanged, since I can't think of any reason why you would want to resolve this to a module import (the browser will definitely *not* do this, and Node won't either, it's probably a bug that webpack does this).

AliMD:
I love esbuild and I'm waiting for you to officially support ESM capability in worker.
I really don't like having to migrate from ESBuild to another tools like Parcel.
Could you please give a timeline for this feature?
Thank you in advance.

evanw:
FYI: The glob pattern part of this PR has been released in [version 0.19.0](https://github.com/evanw/esbuild/releases/tag/v0.19.0). The documentation for this feature is here: https://esbuild.github.io/api/#glob.

ggoodman:
@evanw do you think that the URL syntax for bundling is something that you'll continue exploring at this point or are we waiting for more clear signals from the ecosystem on how to tackle these sorts of non ESM import / export bundle boundaries?

AliMD:
whats up?

matthieusieben:
FWIW, I don't think the files referenced using `new URL(..., import.meta.url)` should be processed in any way. In particular, if the referenced file is a JS, it should *not* get bundled. The reason for this is that JS files are not the only files that can be referenced this way. Any file can, for any environment. For example, a backend JS code could reference a JS asset for the front-end. These files should not be mixed together (using splitting/bundling).

If, as a user of esbuild, you need to reference another JS bundle from a `new URL(..., import.meta.url)`, then that other file should be bundled on its own, in a separate process.

This would make the use of `new URL(..., import.meta.url)` more consistent (and simpler to adopt).

matthieusieben:
If you *do* want to support adding referenced JS files to the bundle, then I would suggest to add an argument to the cli:

```
esbuild app.js --other-entry-point=worker.js
```

In this scenario, only the `new URL('worker.js', import.meta.url)` would be bundled. Any other `new URL('my-file.js', import.meta.url)` would simply be copied over as a static asset.

ArnaudBarre:
@evanw having this being merge will fix multiple issues in Vite dependency optimizer. We plan to have this working by default on Rolldown, but this is still far from being usable in Vite.

PR #2509, opened 2022-08-31
support TypeScript `satisfies`

magic-akari:
- follow https://github.com/microsoft/TypeScript/pull/46827


evanw:
This doesn't appear to be released yet, so I'm not going to do this. The TypeScript team has changed their mind and unshipped features before a release in the past. Marking this PR as `unactionable`. New TypeScript features will be implemented in esbuild after they have been shipped, not before.

magic-akari:
Yes, I will mark this pull request as `read for review` when it is released.

PR #2547, opened 2022-09-14
Fix subpath imports with Yarn PnP

### LINKED ISSUE #2545: Could not resolve "#node-web-compat" - subpath imports with esbuild and yarn pnp
**Describe the bug**
I'm facing an issue with subpath imports (as used in aws-jwt-verify) in combination with yarn berry and esbuild.
Build fails with 

```
X [ERROR] Could not resolve "#node-web-compat"

    .yarn/cache/aws-jwt-verify-npm-3.1.0-1694314cd9-b978817018.zip/node_modules/aws-jwt-verify/dist/esm/jwt-rsa.js:8:30:
      8 │ import { nodeWebCompat } from "#node-web-compat";
        ╵                               ~~~~~~~~~~~~~~~~~~

  You can mark the path "#node-web-compat" as external to exclude it from the bundle, which will
  remove this error.
```

Workaround is to use a different linker mode in yarn, e.g. pnpm

**To Reproduce**
Reproducer can be found [here](https://github.com/F43nd1r/yarn-berry-esbuild-aws-jwt-verify-reproducer).

Note: This issue was also reported [here](https://github.com/awslabs/aws-jwt-verify/issues/88), but it seems more related to esbuild.

evanw:
Node has a little-used feature called [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) which are package-internal imports that start with `#` and that go through the `imports` map in `package.json`. Previously esbuild had a bug that caused esbuild to not handle these correctly in packages installed via Yarn's "Plug'n'Play" installation strategy. The problem was that subpath imports were being checked after Yarn PnP instead of before. This PR reorders these checks, which should allow subpath imports to work in this case.

_Note: Tests are currently failing due to a bug in Yarn itself: https://github.com/yarnpkg/berry/issues/3843. This PR is blocked on that bug being fixed._

Fixes #2545


F43nd1r:
Heads up: https://github.com/yarnpkg/berry/issues/3843 was fixed in https://github.com/yarnpkg/berry/pull/4862

evanw:
I tried updating to Yarn `4.0.0-rc.22` but it looks like it has a bug regarding paths on Windows. The package manifest that Yarn generates contains `../C:/` which I believe is invalid on Windows:

```
["@tokenizer/token", [\
  ["npm:0.3.0", {\
    "packageLocation": "../../../../../../C:/Users/runneradmin/AppData/Local/Yarn/Berry/cache/@tokenizer-token-npm-0.3.0-4441352cc5-9.zip/node_modules/@tokenizer/token/",\
    "packageDependencies": [\
      ["@tokenizer/token", "npm:0.3.0"]\
    ],\
    "linkType": "HARD"\
  }]\
]],\
```

This happens because GitHub Actions puts the repo on the `D:\` drive but Yarn puts its caches on the `C:\` drive, and Yarn is attempting to use a relative path instead of an absolute path to refer to a file on another drive. You aren't allowed to navigate between drives using a relative path on Windows. Yarn would have to either use an absolute path instead of a relative path here, or Yarn would have to put their cache on the same drive as the repo.

It looks like I need to wait for this issue to be fixed in a future release of Yarn before landing this change.

evanw:
It looks like maybe Yarn 4 changed the default of `enableGlobalCache: false` to `enableGlobalCache: true` which doesn't work on Windows if there are multiple drives. But perhaps I can explicitly set `enableGlobalCache: false` for tests to work around this bug in Yarn 4.

merceyz:
It works in Yarn because internally it treats it as a posix path where there are no such limitations, it only turns it into a win32 path when passing it to the `fs` module.

Issue #2401, opened 2022-07-20
Relative path for nested sourcemaps


ASDFGerte:
When consuming existing sourcemaps, the path to source isn't updated on bundling.

For simplicity, assume the following project structure:

```text
📁
├─📁 src
│ ├─📁 sub
│ │ └─📄 other.ts
│ └─📄 index.ts
├─📁 bin
└─📁 public
```
When transpiling with tsc from src to bin, without a "sourceRoot" set, the sourcemaps as expected point relative from their resulting path in "bin" to their source in "src" (only relevant parts shown):
 - bin/index.js.map: `{ "sourceRoot": "", "sources": ["../src/index.ts"] }`
 - bin/sub/other.js.map: `{ "sourceRoot": "", "sources": ["../../src/sub/other.ts"] }`

Now bundling "bin/index.js" (the output of tsc) to "public/bundle.js", esbuild doesn't respect the path change, and adds the relative paths as-is:

 - public/bundle.js.map: `{ "sources": ["../src/index.ts", "../../src/sub/other.ts"] }`

This obviously will fail (for "other.ts", "index.ts" works by accident). I was expecting paths to get a prefix of each relative move to "public" (in nodejs-like pseudo-code, to elaborate):
```js
{ "sources": [
  path.join(path.relative("public", "bin"), "../src/index.ts") /* "../src/index.ts" by chance the same */,
  path.join(path.relative("public", "bin/sub"), "../../src/sub/other.ts") /* "../src/sub/other.ts" */
] }
```

I am not sure, whether I misconfigured something, or whether this is a bug. There is a dirty fix, setting a source-root of `"."` for typescript, and `"../src"` for esbuild. However, this seems like a hacky solution, and I'd prefer to see, if I am not simply misunderstanding some concept, and there being a better solution, where the paths don't have to be hard-coded in multiple places.

---

<details>
  <summary>Configs if necessary</summary>

`"esbuild": "^0.14.49"`

`tsconfig.json` (unrelated comments removed, TS 4.7.4):
```jsonc
{
  "include": ["src/**/*"],
  "compilerOptions": {
    "target": "ESNext",
    "jsx": "react-jsxdev",
    "useDefineForClassFields": true,
    "module": "NodeNext",
    "rootDir": "./src",
    "allowJs": true,
    "sourceMap": true,
    // "sourceRoot": ".",
    "outDir": "./bin",
    "removeComments": true,
    "importHelpers": true,
    "newLine": "lf",
[...]

Issue #2283, opened 2022-05-30
Improve source maps, when using plugins for resolving imports


bxantus:
I'm using [esbuild_deno_loader](https://github.com/lucacasonato/esbuild_deno_loader) plugin with es builder to bundle my code (I'm developing for the web, and using deno instead of node.js)

Everything works fine, except for the sourcemaps, as the sources section contains the resolved paths together with the namespace provided by the deno_loader plugin.
The plugin resolves both local files and remote http/https modules in the deno namespace.  

Here are some examples:

```json
{
  "version": 3,
  "sources": 
 [ "deno:file:///C:/work/travelmate/src/map/placeData.ts", 
   "deno:file:///C:/work/travelmate/libs/dmsg/ts/src/lib/location/distance.ts",
  "deno:https://raw.githubusercontent.com/bxantus/xdom/v0.9.2/src/dispose.ts"
]
```

Reading the plugin development guide, this is what it should do, but I think esbuild shouldn't prepend source paths with the namespace inside source maps.
In my opinion the namespace is something different than the URI protocol, so I think we should rely here on the paths returned from the plugin.

I successfully patched the plugin locally, so it resolved local files inside the `file` namespace, but I feel this should be addressed rather here in esbuild.
Let me know what you think, also @lucacasonato what's your opinion about this?

hyrious:
It is possible to change that field by adding inline sourcemap to virtual modules returned by plugins, namely:

```js
onLoad({ filter: /.*/ }, async args => {
  let contents = await esbuild.transform(code, {
    sourcefile: "filename.js",
    sourcemap: "inline"
  })
  return { contents }
})
```

bxantus:
@hyrious: I'm not sure about your comment. I would like to leave the possibility to have sourcemaps not inlined.
Also I don't fully understand what this onLoad code does.
You mean this is the preferred way to solve this issue?

I personally would like it to be addressed automatically or by some kind of config, about hove to treat namespaces when listing filenames.

To avoid further confusion here's how I invoke esbuild from Deno:
```ts
import * as path from "https://deno.land/std@0.141.0/path/mod.ts"
import * as esbuild from "https://deno.land/x/esbuild@v0.14.39/mod.js";
import { denoPlugin } from "https://deno.land/x/esbuild_deno_loader@0.5.0/mod.ts";
// ...
await esbuild.build({
    plugins: [denoPlugin({ importMapURL: path.toFileUrl(path.resolve("src/import_map.json")) })],
[...]

PR #2554, opened 2022-09-16
feat: add support for source-base option

### LINKED ISSUE #2218: Source map sources related to temporary output path
Hi, Evan

Thanks for a wonderful library!

I have a question / something regarding the `sources` property of outputted (inline) sourcemaps. Given I have two files,

```mjs
// foo.mjs
export default function foo () {
  console.log("YOLO!");
}
```

.. and

```mjs
// index.mjs
import foo from "./foo.mjs"

foo();
```

.. and I build like the following

```mjs
esbuild.build({
  entryPoints: ["index.mjs"],
  bundle: true,
  sourcemap: "inline"
});
```

.. I get a nice `sources` property containing `["foo.mjs", "index.mjs"]`. However, this is all outputted to stdout. I'm using `esbuild` in part with Cypress, a testing utility, and preprocessors of Cypress are provided a temporary file location in which they are expected to output bundles. Hence, we build like

```mjs
esbuild.build({
  entryPoints: ["index.mjs"],
  bundle: true,
  sourcemap: "inline",
  outfile: "/tmp/tmp-1153466-5Fh3yAWuH7tu" // <- example value
});
```

At that point, the `sources` property contains

```json
[
  "../home/jonas/projects/esbuikd-sourcemap-example/foo.mjs",
  "../home/jonas/projects/esbuikd-sourcemap-example/index.mjs"
]
```

.. and it becomes impossible to use the sources and determine file locations programmatically without also knowing the build-time CWD and the temporary output directory.

Is there a way to build to a location *and* retain "nice" and library-friendly values for `sources`?

Regards

Jonas

Patrick-Beuks:
Add support-base as option. This can be used when your output file is not in at the base where you want to calculate your sources from

This is inline with settings from other bundlers:
https://webpack.js.org/plugins/source-map-dev-tool-plugin/ -> fileContext
https://parceljs.org/plugin-system/source-maps/ -> projectRoot
https://github.com/terser/terser -> source-map -> base

Fixes #2218


internal/bundler/linker.go
```diff
@@ -5850,6 +5850,16 @@ func (c *linkerContext) generateSourceMapForChunk(
...
+    var sourceAbsDir string
+    if c.options.SourceBase != "" {
+        var succes bool
+        sourceAbsDir, succes = c.fs.Abs(c.options.SourceBase)
+        if !succes {
+            panic("Could not determine source base")
+        }
+    } else {
+        sourceAbsDir = chunkAbsDir
+    }
```
Patrick-Beuks, line 5862: If there is a better way to do this if statement let me know. This is the first time writing go

internal/bundler/linker.go
```diff
@@ -5850,6 +5850,16 @@ func (c *linkerContext) generateSourceMapForChunk(
 	dataForSourceMaps []dataForSourceMap,
 	canHaveShifts bool,
 ) (pieces sourcemap.SourceMapPieces) {
+    var sourceAbsDir string
+    if c.options.SourceBase != "" {
+        var succes bool
+        sourceAbsDir, succes = c.fs.Abs(c.options.SourceBase)
+        if !succes {
+            panic("Could not determine source base")
+        }
```
Patrick-Beuks, line 5859: Let me know if there is a better way to handle errors

Patrick-Beuks:
Might fix the following issues
#2401 #2283

internal/bundler/linker.go
```diff
@@ -5850,6 +5850,16 @@ func (c *linkerContext) generateSourceMapForChunk(
 	dataForSourceMaps []dataForSourceMap,
 	canHaveShifts bool,
 ) (pieces sourcemap.SourceMapPieces) {
+    var sourceAbsDir string
+    if c.options.SourceBase != "" {
+        var succes bool
```
jessica-jorgenson, line 5855:
```suggestion
        var success bool
```

internal/bundler/linker.go
```diff
@@ -5850,6 +5850,16 @@ func (c *linkerContext) generateSourceMapForChunk(
...
+        sourceAbsDir, succes = c.fs.Abs(c.options.SourceBase)
```
jessica-jorgenson, line 5856:
```suggestion
        sourceAbsDir, success = c.fs.Abs(c.options.SourceBase)
```

internal/bundler/linker.go
```diff
@@ -5850,6 +5850,16 @@ func (c *linkerContext) generateSourceMapForChunk(
...
+        if !succes {
```
jessica-jorgenson, line 5857:
```suggestion
        if !success {
```

jessica-jorgenson (changes requested):
Spelling

Patrick-Beuks:
Due to lack of time to update PR to resolve merge conflicts and update to latest changes I am closing this PR.

If someone wants to pick this PR up and needs information how/why this works. You can contact me.

Issue #2555, opened 2022-09-17
[css-minifier] Correctly parsed and compressed <keyframes-name> = <string> 


yisibl:
## Spec

```
<keyframes-name> = <custom-ident> | <string>
```

https://drafts.csswg.org/css-animations/#typedef-keyframes-name

Often `<string>` can be compressed to `<custom-ident>`, However, **CSS-wide keywords** and `none` must remain in quotation marks.

## Input

```css
@keyframes "one two" {
  0% {
    left: 100px;
  }
}

@keyframes "none" {
  to {
    right: 100px;
  }
}

.foo {
  animation-name: "none", "unset", "one two";
}
```

## Actual  output

This WPT PR has more detailed test cases. https://github.com/web-platform-tests/wpt/pull/35903

```css
@keyframes ""{0%{left:100px}}@keyframes ""{to{right:100px}}.foo{animation-name:"none","unset","one two"}
```

## Expected

```css
@keyframes one\ two{0%{left:100px}}@keyframes"none"{to{right:100px}}.foo{animation-name:"none","unset",one\ two}
```

yisibl:
Thanks for the quick fix, but this issue is also expected to compress `<string>` and handle CSS-wide keywords correctly.

The ultimate goal is for esbuild to parse `<keyframes-name>` in its entirety.

I think @CGQAQ's PR is working on these issues and hopefully will reopen this issue.

PR #2556, opened 2022-09-17
feat: @keyframes now support <string> type AND will convert it into ident if possible...

CGQAQ:
... Plus `animation` and `animation-name` mangling as well

fixes: #2555

### DIFF
--- a/internal/css_parser/css_parser.go
+++ b/internal/css_parser/css_parser.go
@@ -832,11 +832,15 @@ abortRuleParser:
 
 		if p.peek(css_lexer.TIdent) {
 			name = p.decoded()
+			if name == "none" {
+				p.log.AddID(logger.MsgID_CSS_CSSSyntaxError, logger.Warning, &p.tracker, atRange, "Expected identifier but found `none` which is invalid in this context")
+				p.prevError = p.current().Range.Loc
+			}
 			p.advance()
 			isStringName = false
 		} else if p.peek(css_lexer.TString) {
 			name = p.decoded()
-			isStringName = !p.options.MinifySyntax || cssWideAndReservedKeywords[name]
+			isStringName = !p.options.MinifySyntax || cssWideAndReservedKeywords[name] || name == "none"
 			p.advance()
 		} else if !(p.expect(css_lexer.TIdent) && p.expect(css_lexer.TString)) && !p.peek(css_lexer.TOpenBrace) {
 			// Consider string names a syntax error even though they are allowed by
--- a/internal/css_parser/css_parser.go
+++ b/internal/css_parser/css_parser.go
@@ -828,11 +828,17 @@ abortRuleParser:
 	case "keyframes", "-webkit-keyframes", "-moz-keyframes", "-ms-keyframes", "-o-keyframes":
 		p.eat(css_lexer.TWhitespace)
 		var name string
+		var isStringName bool
 
 		if p.peek(css_lexer.TIdent) {
 			name = p.decoded()
 			p.advance()
-		} else if !p.expect(css_lexer.TIdent) && !p.eat(css_lexer.TString) && !p.peek(css_lexer.TOpenBrace) {
+			isStringName = false
+		} else if p.peek(css_lexer.TString) {
+			name = p.decoded()
+			isStringName = !p.options.MinifySyntax || cssWideAndReservedKeywords[name]
+			p.advance()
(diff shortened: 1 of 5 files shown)


internal/css_parser/css_parser_test.go
```diff
@@ -1013,6 +1013,23 @@ func TestLegalComment(t *testing.T) {
...
 	expectPrinted(t, "@keyframes name{}", "@keyframes name {\n}\n")
+	// string name valid to convert into ident
+	expectPrinted(t, `@keyframes "name"{}`, "@keyframes \"name\" {\n}\n")
+	expectPrintedMinify(t, `@keyframes "name"{}`, "@keyframes \"name\"{}")
+	expectPrintedMangleMinify(t, `@keyframes "name"{}`, "@keyframes name{}")
+
+	// string name valid to convert into ident but need escape
+	expectPrinted(t, `@keyframes "foo bar"{}`, "@keyframes \"foo bar\" {\n}\n")
+	expectPrintedMinify(t, `@keyframes "foo bar"{}`, "@keyframes \"foo bar\"{}")
+	expectPrintedMangleMinify(t, `@keyframes "foo bar"{}`, "@keyframes foo\\ bar{}")
```
yisibl, line 1024: Add test `@keyframes "none" {}`

yisibl: Because `none` is not included in the definition of `cssWideAndReservedKeywords`.

https://github.com/evanw/esbuild/blob/4a6f64cd81c996aa2868f467ae34e9af46405861/internal/css_parser/css_decls_font_family.go#L24-L34

internal/css_parser/css_parser.go
```diff
@@ -828,11 +828,21 @@ abortRuleParser:
...
+			isStringName = false
+		} else if p.peek(css_lexer.TString) {
+			name = strings.ToLower(p.decoded())
+			isStringName = !p.options.MinifySyntax || cssWideAndReservedKeywords[name] || name == "none"
 			p.advance()
-		} else if !p.expect(css_lexer.TIdent) && !p.eat(css_lexer.TString) && !p.peek(css_lexer.TOpenBrace) {
+		} else if !(p.expect(css_lexer.TIdent) && p.expect(css_lexer.TString)) && !p.peek(css_lexer.TOpenBrace) {
 			// Consider string names a syntax error even though they are allowed by
 			// the specification and they work in Firefox because they do not work in
 			// Chrome or Safari.
```
evanw, line 848: Not sure if you saw this comment or not but lack of support for this was deliberate. This is currently a Firefox-only feature, which I don't think makes sense to support natively in esbuild at this time.

I agree that the bug that changes the contents of `@keyframes` rules with unsupported syntax should still be fixed. These should be passed through unmodified. I will fix that part.

CGQAQ: I currently working on blink, and the webkit one is already merged

ref: https://github.com/WebKit/WebKit/pull/4234
      https://chromium-review.googlesource.com/c/chromium/src/+/3865188

CGQAQ:
I saw you submitted a [commit](https://github.com/evanw/esbuild/commit/c9b24d829cfb8a08d4ade8d7d78bee568f163270) to fix #2555 which makes it an unrecognized CSS rule,  I think it need to be reverted to land my PR land to conform the [CSS standard](https://www.w3.org/TR/css-animations-1/#typedef-keyframes-name
)

evanw:
> I think it need to be reverted to land my PR land to conform the [CSS standard](https://www.w3.org/TR/css-animations-1/#typedef-keyframes-name)

It should already confirm to the CSS standard after the latest release. The syntax should now be passed through unmodified, so you should now be able to use this syntax with esbuild's latest release. Do you have some CSS code for which that isn't true?

internal/css_parser/css_parser.go
yisibl: This is not a Firefox-only syntax, the prefixed `@-webkit-keyframes / @-moz-keyframes` always support `<string>`, and there are [UseCounter](https://codereview.chromium.org/651123003/) in Chrome that say: `The usage counter for strings in prefixed case is too high`.

evanw:
I believe the commit you linked to in your previous comment should already implement what this PR is trying to do. One difference is that the string is always converted to an identifier, but this is because Chrome has refused to implement support for string names (at least as far as I can tell). Leaving the string name as-is would result in broken code in all Chromium-based browsers.

PR #2621, opened 2022-10-20
feat: inline var in function body when possible

unbyte:
This PR inlines `var` declarations like how `let/const`s are processed.

It's to be applied only when:
1. the `var` declaration belongs to a function body, and
2. the function is an arrow function, or
3. the `arguments` is shadowed

This PR significantly reduced the bundle size of our internal large project, because our target is es5 and there are many code structures like `prop: function() { var x = value; return x; }`

### Input
```javascript
function test() {
  var x = 1;
  return x;
}
```
### Current Output
```javascript
function test() {
  var x = 1;
  return x;
}
```
### New Output
```javascript
function test() {
  return 1;
}
```


kenbern:
Any reason why this hasn't been implemented or merged?

Issue #1439, opened 2021-07-10
How to keep all my comments code?


anchengjian:
How to keep all my comments code, not just the LegalComments ???

evanw:
This is not supported. You should use another tool instead.

anchengjian:
why ?

heyheyhello:
@anchengjian It's not part of the AST (kinda; I mean legal comments and `__PURE__` comments are...). Evan explains why here https://github.com/evanw/esbuild/issues/221#issuecomment-653804184 

It's a bummer I know, but handling comments is hard logic to code correctly.

I'd like them to stay too. Not sure if this is your use case at all but: my team publishes npm packages that other internal teams use. It's common for people to ctrl+click an import to want to see the node_modules code. With webpack, we'd see close-to-sourcecode code that had comments. After moving to esbuild we lost that. Our workaround was to use `--banner` to put a comment linking to the GitLab source code so people can read the comments in their browser. It's not ideal but might work if you want to at least redirect people to commented code...

Issue #221, opened 2020-07-04
Add option to don't drop comments


timocov:
Related to #218 and #220.

Tools like `uglifyjs` or `terser` may rely on some comments like `/*@__PURE__*/` (see `side_effects` on https://github.com/terser/terser/blob/504b9677d2d6086025ead5afb713e151877944b2/README.md#compress-options).

In some projects you can't use TypeScript in `transpileOnly` mode (or don't want because you have already compiled code and just want to process JS files itself). The TypeScript leaves a comment like `/** @class */` right before class' IIFE, which could be replaced (somehow) with `/*@__PURE__*/` so `terser`/`uglifyjs` can understand it and do its magic with them (see https://github.com/mishoo/UglifyJS2/issues/2279).

It seems that `esbuild` drop all comments (licences as well) which does all comments-dependent processing unreal (dropping licences is the issue as well).

evanw:
The esbuild parser was designed for speed. It's not designed to be a code formatter. It will likely be extremely difficult to modify esbuild to ensure that all comments are preserved and in their original locations, especially as code can be moved around due to syntax transformations. I don't think esbuild is the right tool if you need to do this, and I am not planning on adding general comment preservation to esbuild. You will want to use other tools to do this instead.

I can see adding a special case for statement-level comments only. I believe I should be able to guarantee that those are preserved without rewriting the parser. That should be sufficient to cover license comments I think.

timocov:
> I can see adding a special case for statement-level comments only.

Is that comments "attached" to a nodes? I mean a comments right before a node  something like it's in TypeScript AST). If so, I think it might cover cases with "class" comment as well.

evanw:
> Is that comments "attached" to a nodes?

Attaching comments to nodes makes it easy for them to be accidentally lost. Instead I'm planning on creating a new kind of statement AST node for statement-level comments.

evanw:
Relevant to this issue: I'm currently adding limited support for `__PURE__` comments in esbuild (interpreting them, not passing them through). It's being done as part of some ongoing work for tree shaking. See https://github.com/evanw/esbuild/issues/86#issuecomment-625923444 for context.

evanw:
I just published a release with support for `__PURE__` comments. I also added pass-through printing of them when minification is disabled.

evanw:
I closed this because both `__PURE__` comments and license comments have been implemented. I believe that should be sufficient for your use case.

PR #2632, opened 2022-10-26
feat: support to preserve specific comments

unbyte:
According to #1439 and https://github.com/evanw/esbuild/issues/221#issuecomment-653804184, it seems that esbuild is not tend to supoort to preserve comments, but in our internal project we need to keep certain special placeholder comments for some tools to inject code, so I simply implement the basic feature we need and open this PR only to help others with the same needs get this patch easily.

To apply this patch:
```bash
$ curl -L https://github.com/evanw/esbuild/pull/2632.patch | git am
```

## Example
CLI
```bash
$ esbuild --outfile=output.js --preserve-comments=\[PositionFor input.js
```

Input
```javascript
/* [PositionForAppBegin] */
App({})
/* [PositionForAppEnd] */
/* should be removed */
```

Output
```javascript
/* [PositionForAppBegin] */
App({});
/* [PositionForAppEnd] */
```


benmccann:
Could this also be used to preserve all comments by passing something like `.*`? Storybook has a need to preserve comments as they're used for automatically generating documentation within Storybook. In the Svelte plugin, we need to first run any TypeScript through a compiler to convert to JavaScript so that the Svelte compiler can understand it. The fact that esbuild is removing comments is causing us major headaches (https://github.com/storybookjs/storybook/pull/24096)

Issue #2628, opened 2022-10-24
`target` in tsconfig doesn't affect esbuild's runtime library


mengxinssfd:
I found that after building the code, there is an extra `__spreadValues` function, which is written inside `var prop in b ||= {}`

```javascript
var __spreadValues = (a, b) => {
  for (var prop in b ||= {})
    if (__hasOwnProp.call(b, prop))
      __defNormalProp(a, prop, b[prop]);
  if (__getOwnPropSymbols)
    for (var prop of __getOwnPropSymbols(b)) {
      if (__propIsEnum.call(b, prop))
        __defNormalProp(a, prop, b[prop]);
    }
  return a;
};
```
It's using the `ECMAScript 2021 - (ES12)` feature, when I set the `target` to a version before ES12, it doesn't work

mengxinssfd:
> Can you provide a self-contained code sample and esbuild configuration that will reproduce the problem?


You can look at this repository <https://github.com/mengxinssfd/esbuild-isuue--2628>.

Sorry, what I said earlier was not correct. `esbuild.build({target:'ES 2017'})` is useful.

It is accurate to use the `target` inside `tsconfig.json` without specifying `target:'ES version'`.

heyheyhello:
@evanw I ran into this just now, the issue is that `||=` is used despite the tsconfig target asking for ES6. Here's a repro:

```
$ echo 'var a = { a:1 }; export default { b:2, ...a }' | npx esbuild --loader=ts --tsconfig-raw='{"compilerOptions": { "target": "es6" }}'

var __defProp = Object.defineProperty;
var __getOwnPropSymbols = Object.getOwnPropertySymbols;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __propIsEnum = Object.prototype.propertyIsEnumerable;
var __defNormalProp = (obj, key, value) => key in obj ? __defProp(obj, key, { enumerable: true, configurable: true, writable: true, value }) : obj[key] = value;
var __spreadValues = (a2, b) => {
  for (var prop in b ||= {})
    if (__hasOwnProp.call(b, prop))
      __defNormalProp(a2, prop, b[prop]);
  if (__getOwnPropSymbols)
    for (var prop of __getOwnPropSymbols(b)) {
      if (__propIsEnum.call(b, prop))
        __defNormalProp(a2, prop, b[prop]);
    }
  return a2;
};
var a = { a: 1 };
export default __spreadValues({ b: 2 }, a);
```

[...]

PR #2687, opened 2022-11-21
Consider TypeScript's target in `parseRuntime()`

heyheyhello:
Closes https://github.com/evanw/esbuild/issues/2628

Currently esbuild emits `||=` even when TS target is set to ES6:

```
$ echo 'var a = { a:1 }; window.xyz ||= a; export default { b:2, ...a }' | ./esbuild --loader=ts --tsconfig-raw='{"compilerOptions": { "target": "es6" }}'
[...]
var __spreadValues = (a2, b) => {
  for (var prop in b ||= {})
    [...]
  return a2;
};
var a = { a: 1 };
window.xyz || (window.xyz = a);
export default __spreadValues({ b: 2 }, a);
```

This PR fixes the output to lower `||=` into `b || (b = {})`:

```
$ echo 'var a = { a:1 }; window.xyz ||= a; export default { b:2, ...a }' | ./esbuild --loader=ts --tsconfig-raw='{"compilerOptions": { "target": "es6" }}'
[...]
var __spreadValues = (a2, b) => {
  for (var prop in b || (b = {}))
   [...]
  return a2;
};
var a = { a: 1 };
window.xyz || (window.xyz = a);
export default __spreadValues({ b: 2 }, a);
```

Disclaimer: I've never written Go and I debugged this with a dozen `fmt.Println()` calls. Sorry I'm not up for writing tests. Also see PR comment about the cache key.

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -2487,6 +2487,9 @@ type runtimeCacheKey struct {
 	unsupportedJSFeatures compat.JSFeature
 	minifySyntax          bool
 	minifyIdentifiers     bool
+	// TODO: Safe to use a pointer here? Is the target itself easier/safer?
+	// tsTargetValue	  *config.TSTarget.value?
+	tsTarget			  *config.TSTarget
 }
 
 type runtimeCache struct {
@@ -2502,6 +2505,7 @@ func (cache *runtimeCache) parseRuntime(options *config.Options) (source logger.
 		unsupportedJSFeatures: options.UnsupportedJSFeatures,
 		minifySyntax:          options.MinifySyntax,
 		minifyIdentifiers:     options.MinifyIdentifiers,
+		tsTarget:              options.TSTarget,
 	}
 
 	// Determine which source to use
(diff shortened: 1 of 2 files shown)


internal/bundler/bundler.go
```diff
@@ -2487,6 +2487,9 @@ type runtimeCacheKey struct {
 	unsupportedJSFeatures compat.JSFeature
 	minifySyntax          bool
 	minifyIdentifiers     bool
+	// TODO: Safe to use a pointer here? Is the target itself easier/safer?
+	// tsTargetValue	  *config.TSTarget.value?
+	tsTarget			  *config.TSTarget
```
heyheyhello, line 2492: Please review/edit this line - I'll defer to you on whether an ES version number like `2015` is more appropriate than a pointer.

heyheyhello:
@evanw Is this ok to be merged? My company is patching esbuild to workaround it but would prefer official upstream release. Thank you! 👏

evanw:
I’m closing this PR as #2628 has been closed. The `target` field of `tsconfig.json` no longer affects esbuild’s syntax transforms.

PR #2816, opened 2023-01-12
Version 0.17.0: rewrite the incremental build apis

### LINKED ISSUE #621: Can't bundle html file along using file loader
I want esbuild to copy my index.html file along with the js bundle to the `public` folder so that I can use it with the `serve` api. I'm getting this message:

```console
$ esbuild 'src/index.tsx' --bundle --loader:.html=file --loader:.woff2=file --outdir=public
 > src/index.tsx: warning: Ignoring this import because "src/index.html" was marked as having no side effects
    1 │ import './index.html'
      ╵        ~~~~~~~~~~~~~~
```

What am I doing wrong?

### LINKED ISSUE #802: Add Live Reload to the esbuild serve
I love the `servedir=` option added in latest [not yet released] release. This makes development lifecycle much easier without needing to start multiple servers.

I also consider live reload functionality to be most essential to boost development lifecycle. A live reload server basically reloads the entire web page when src code changes. The underlying communication mechanism can be either SSE or websockets based.

Would you be open to add Live Reload to esbuild serve mode?

PS: I've been doing this outside esbuild with https://github.com/osdevisnot/sorvor  and I'm open to try contributing this feature to esbuild.

### LINKED ISSUE #805: Cannot use "watch" with "serve"
Hi, I am trying to replace `live-server` for my development with the new `--servedir=` but having problems getting it to watch the files and serve them. Currently, I have:
```
#!/bin/bash

npx esbuild --watch --bundle --minify --sourcemap src/mod.ts --outfile=dist/bundle.js &
npx live-server ./dist &
wait
```
As soon as I save then the browser reloads the file. I see https://github.com/evanw/esbuild/issues/802 , but I don't see the use for `--servedir=` unless it refreshes when the files changes somehow? 

Can I now replace `live-server` with esbuild or am I confused about the new feature? Cheers.

### LINKED ISSUE #980: Initial notification to watch.onRebuild
I added the `watch.onRebuild` callback to a build-script, and it immediately seemed like the script had just stalled.

At closer examination, it does in fact build - it's just that there's no notification to this callback for the first build.

Can we trigger a notification during the first build on start-up please? (Most build-tools do, afaik.)

Alternatively, if you think there's a real use-case where someone really only wants a notification for all builds except the first one, perhaps consider adding a second, identical callback, e.g. `onBuild`, which would get called for every build.

(My use-case is I'd like to automatically run my tests after each incremental build - and so, I'd expect the tests to run at startup, without having to first make an arbitrary change to a file to trigger it.)

### LINKED ISSUE #1037: Watch mode behavior on failed first builds
I've noticed that watch mode only works if the first build is successful. Is this intentional?

My use case is a large repository with many apps and a single development server that lazily builds these apps in watch mode as they are requested. The server also establishes a web socket connection so the apps reload on successful rebuilds or render an error overlay if the build failed. Ideally the server could rely on esbuild's watch mode completely to know when to send these web socket messages, but currently I would have to set up my own watcher and retry starting esbuild in watch mode in the case where the build fails on the first try.

### LINKED ISSUE #1384: Plugins' onEnd callback isn't triggerred in serve mode
I have 
```
      build.onStart(() => {
        console.log('build end')
      })
      build.onStart(() => {
        console.log('build started')
      })
```
in `build` and everything works as expected. Both `build started` and `build end` are printed properly.

However,if I use `serve` with
```
  esbuildServe(serveOptions, esbuildOptions).then(server => {
    console.log(green(`hosted on http://${server.host}:${server.port}`));
    registerCleanup(() => server.stop());
  }).catch(e => {
    console.error(e);
    process.exit(1);
  })
```
I got `hosted on http://0.0.0.0:8000` and multiple `build started` only. I am suspecting the onEnd callback isn't called properly called in serve mode?

### LINKED ISSUE #1606: Incremental mode on errors
The `rebuild` function is only provided when esbuild manages to finish a build. It would be nice if it could instead always return it (even on errors), so that we can decrease the feedback loop when fixing the build.

I tried doing it [myself](https://github.com/evanw/esbuild/compare/master...arcanis:patch-1) but although it almost works it crashes with the following right after it finishes the rebuild:

```
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [semacquire]:
sync.runtime_Semacquire(0xc00010cd58)
        runtime/sema.go:56 +0x45
sync.(*WaitGroup).Wait(0xc00010cd50)
        sync/waitgroup.go:130 +0x65
main.runService(0x7ffeefbff401)
        github.com/evanw/esbuild/cmd/esbuild/service.go:138 +0x636
main.main()
        github.com/evanw/esbuild/cmd/esbuild/main.go:200 +0x3ae

goroutine 18 [chan receive]:
main.runService.func1(0xc00012e140, 0xc00010cd50)
        github.com/evanw/esbuild/cmd/esbuild/service.go:66 +0x4a
created by main.runService
        github.com/evanw/esbuild/cmd/esbuild/service.go:64 +0x1fc

goroutine 19 [chan receive]:
main.(*serviceType).sendRequest(0xc00012e140, 0x148ca60, 0xc066251980, 0x7, 0xc066278088)
        github.com/evanw/esbuild/cmd/esbuild/service.go:163 +0x11e
main.runService.func2(0xc00012e140)
        github.com/evanw/esbuild/cmd/esbuild/service.go:92 +0x47
created by main.runService
        github.com/evanw/esbuild/cmd/esbuild/service.go:89 +0x6ae
```

### LINKED ISSUE #1650: cant use watch with servedir error clarification, and node config.js poorly placed in documentation
The better error message there when using watch with servedir flags is that servedir flag is only necessary. I'd also like to point out that I don't think the docs actually mention anywhere near the intro sections that you're suppose to use esbuild with node if using config file which servedir I think requires, therefore you have one of the more common use cases by in large undocumented in getting started.

### LINKED ISSUE #1771: Move index.html from src to dist folder
I am learning esbuild and want to test with a simple project. I am struck on moving the src/index.html to dist/index.html

Is there any loader available that runs or moves only once, and watches for changes in html files.

[https://github.com/alaksandarjesus/esbuilder] (github link to project setup)

### LINKED ISSUE #2104: Watch mode rebuilds cause touch/update to every output file
Hi. Is there a way to configure Esbuild in watch mode to only touch/rewrite the bundles (output files) that actually changed with the rebuild.

I'd even be willing to sacrifice a bit of speed.

Use case: I'm running a separate [watch task](https://www.npmjs.com/package/onchange) to monitor the contents of the outdir, and perform further actions on the files that changed.

The fact that every output file fires a change-event on every incremental rebuild is making things very busy+slow.

I've done my best to hunt for answers, but to no avail. Any suggestions?

### LINKED ISSUE #2169: Minor suggestion for Serve API docs
Looking at the Serve docs explaining a proxy, it would be helpful for learners if there were a couple of links to the Node docs. Not being very experienced with Node, on seeing the example my first reaction was to look for `http` on npm (very wrong). Then Node's documentation for `http` didn't make it clear that that HTTPS requires using `https` instead - I did find it, but only after a bit of poking around.

My suggestions:

Old: "Here's a simple example of a proxy server to get you started."
New: "Here's a simple example of a proxy server to get you started, using Node's [`http`](https://nodejs.org/docs/latest-v8.x/api/http.html)."

Old: "Adding support for HTTPS using your own self-signed certificates"
New: "Adding support for HTTPS using Node's [`https`](https://nodejs.org/docs/latest-v8.x/api/https.html) and your own self-signed certificates"

### LINKED ISSUE #2280: an error occurs when first build with incremental is true, and can't get the build instance
When I use build method with config `incremental: true`，such as：

```js
const esbuild = require('esbuild');
const path = require('path');

async function main() {
  try {
    const instance = await esbuild.build({
      incremental: true,
      bundle: true,
      minify: false,
      watch: true,
      entryPoints: [path.join(__dirname, './index.ts')],
    });

    console.log(`重编译实例：${!!instance}`);
  } catch (e) {
    console.log(e);
  }
}

main();
```

If the first build occurs a error, for example, the code in `index.ts`：

```ts
let let
```

I can't get result of `esbuild.build`, it will be into `catch` statement.

I think do not throw errors here and set errors to `errors` property of result, it can be better.

### LINKED ISSUE #2393: `esbuild.serve` does not emit any URLs / ports, even with `logLevel: 'info'`
The ClI works

```js
❯ npx esbuild --serve --bundle index.js --outdir=a

 > Local:   http://127.0.0.1:8000/
 > Network: http://10.9.8.45:8000/
```

The API doesn't output anything:

```sh
❯ WATCH=true node esbuild.config.mjs
// nothing
```
```js
//esbuild.config.mjs
import process from 'node:process';
import esbuild from 'esbuild';

const serve = {
	servedir: 'public',
};
const build = {
	entryPoints: ['index.js'],
	outdir: 'public',
	logLevel: 'info',
};

if (process.env.WATCH === 'true') {
	await esbuild.serve(serve, build);
} else {
	await esbuild.build(build);
}
```

Any config to make it work? Or is this functionality missing?

### LINKED ISSUE #2499: WatchMode.onRebuild not called if initial build failed
If you have multiple esbuild watchers running and one of them fails, that watcher will stop calling onRebuild even though the watcher is still running and rebuilding that file

Build Script
```js
import esbuild from "esbuild";

const watch = {
    onRebuild: (err) => {
        if (err) console.error("Build Error", err.message);
        else console.log("Rebuilt!");
    }
};

await Promise.all([
    esbuild.build({
        entryPoints: ["index.js"],
        outfile: "out.js",
        watch
    }), esbuild.build({
        entryPoints: ["other.js"],
        outfile: "otherOut.js",
        watch
    })
]).catch(() => 0);
```

if index.js has an error initially and you then change index.js to fix the error and add something new, esbuild will rebuild it but not print the "Rebuilt". It works as expected if the initial build succeeds

https://user-images.githubusercontent.com/45497981/187244627-98664e4a-fee6-41dc-8101-f001ccf5dbfe.mp4

### LINKED ISSUE #2576: consider separating the "rebuild" and "dirty monitoring" concepts
Hello, thanks for this great tool!

I'm running into something similar to what people are discussing in https://github.com/evanw/esbuild/issues/805 — trying to get some of the functionality from `serve` and `watch` at the same time — specifically being able to build in memory, have it served on a throwaway port, but also be notified when input files change. (I'm doing this inside an Electron app built for creative coding / generative art.)

I understand that `watch` and `serve` aren't compatible, because they rebuild at different times, so enabling both would end up with duplicate work. But I think there might be a slight tweak to the abstraction, which allows for the idea of "monitoring dirty files" without being tightly coupled to rebuilding (as it is in `watch` mode).

For example (in pseudocode) if there was an `onDirty` callback, you could actually define watch mode yourself as:

```ts
let result = await build({
  incremental: true,
  onDirty() {
    result.rebuild()
  }
})
```

But this would also allow for using `serve` and monitoring files, but without the double rebuilding from `watch`:

```ts
let result  = await serve({
  onDirty() {
    // invoke whatever you want. in my case it's some logic to
    // notify existing canvases that they should reload the sketch
  }
})
```

What do you think? I might be missing something, but being able to tap into Esbuild's nice and simple logic for knowing when input files change would be really valuable even when using `serve`.

### LINKED ISSUE #2625: Inconsistent logging for plugin errors (result.errors.push vs throw)
Repro:
```js
build({
  entryPoints: ["./file.ts"],
  write: false,
  plugins: [
    {
      name: "test",
      setup(b) {
        b.onEnd((result) => {
          // throw new Error("Error description"); // <-- This get a nice log
          // This is not logged (but correctly fails the build)
          result.errors.push({
            pluginName: "test",
            id: "some-id",
            text: "Error description",
            detail: null,
            notes: [],
            location: null,
          });
        });
      },
    },
  ],
});
```

evanw:
At a high level, the breaking changes in this PR fix some long-standing issues with the design of esbuild's incremental, watch, and serve APIs. This PR also introduces some exciting new features such as live reloading. In detail:

* Move everything related to incremental builds to a new `context` API ([#1037](https://github.com/evanw/esbuild/issues/1037), [#1606](https://github.com/evanw/esbuild/issues/1606), [#2280](https://github.com/evanw/esbuild/issues/2280), [#2418](https://github.com/evanw/esbuild/issues/2418))

    This change removes the `incremental` and `watch` options as well as the `serve()` method, and introduces a new `context()` method. The context method takes the same arguments as the `build()` method but only validates its arguments and does not do an initial build. Instead, builds can be triggered using the `rebuild()`, `watch()`, and `serve()` methods on the returned context object. The new context API looks like this:

    ```js
    // Create a context for incremental builds
    const context = await esbuild.context({
      entryPoints: ['app.ts'],
      bundle: true,
    })

    // Manually do an incremental build
    const result = await context.rebuild()

    // Enable watch mode
    await context.watch()

    // Enable serve mode
    await context.serve()

    // Dispose of the context
    context.dispose()
    ```

    The switch to the context API solves a major issue with the previous API which is that if the initial build fails, a promise is thrown in JavaScript which prevents you from accessing the returned result object. That prevented you from setting up long-running operations such as watch mode when the initial build contained errors. It also makes tearing down incremental builds simpler as there is now a single way to do it instead of three separate ways.

    In addition, this PR also makes some subtle changes to how incremental builds work. Previously every call to `rebuild()` started a new build. If you weren't careful, then builds could actually overlap. This doesn't cause any problems with esbuild itself, but could potentially cause problems with plugins (esbuild doesn't even give you a way to identify which overlapping build a given plugin callback is running on). Overlapping builds also arguably aren't useful, or at least aren't useful enough to justify the confusion and complexity that they bring. With this PR, there is now only ever a single active build per context. Calling `rebuild()` before the previous rebuild has finished now "merges" with the existing rebuild instead of starting a new build.

* Allow using `watch` and `serve` together ([#805](https://github.com/evanw/esbuild/issues/805), [#1650](https://github.com/evanw/esbuild/issues/1650), [#2576](https://github.com/evanw/esbuild/issues/2576))

    Previously it was not possible to use watch mode and serve mode together. The rationale was that watch mode is one way of automatically rebuilding your project and serve mode is another (since serve mode automatically rebuilds on every request). However, people want to combine these two features to make "live reloading" where the browser automatically reloads the page when files are changed on the file system.

    This PR now allows you to use these two features together. You can only call the `watch()` and `serve()` APIs once each per context, but if you call them together on the same context then esbuild will automatically rebuild both when files on the file system are changed *and* when the server serves a request.

* Support "live reloading" through server-sent events ([#802](https://github.com/evanw/esbuild/issues/802))

    [Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) are a simple way to pass one-directional messages asynchronously from the server to the client. Serve mode now provides a `/esbuild` endpoint with an `change` event that triggers every time esbuild's output changes. So you can now implement simple "live reloading" (i.e. reloading the page when a file is edited and saved) like this:

    ```js
    new EventSource('/esbuild').addEventListener('change', () => location.reload())
    ```

    The event payload is a JSON object with the following shape:

    ```ts
    interface ChangeEvent {
      added: string[]
      removed: string[]
      updated: string[]
    }
    ```

    This JSON should also enable more complex live reloading scenarios. For example, the following code hot-swaps changed CSS `<link>` tags in place without reloading the page (but still reloads when there are other types of changes):

    ```js
    new EventSource('/esbuild').addEventListener('change', e => {
      const { added, removed, updated } = JSON.parse(e.data)
      if (!added.length && !removed.length && updated.length === 1) {
        for (const link of document.getElementsByTagName("link")) {
          const url = new URL(link.href)
          if (url.host === location.host && url.pathname === updated[0]) {
            const next = link.cloneNode()
            next.href = updated[0] + '?' + Math.random().toString(36).slice(2)
            next.onload = () => link.remove()
            link.parentNode.insertBefore(next, link.nextSibling)
            return
          }
        }
      }
      location.reload()
    })
    ```

    Implementing live reloading like this has a few known caveats:

    * These events only trigger when esbuild's output changes. They do not trigger when files unrelated to the build being watched are changed. If your HTML file references other files that esbuild doesn't know about and those files are changed, you can either manually reload the page or you can implement your own live reloading infrastructure instead of using esbuild's built-in behavior.

    * The `EventSource` API is supposed to automatically reconnect for you. However, there's a bug in Firefox that breaks this if the server is ever temporarily unreachable: https://bugzilla.mozilla.org/show_bug.cgi?id=1809332. Workarounds are to use any other browser, to manually reload the page if this happens, or to write more complicated code that manually closes and re-creates the `EventSource` object if there is a connection error. I'm hopeful that this bug will be fixed.

    * Browser vendors have decided to not implement HTTP/2 without TLS. This means that each `/esbuild` event source will take up one of your precious 6 simultaneous per-domain HTTP/1.1 connections. So if you open more than six HTTP tabs that use this live-reloading technique, you will be unable to use live reloading in some of those tabs (and other things will likely also break). The workaround is to enable HTTPS, which is now possible to do in esbuild itself (see below).

* Add built-in support for HTTPS ([#2169](https://github.com/evanw/esbuild/issues/2169))

    You can now tell esbuild's built-in development server to use HTTPS instead of HTTP. This is sometimes necessary because browser vendors have started making modern web features unavailable to HTTP websites. Previously you had to put a proxy in front of esbuild to enable HTTPS since esbuild's development server only supported HTTP. But with this release, you can now enable HTTPS with esbuild without an additional proxy.

    To enable HTTPS with esbuild:

    1. Generate a self-signed certificate. There are many ways to do this. Here's one way, assuming you have `openssl` installed:

        ```
        openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 9999 -nodes -subj /CN=127.0.0.1
        ```

    2. Add `--keyfile=key.pem` and `--certfile=cert.pem` to your esbuild development server command
    3. Click past the scary warning in your browser when you load your page

    If you have more complex needs than this, you can still put a proxy in front of esbuild and use that for HTTPS instead. Note that if you see the message "Client sent an HTTP request to an HTTPS server" when you load your page, then you are using the incorrect protocol. Replace `http://` with `https://` in your browser's URL bar.

    Keep in mind that esbuild's HTTPS support has nothing to do with security. The only reason esbuild now supports HTTPS is because browsers have made it impossible to do local development with certain modern web features without jumping through these extra hoops. *Please do not use esbuild's development server for anything that needs to be secure.* It's only intended for local development and no considerations have been made for production environments whatsoever.

* Better support copying `index.html` into the output directory ([#621](https://github.com/evanw/esbuild/issues/621), [#1771](https://github.com/evanw/esbuild/issues/1771))

    Right now esbuild only supports JavaScript and CSS as first-class content types. Previously this meant that if you were building a website with a HTML file, a JavaScript file, and a CSS file, you could use esbuild to build the JavaScript file and the CSS file into the output directory but not to copy the HTML file into the output directory. You needed a separate `cp` command for that.

    Or so I thought. It turns out that the `copy` loader added in version 0.14.44 of esbuild is sufficient to have esbuild copy the HTML file into the output directory as well. You can add something like `index.html --loader:.html=copy` and esbuild will copy `index.html` into the output directory for you. The benefits of this are a) you don't need a separate `cp` command and b) the `index.html` file will automatically be re-copied when esbuild is in watch mode and the contents of `index.html` are edited. This also goes for other non-HTML file types that you might want to copy.

    This pretty much already worked. The one thing that didn't work was that esbuild's built-in development server previously only supported implicitly loading `index.html` (e.g. loading `/about/index.html` when you visit `/about/`) when `index.html` existed on the file system. Previously esbuild didn't support implicitly loading `index.html` if it was a build result. That bug has been fixed with this PR so it should now be practical to use the `copy` loader to do this.

* Fix `onEnd` not being called in serve mode ([#1384](https://github.com/evanw/esbuild/issues/1384))

    Previous releases had a bug where plugin `onEnd` callbacks weren't called when using the top-level `serve()` API. This API no longer exists and the internals have been reimplemented such that `onEnd` callbacks should now always be called at the end of every build.

* Incremental builds now write out build results differently ([#2104](https://github.com/evanw/esbuild/issues/2104))

    Previously build results were always written out after every build. However, this could cause the output directory to fill up with files from old builds if code splitting was enabled, since the file names for code splitting chunks contain content hashes and old files were not deleted.

    With this PR, incremental builds in esbuild will now delete old output files from previous builds that are no longer relevant. Subsequent incremental builds will also no longer overwrite output files whose contents haven't changed since the previous incremental build.

* The `onRebuild` watch mode callback was removed ([#980](https://github.com/evanw/esbuild/issues/980), [#2499](https://github.com/evanw/esbuild/issues/2499))

    Previously watch mode accepted an `onRebuild` callback which was called whenever watch mode rebuilt something. This was not great in practice because if you are running code after a build, you likely want that code to run after every build, not just after the second and subsequent builds. This PR removes option to provide an `onRebuild` callback. You can create a plugin with an `onEnd` callback instead. The `onEnd` plugin API already exists, and is a way to run some code after every build.

* You can now return errors from `onEnd` ([#2625](https://github.com/evanw/esbuild/issues/2625))

    It's now possible to add additional build errors and/or warnings to the current build from within your `onEnd` callback by returning them in an array. This is identical to how the `onStart` callback already works. The evaluation of `onEnd` callbacks have been moved around a bit internally to make this possible.

    Note that the build will only fail (i.e. reject the promise) if the additional errors are returned from `onEnd`. Adding additional errors to the result object that's passed to `onEnd` won't affect esbuild's behavior at all.

* Print URLs and ports from the Go and JS APIs ([#2393](https://github.com/evanw/esbuild/issues/2393))

    Previously esbuild's CLI printed out something like this when serve mode is active:

    ```
     > Local:   http://127.0.0.1:8000/
     > Network: http://192.168.0.1:8000/
    ```

    The CLI still does this, but now the JS and Go serve mode APIs will do this too. This only happens when the log level is set to `verbose`, `debug`, or `info` but not when it's set to `warning`, `error`, or `silent`.

### Upgrade guide for existing code:

* Rebuild (a.k.a. incremental build):

    Before:

    ```js
    const result = await esbuild.build({ ...buildOptions, incremental: true });
    builds.push(result);
    for (let i = 0; i < 4; i++) builds.push(await result.rebuild());
    await result.rebuild.dispose(); // To free resources
    ```

    After:

    ```js
    const ctx = await esbuild.context(buildOptions);
    for (let i = 0; i < 5; i++) builds.push(await ctx.rebuild());
    await ctx.dispose(); // To free resources
    ```

    Previously the first build was done differently than subsequent builds. Now both the first build and subsequent builds are done using the same API.

* Serve:

    Before:

    ```js
    const serveResult = await esbuild.serve(serveOptions, buildOptions);
    ...
    serveResult.stop(); await serveResult.wait; // To free resources
    ```

    After:

    ```js
    const ctx = await esbuild.context(buildOptions);
    const serveResult = await ctx.serve(serveOptions);
    ...
    await ctx.dispose(); // To free resources
    ```

* Watch:

    Before:

    ```js
    const result = await esbuild.build({ ...buildOptions, watch: true });
    ...
    result.stop(); // To free resources
    ```

    After:

    ```js
    const ctx = await esbuild.context(buildOptions);
    await ctx.watch();
    ...
    await ctx.dispose(); // To free resources
    ```

* Watch with `onRebuild`:

    Before:

    ```js
    const onRebuild = (error, result) => {
      if (error) console.log('subsequent build:', error);
      else console.log('subsequent build:', result);
    };
    try {
      const result = await esbuild.build({ ...buildOptions, watch: { onRebuild } });
      console.log('first build:', result);
      ...
      result.stop(); // To free resources
    } catch (error) {
      console.log('first build:', error);
    }
    ```

    After:

    ```js
    const plugins = [{
      name: 'my-plugin',
      setup(build) {
        let count = 0;
        build.onEnd(result => {
          if (count++ === 0) console.log('first build:', result);
          else console.log('subsequent build:', result);
        });
      },
    }];
    const ctx = await esbuild.context({ ...buildOptions, plugins });
    await ctx.watch();
    ...
    await ctx.dispose(); // To free resources
    ```

    The `onRebuild` function has now been removed. The replacement is to make a plugin with an `onEnd` callback.

    Previously `onRebuild` did not fire for the first build (only for subsequent builds). This was usually problematic, so using `onEnd` instead of `onRebuild` is likely less error-prone. But if you need to emulate the old behavior of `onRebuild` that ignores the first build, then you'll need to manually count and ignore the first build in your plugin (as demonstrated above).

Notice how all of these API calls are now done off the new context object. You should now be able to use all three kinds of incremental builds (`rebuild`, `serve`, and `watch`) together on the same context object. Also notice how calling `dispose` on the context is now the common way to discard the context and free resources in all of these situations.

Fixes #621
Fixes #802
Fixes #805
Fixes #980
Fixes #1037
Fixes #1384
Fixes #1606
Fixes #1650
Fixes #1771
Fixes #2104
Fixes #2169
Fixes #2280
Fixes #2393
Fixes #2418
Fixes #2499
Fixes #2576
Fixes #2625

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,255 @@
 # Changelog
 
+## Unreleased
+
+**This release deliberately contains backwards-incompatible changes.** To avoid automatically picking up releases like this, you should either be pinning the exact version of `esbuild` in your `package.json` file (recommended) or be using a version range syntax that only accepts patch upgrades such as `^0.16.0` or `~0.16.0`. See npm's documentation about [semver](https://docs.npmjs.com/cli/v6/using-npm/semver/) for more information.
+
+At a high level, the breaking changes in this release fix some long-standing issues with the design of esbuild's incremental, watch, and serve APIs. This release also introduces some exciting new features such as live reloading. In detail:
+
+* Move everything related to incremental builds to a new `context` API ([#1037](https://github.com/evanw/esbuild/issues/1037), [#1606](https://github.com/evanw/esbuild/issues/1606), [#2280](https://github.com/evanw/esbuild/issues/2280), [#2418](https://github.com/evanw/esbuild/issues/2418))
+
+    This change removes the `incremental` and `watch` options as well as the `serve()` method, and introduces a new `context()` method. The context method takes the same arguments as the `build()` method but only validates its arguments and does not do an initial build. Instead, builds can be triggered using the `rebuild()`, `watch()`, and `serve()` methods on the returned context object. The new context API looks like this:
+
+    ```js
+    // Create a context for incremental builds
+    const context = await esbuild.context({
+      entryPoints: ['app.ts'],
+      bundle: true,
+    })
+
+    // Manually do an incremental build
+    const result = await context.rebuild()
+
+    // Enable watch mode
+    await context.watch()
+
+    // Enable serve mode
+    await context.serve()
+
+    // Dispose of the context
+    context.dispose()
+    ```
+
+    The switch to the context API solves a major issue with the previous API which is that if the initial build fails, a promise is thrown in JavaScript which prevents you from accessing the returned result object. That prevented you from setting up long-running operations such as watch mode when the initial build contained errors. It also makes tearing down incremental builds simpler as there is now a single way to do it instead of three separate ways.
+
+    In addition, this release also makes some subtle changes to how incremental builds work. Previously every call to `rebuild()` started a new build. If you weren't careful, then builds could actually overlap. This doesn't cause any problems with esbuild itself, but could potentially cause problems with plugins (esbuild doesn't even give you a way to identify which overlapping build a given plugin callback is running on). Overlapping builds also arguably aren't useful, or at least aren't useful enough to justify the confusion and complexity that they bring. With this release, there is now only ever a single active build per context. Calling `rebuild()` before the previous rebuild has finished now "merges" with the existing rebuild instead of starting a new build.
+
+* Allow using `watch` and `serve` together ([#805](https://github.com/evanw/esbuild/issues/805), [#1650](https://github.com/evanw/esbuild/issues/1650), [#2576](https://github.com/evanw/esbuild/issues/2576))
+
+    Previously it was not possible to use watch mode and serve mode together. The rationale was that watch mode is one way of automatically rebuilding your project and serve mode is another (since serve mode automatically rebuilds on every request). However, people want to combine these two features to make "live reloading" where the browser automatically reloads the page when files are changed on the file system.
+
+    This release now allows you to use these two features together. You can only call the `watch()` and `serve()` APIs once each per context, but if you call them together on the same context then esbuild will automatically rebuild both when files on the file system are changed *and* when the server serves a request.
+
+* Support "live reloading" through server-sent events ([#802](https://github.com/evanw/esbuild/issues/802))
+
+    [Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) are a simple way to pass one-directional messages asynchronously from the server to the client. Serve mode now provides a `/esbuild` endpoint with an `change` event that triggers every time esbuild's output changes. So you can now implement simple "live reloading" (i.e. reloading the page when a file is edited and saved) like this:
+
+    ```js
+    new EventSource('/esbuild').addEventListener('change', () => location.reload())
+    ```
+
+    The event payload is a JSON object with the following shape:
+
+    ```ts
+    interface ChangeEvent {
+      added: string[]
+      removed: string[]
+      updated: string[]
+    }
+    ```
+
+    This JSON should also enable more complex live reloading scenarios. For example, the following code hot-swaps changed CSS `<link>` tags in place without reloading the page (but still reloads when there are other types of changes):
+
+    ```js
+    new EventSource('/esbuild').addEventListener('change', e => {
+      const { added, removed, updated } = JSON.parse(e.data)
+      if (!added.length && !removed.length && updated.length === 1) {
+        for (const link of document.getElementsByTagName("link")) {
+          const url = new URL(link.href)
+          if (url.host === location.host && url.pathname === updated[0]) {
+            const next = link.cloneNode()
+            next.href = updated[0] + '?' + Math.random().toString(36).slice(2)
+            next.onload = () => link.remove()
+            link.parentNode.insertBefore(next, link.nextSibling)
+            return
+          }
+        }
+      }
+      location.reload()
+    })
+    ```
+
+    Implementing live reloading like this has a few known caveats:
+
+    * These events only trigger when esbuild's output changes. They do not trigger when files unrelated to the build being watched are changed. If your HTML file references other files that esbuild doesn't know about and those files are changed, you can either manually reload the page or you can implement your own live reloading infrastructure instead of using esbuild's built-in behavior.
+
+    * The `EventSource` API is supposed to automatically reconnect for you. However, there's a bug in Firefox that breaks this if the server is ever temporarily unreachable: https://bugzilla.mozilla.org/show_bug.cgi?id=1809332. Workarounds are to use any other browser, to manually reload the page if this happens, or to write more complicated code that manually closes and re-creates the `EventSource` object if there is a connection error. I'm hopeful that this bug will be fixed.
+
+    * Browser vendors have decided to not implement HTTP/2 without TLS. This means that each `/esbuild` event source will take up one of your precious 6 simultaneous per-domain HTTP/1.1 connections. So if you open more than six HTTP tabs that use this live-reloading technique, you will be unable to use live reloading in some of those tabs (and other things will likely also break). The workaround is to enable HTTPS, which is now possible to do in esbuild itself (see below).
+
+* Add built-in support for HTTPS ([#2169](https://github.com/evanw/esbuild/issues/2169))
+
+    You can now tell esbuild's built-in development server to use HTTPS instead of HTTP. This is sometimes necessary because browser vendors have started making modern web features unavailable to HTTP websites. Previously you had to put a proxy in front of esbuild to enable HTTPS since esbuild's development server only supported HTTP. But with this release, you can now enable HTTPS with esbuild without an additional proxy.
+
+    To enable HTTPS with esbuild:
+
+    1. Generate a self-signed certificate. There are many ways to do this. Here's one way, assuming you have `openssl` installed:
+
+        ```
+        openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 9999 -nodes -subj /CN=127.0.0.1
+        ```
+
+    2. Add `--keyfile=key.pem` and `--certfile=cert.pem` to your esbuild development server command
+    3. Click past the scary warning in your browser when you load your page
+
+    If you have more complex needs than this, you can still put a proxy in front of esbuild and use that for HTTPS instead. Note that if you see the message "Client sent an HTTP request to an HTTPS server" when you load your page, then you are using the incorrect protocol. Replace `http://` with `https://` in your browser's URL bar.
+
+    Keep in mind that esbuild's HTTPS support has nothing to do with security. The only reason esbuild now supports HTTPS is because browsers have made it impossible to do local development with certain modern web features without jumping through these extra hoops. *Please do not use esbuild's development server for anything that needs to be secure.* It's only intended for local development and no considerations have been made for production environments whatsoever.
+
+* Better support copying `index.html` into the output directory ([#621](https://github.com/evanw/esbuild/issues/621), [#1771](https://github.com/evanw/esbuild/issues/1771))
+
+    Right now esbuild only supports JavaScript and CSS as first-class content types. Previously this meant that if you were building a website with a HTML file, a JavaScript file, and a CSS file, you could use esbuild to build the JavaScript file and the CSS file into the output directory but not to copy the HTML file into the output directory. You needed a separate `cp` command for that.
+
+    Or so I thought. It turns out that the `copy` loader added in version 0.14.44 of esbuild is sufficient to have esbuild copy the HTML file into the output directory as well. You can add something like `index.html --loader:.html=copy` and esbuild will copy `index.html` into the output directory for you. The benefits of this are a) you don't need a separate `cp` command and b) the `index.html` file will automatically be re-copied when esbuild is in watch mode and the contents of `index.html` are edited. This also goes for other non-HTML file types that you might want to copy.
+
+    This pretty much already worked. The one thing that didn't work was that esbuild's built-in development server previously only supported implicitly loading `index.html` (e.g. loading `/about/index.html` when you visit `/about/`) when `index.html` existed on the file system. Previously esbuild didn't support implicitly loading `index.html` if it was a build result. That bug has been fixed with this release so it should now be practical to use the `copy` loader to do this.
+
+* Fix `onEnd` not being called in serve mode ([#1384](https://github.com/evanw/esbuild/issues/1384))
+
+    Previous releases had a bug where plugin `onEnd` callbacks weren't called when using the top-level `serve()` API. This API no longer exists and the internals have been reimplemented such that `onEnd` callbacks should now always be called at the end of every build.
+
+* Incremental builds now write out build results differently ([#2104](https://github.com/evanw/esbuild/issues/2104))
+
+    Previously build results were always written out after every build. However, this could cause the output directory to fill up with files from old builds if code splitting was enabled, since the file names for code splitting chunks contain content hashes and old files were not deleted.
+
+    With this release, incremental builds in esbuild will now delete old output files from previous builds that are no longer relevant. Subsequent incremental builds will also no longer overwrite output files whose contents haven't changed since the previous incremental build.
+
+* The `onRebuild` watch mode callback was removed ([#980](https://github.com/evanw/esbuild/issues/980), [#2499](https://github.com/evanw/esbuild/issues/2499))
+
+    Previously watch mode accepted an `onRebuild` callback which was called whenever watch mode rebuilt something. This was not great in practice because if you are running code after a build, you likely want that code to run after every build, not just after the second and subsequent builds. This release removes option to provide an `onRebuild` callback. You can create a plugin with an `onEnd` callback instead. The `onEnd` plugin API already exists, and is a way to run some code after every build.
+
+* You can now return errors from `onEnd` ([#2625](https://github.com/evanw/esbuild/issues/2625))
+
+    It's now possible to add additional build errors and/or warnings to the current build from within your `onEnd` callback by returning them in an array. This is identical to how the `onStart` callback already works. The evaluation of `onEnd` callbacks have been moved around a bit internally to make this possible.
+
+    Note that the build will only fail (i.e. reject the promise) if the additional errors are returned from `onEnd`. Adding additional errors to the result object that's passed to `onEnd` won't affect esbuild's behavior at all.
+
+* Print URLs and ports from the Go and JS APIs ([#2393](https://github.com/evanw/esbuild/issues/2393))
+
+    Previously esbuild's CLI printed out something like this when serve mode is active:
+
+    ```
+     > Local:   http://127.0.0.1:8000/
+     > Network: http://192.168.0.1:8000/
+    ```
+
+    The CLI still does this, but now the JS and Go serve mode APIs will do this too. This only happens when the log level is set to `verbose`, `debug`, or `info` but not when it's set to `warning`, `error`, or `silent`.
+
+### Upgrade guide for existing code:
+
+* Rebuild (a.k.a. incremental build):
+
+    Before:
+
+    ```js
+    const result = await esbuild.build({ ...buildOptions, incremental: true });
+    builds.push(result);
+    for (let i = 0; i < 4; i++) builds.push(await result.rebuild());
+    await result.rebuild.dispose(); // To free resources
+    ```
+
+    After:
+
+    ```js
+    const ctx = await esbuild.context(buildOptions);
+    for (let i = 0; i < 5; i++) builds.push(await ctx.rebuild());
+    await ctx.dispose(); // To free resources
+    ```
+
+    Previously the first build was done differently than subsequent builds. Now both the first build and subsequent builds are done using the same API.
+
+* Serve:
+
+    Before:
+
+    ```js
+    const serveResult = await esbuild.serve(serveOptions, buildOptions);
+    ...
+    serveResult.stop(); await serveResult.wait; // To free resources
+    ```
+
+    After:
+
+    ```js
+    const ctx = await esbuild.context(buildOptions);
+    const serveResult = await ctx.serve(serveOptions);
+    ...
+    await ctx.dispose(); // To free resources
+    ```
+
+* Watch:
+
+    Before:
+
+    ```js
+    const result = await esbuild.build({ ...buildOptions, watch: true });
+    ...
+    result.stop(); // To free resources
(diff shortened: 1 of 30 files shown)


merceyz:
> You can create a plugin with an onEnd callback instead. The onEnd plugin API already exists, and is a way to run some code after every build.

Would it be an option to have `watch` etc. return an async iterator that gave you access to every build result as well?

evanw:
> Would it be an option to have `watch` return an async iterator that provided the build results as well?

Sure. But async iterators are a pretty obscure language feature so that doesn't seem like the best approach to me.

The `onRebuild` callback was removed because it was a bad API. One problem is that it arguably should have behaved like `onEnd` instead. Another problem is that the desire to get the build results isn't exclusive to watch mode, so having this API be specific watch mode was weird. I may add another API for this at some point in the future if the need arises. Adding a new API is relatively easy because it's not disruptive. This release focuses on the disruptive breaking changes that are necessary to fix these underlying API issues.

merceyz:
Yeah, that's true but it's alright though, I can get an async iterator from a `PassThrough` stream and a plugin with an `onEnd` callback if I need one.

clshortfuse:
Nice work!

The FireFox issue with EventSource is that it closes the connection on failure, which isn't part of spec. It should only close (and stop reconnecting) when `.close()` is called. You have to construct a new instance and rebind all the events.

I'll share a wrapper I use to work around this. There's also some automatic JSON parsing that's part of it, but can be removed. The logic can probably be written as a build plugin.

https://gist.github.com/clshortfuse/2120c0e87da8e28c03ae45b127ed5b54

jasonwilliams:
Hey @evanw happy new year. These changes look fantastic. 

Am I right in thinking watch still doesn’t output anything at all when started, but if we wanted to print some text (“watch mode started”) we would need to use onEnd? The issue was originally this https://github.com/evanw/esbuild/issues/980#issuecomment-886071953

evanw:
> Am I right in thinking watch still doesn’t output anything at all when started

That doesn't sound right. Watch mode should print out text like this:

```
[watch] build finished, watching for changes...
```

This happens at the `info` [log level](https://esbuild.github.io/api/#log-level), which is enabled by default for the CLI but not for the JS API. Make sure to set `logLevel: 'info'` if you're using the JS API and you want this.

> if we wanted to print some text (“watch mode started”) we would need to use onEnd?

If you wanted to print something similar, you could do it with a plugin like this:

```js
let examplePlugin = {
  name: 'build-logs',
  setup(build) {
    build.onStart(() => console.log('build started'))
    build.onEnd(() => console.log('build finished'))
  },
}
```

jasonwilliams:
> That doesn't sound right. Watch mode should print out text like this:

Sorry I don’t think I was clear, https://github.com/evanw/esbuild/issues/980#issuecomment-886071953 was raised because that message only appeared on the second and subsequent builds but not on the first one after calling watch. 

Am I right I’m thinking we’ll now get that message after calling watch.

clshortfuse:
So, working on that live reload. I wish there was a way to inject something other that files, so that I don't have to make a special entrypoint just for live-reload or have to build a second file just for esbuild to inject from. I started going down the path of making a plug-in, but I would have to attach to every entrypoint. Starting realizing that wasn't going to be possible:

````js
if (live) {
  buildOptions.plugins.push({
    name: 'live reload',
    setup(build) {
      const { entryPoints } = build.initialOptions;
      const fileNames = Array.isArray(entryPoints) ? entryPoints : [...Object.values(entryPoints)];
      const regex = fileNames
        .map((e) => `(${e.replace(/[$()*+.?[\\\]^{|}]/g, '\\$&')})`)
        .join('|');
      console.log(regex);

      build.onLoad({ filter: new RegExp(regex) }, async (args) => {
        // Would have to recursively build
      });
    },
  });
}
````

So, I'm trying the injection route, but since it's basically an ESM export module, that doesn't work for me either, since I would like that injection to run like a polyfill would. 

````js
const injections = [];
if (!isProduction && live) {
  injections.push('./esbuild.live.js');
}
````

I apologize if there is a way to inject runtime code to entrypoints, but I'm not that experienced with the plugin API.

evanw:
> Am I right in thinking we’ll now get that message after calling watch.

Yes. Calling `watch()` starts watch mode and the first thing watch mode does is to do a rebuild to discover what it should be watching. Since esbuild’s watch mode is precise (it only triggers on changes that can affect build output), it needs to do an initial build to discover the dependencies of the build.

PR #2829, opened 2023-01-15
Remove unreachable code on js_print.go

lfantonioli:
Since `if value != value` will always be false, I think the if statement's code is unreachable.
I tried to infer if it was a typo, but I couldn't find a plausible replacement for the "value" variable in this if statement.


evanw:
Load [this](https://go.dev/play/p/w2f02BlqBHb) and click "Run".

lfantonioli:
Sorry, my mistake.

PR #2879, opened 2023-01-30
Move the (recommended) live event reload source to `/.well-known/esbuild`.

lgarron:
This updates https://esbuild.github.io/api/#live-reload to support `/.well-known/esbuild` in addition to `/esbuild` and updates the release notes to recommend it. The docs would need a separate update.

I'd like to propose this change because the `/.well-known/` directory is a standard location to define a hardcoded path like this without collisions for real website content[^1].
It is not necessary to register such a path to use it, and it's unlikely that another use case would "legitimately" start using `/.well-known/esbuild`, but it's possible to register the path at https://github.com/protocol-registries/well-known-uris to reduce the risk of collisions even further.

[^1]: In this case, it might seem unlikely that `/esbuild` would be used as a path by an app. But it's common to use such paths to server "preview" versions of sites, named on new tools that use them. So an app at `https://example.com/` might plausibly host their preview site at `https://example.com/esbuild` if they're in the process of switching to `esbuild`. If one was building a product with `esbuild`-related features, it might also be reasonable to host a splash page for this at `/esbuild`. Apps and sites can work around this by renaming the relevant path if they want to use live reloading in development, but it's easy to support a much less collision-prone path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@
 
     This makes esbuild's TypeScript output smaller and faster when processing code that does this. I noticed this issue when I ran the TypeScript compiler's source code through esbuild's bundler. Now that the TypeScript compiler is going to be bundled with esbuild in the upcoming TypeScript 5.0 release, improvements like this will also improve the TypeScript compiler itself!
 
+* The recommended event source for live reload has changed to a custon [well-known URI](https://en.wikipedia.org/wiki/Well-known_URI) at `/.well-known/esbuild`. The old `/esbuild` path will continue to work, but consider updating the URL:
+
+    ```js
+    new EventSource('/.well-known/esbuild').addEventListener('change', () => location.reload())
+    ```
+
+
 ## 0.17.5
 
 * Parse `const` type parameters from TypeScript 5.0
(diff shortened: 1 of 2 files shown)


evanw:
The `/esbuild` event source path is deliberately only active when you explicitly request it with an `EventSource` object (specifically when the `Accept` header is equal to `text/event-stream`). Otherwise esbuild will fall back to its normal behavior. So esbuild's live reload can actually already coexist with e.g. a splash page at `/esbuild` without an issue. I assume you filed this issue without trying it first. You are welcome to try it to see for yourself that it already works.

lgarron:
> I assume you filed this issue without trying it first.

Indeed, I had some trouble debugging and figured I would send a PR rather than just filing an issue.

I would still advocate for using a `/.well-known` path as the place where I would generally expect to see such a URL placed as good practice[^1], and to avoid the need to sniff MIME types. But of course, it's your call!

[^1] e.g. to minimize assumptions for interoperating with other tooling. I know I've considered placing a proxy in front of `esbuild` at least once (to implement HTTP range requests, needed for `<video>` seeking in Chrome), and it probably won't be the last.

lgarron:
Closing for now. Feel free to reopen if you find my reasoning compelling.

PR #2897, opened 2023-02-06
Updated the deprecated ioutil dependency

zachary-walters:
The ioutil package has been deprecated as of Go v1.16: https://go.dev/doc/go1.16#ioutil

This commit replaces ioutil functions with their respective io/os functions.


evanw:
Thank you for the suggestion. However, I’m not going to do this given that a) this doesn’t actually fix anything and b) this breaks the build for people using old versions of Go.

PR #2968, opened 2023-03-05
Support native system access from wasm

### LINKED ISSUE #690: Provide virtual input file system to .build()
## Description

Allow an initial partial filesystem seed to be provided as an input to a `.build()` operation.

## Use-case

In order to increase the common capabilities between the WASM-based and binary-based implementations, it would be helpful to have some measure of control over the filesystem. I thought that one potential avenue for this that would avoid expensive and complicated coordination between `esbuild` and the controlling system might be to allow an initial FS snapshot to be seeded.

For my use-case, this would allow me to fully embrace `esbuild-wasm` and have platform and runtime environment portability.

lucacasonato:
This commit adds support for native system access for WASM. This is used
through a new `wasmSystemAccess` option on the `esbuild.initialize`
options bag. This options bag can contain either a specifier or
namespace for a Node.js compatible `fs` and `process` module. During
setup these are injected into `globalThis.fs` and `globalThis.process`,
where Go's WASM runtime uses them to back the regular Go FS APIs.
Finally, if this option is set, we set the `hasFS` option on
`createChannel`, and everything just works :)

Example for Deno:

```ts
import * as esbuild from "./deno/wasm.js"

await esbuild.initialize({ wasmSystemAccess: {
  fsSpecifier: "node:fs",
  processSpecifier: "node:process"
} })

await esbuild.build({
  format: "esm",
  entryPoints: ["./lib/deno/mod.ts"],
  target: "esnext",
  outfile: "./out.js"
})
```

Closes #690

### DIFF
--- a/lib/deno/wasm.ts
+++ b/lib/deno/wasm.ts
@@ -60,7 +60,7 @@ let initializePromise: Promise<Service> | undefined
 let stopService: (() => void) | undefined
 
 let ensureServiceIsRunning = (): Promise<Service> => {
-  return initializePromise || startRunningService('esbuild.wasm', undefined, true)
+  return initializePromise || startRunningService('esbuild.wasm', undefined, true, undefined)
 }
 
 export const initialize: typeof types.initialize = async (options) => {
@@ -68,8 +68,9 @@ export const initialize: typeof types.initialize = async (options) => {
   let wasmURL = options.wasmURL
   let wasmModule = options.wasmModule
   let useWorker = options.worker !== false
+  let wasmSystemAccess = options.wasmSystemAccess
   if (initializePromise) throw new Error('Cannot call "initialize" more than once')
-  initializePromise = startRunningService(wasmURL || 'esbuild.wasm', wasmModule, useWorker)
+  initializePromise = startRunningService(wasmURL || 'esbuild.wasm', wasmModule, useWorker, wasmSystemAccess)
   initializePromise.catch(() => {
     // Let the caller try again if this fails
     initializePromise = void 0
@@ -77,7 +78,10 @@ export const initialize: typeof types.initialize = async (options) => {
   await initializePromise
 }
 
-const startRunningService = async (wasmURL: string | URL, wasmModule: WebAssembly.Module | undefined, useWorker: boolean): Promise<Service> => {
+const startRunningService = async (
+  wasmURL: string | URL, wasmModule: WebAssembly.Module | undefined,
+  useWorker: boolean, wasmSystemAccess: types.WasmSystemAccess | undefined,
+): Promise<Service> => {
   let worker: {
     onmessage: ((event: any) => void) | null
     postMessage: (data: Uint8Array | ArrayBuffer | WebAssembly.Module) => void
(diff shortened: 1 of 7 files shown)


lucacasonato:
The test failure on Windows is related to https://github.com/denoland/deno/issues/18053. Shall I ignore Windows for now?

evanw:
> However this still does not work on Windows, because Go assumes unix-style paths when running within WASM, which is incompatible with Windows backed fs syscalls. This is an upstream limitation in Go's WASM target. As such, I have disabled support on Windows for now and added a doc comment.

Yes, this is unfortunate. I have filed an issue with Go about this and they have indicated that they do not plan on fixing this: https://github.com/golang/go/issues/43768. However, I have tried to work around this by forking the Go standard library and making the Unix-vs-Windows behavior a dynamic run-time decision instead of a static compile-time decision:

https://github.com/evanw/esbuild/blob/3b38ce37a148312d1b70084618932a45001e2e86/internal/fs/fs_real.go#L67-L73

Specifically the `CheckIfWindows()` function is a static compile-time decision for native builds (see [Unix](https://github.com/evanw/esbuild/blob/3b38ce37a148312d1b70084618932a45001e2e86/internal/fs/iswin_other.go) vs [Windows](https://github.com/evanw/esbuild/blob/3b38ce37a148312d1b70084618932a45001e2e86/internal/fs/iswin_windows.go)) except for WASM where it [happens at run-time](https://github.com/evanw/esbuild/blob/3b38ce37a148312d1b70084618932a45001e2e86/internal/fs/iswin_wasm.go). So esbuild is not actually using Go's built-in path manipulation code at all, and Go's limitations regarding this edge case should not apply. AFAIK this has been working fine when running esbuild with node's WebAssembly implementation and node's `fs` API.

So in theory esbuild should use Windows-style paths on Windows even with WASM and this should work. I wonder why this isn't working with your implementation. Ideally this would work on Windows too, as Windows is a platform that esbuild works hard to support natively. Do you know why esbuild's Windows-style path manipulation isn't happening in your case given that esbuild has built-in support for it? Is `CheckIfWindows()` not working in your case, or is it something else?

evanw:
Regarding this PR:

My first impression of this is that the `wasmSystemAccess` API seems too low-level. I think ideally this would just work without needing to configure things like this. You can already use esbuild's WASM-based API in node with the file system without extra configuration, and it seems like it would be a shame for esbuild's Deno API to be unnecessarily more complicated than esbuild's node API.

Also I don't necessarily want to expose the ability to inject a custom file system like this. That's already an existing feature request (#690) and I'd like to expose that ability more uniformly to all API environments that esbuild supports instead of only to Deno + WASM.

jakebailey:
Go just landed `GOOS=wasip1`; is it possible that this could be an easier way to embed things rather than exending the API?

Here's it working with `wazero`, passing through my env and FS:

image

jakebailey:
Took a couple weeks of trying all of the WASI shims + a fixed `os.Open` bug in Go, but, I got my playground to start using WASI (so long as you limit yourself to CLI args).

[Here's an example.](https://jakebailey.dev/esbuild-playground/#v3=PTAEAEDMEsBsFMB2BDAtvAXKZAnA5gM4B0AVgQPaIBQA2gEQC0DARgK6IAmCdANKIw0jkcqZABcAvPAKpe-JgQAOsaGLHREeOQPKsxHaDgm6x2jR3gAPImIKXtyDhZw2CdALpUqICDAQp0LHMrV0sqaFRFYTFQAG9sJ1AAX1BIHHJUfiJgR2c6AG5wyOjQWHI8ABVyADE4eFT0zLpsssqauoKvKyicGMh2AGN1SlAAQUVFAAoASjiqUFAceDFWHERQSYAeRQA+AAl4WDK+AHdhWA4AQjjcyYBGPgAmaaTN4F3pwqSuyx6+weG61aMzmC1aVVqCEmdDO5Do0yo328YCgdQCmASzlcVG6JX6iCG0BGt2QWEQrFQzHgOD4zDJFKpOFmsXmi2Wq3WyFAAGpQMwvl4fKj-GgMeD2ghSAQqANKAQ+gRQBI2QBHViGeDQyBuT5eVDkDisSW43qK5WTdAEAjIPDwWYSHag1LEZATJAcSHwADKAE8CdDgK0bJZTHxLdbbbqkkA)

So, to answer my own question, yes, it's very possible!

nestarz:
thanks @lucacasonato , any plan to have your PR merged or deliver wasm binaries for use (for example in Deno Fresh) ?

PR #2994, opened 2023-03-15
types: update Deno NS types

jsejcksn:
This PR updates the Deno namespace types.

It looks like the current types are from [the initial feature commit](https://github.com/evanw/esbuild/pull/1145/commits/cfedc07472e32b5d812107d598a644e0c7dceb10) from a couple of years ago at `2021-04-13T09:38:00Z`.

It's not obvious to me how this types file was created: AFAICT it's not a byte-exact representation of any artifact from a commit in the history of the Deno repository, but it's pretty close to [this one from the `1.8.3` release](https://github.com/denoland/deno/blob/v1.8.3/cli/dts/lib.deno.ns.d.ts) which was [the latest release at the time of the commit](https://github.com/denoland/deno/tags?after=v1.9.1).

<details>
<summary>
Examining the diff shows changes in insignificant whitespace and that the <a href="https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html">TypeScript triple-slash directives</a> were converted to plain comments:
</summary>

<pre><code>&gt; diff <a href="https://raw.githubusercontent.com/denoland/deno/v1.8.3/cli/dts/lib.deno.ns.d.ts">deno_version</a> <a href="https://raw.githubusercontent.com/evanw/esbuild/0c8a0a901d9a6c7bbff9b4dd347c8a3f65f6c6dd/lib/deno/lib.deno.ns.d.ts">esbuild_version</a></code></pre>

```diff
3,4c3,4
< /// <reference no-default-lib="true" />
< /// <reference lib="esnext" />
---
> // <reference no-default-lib="true" />
> // <reference lib="esnext" />
765,771c765,771
<       Reader,
<       ReaderSync,
<       Writer,
<       WriterSync,
<       Seeker,
<       SeekerSync,
<       Closer {
---
>     Reader,
>     ReaderSync,
>     Writer,
>     WriterSync,
>     Seeker,
>     SeekerSync,
>     Closer {
1947c1947
<      * 
---
>      *
1950c1950
<      * 
---
>      *
```
</details>

Using this deduction as a basis, this updates the type declaration files by copying from the [latest Deno release `1.31.2`](https://github.com/denoland/deno/releases/tag/v1.31.2):

1. https://github.com/denoland/deno/blob/v1.31.2/cli/tsc/dts/lib.deno.ns.d.ts
2. https://github.com/denoland/deno/blob/v1.31.2/ext/net/lib.deno_net.d.ts

and replaces the [lib](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-lib-) reference [in the first file](https://github.com/denoland/deno/blob/v1.31.2/cli/tsc/dts/lib.deno.ns.d.ts#L5) with a relative [path](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-) reference.

### DIFF
--- a/lib/deno/lib.deno.ns.d.ts
+++ b/lib/deno/lib.deno.ns.d.ts
@@ -1,13 +1,31 @@
-// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 
-// <reference no-default-lib="true" />
-// <reference lib="esnext" />
+/// <reference no-default-lib="true" />
+/// <reference lib="esnext" />
+/// <reference path="./lib.deno_net.d.ts" />
 
-/** Deno provides extra properties on `import.meta`.  These are included here
+/** Deno provides extra properties on `import.meta`. These are included here
  * to ensure that these are still available when using the Deno namespace in
- * conjunction with other type libs, like `dom`. */
+ * conjunction with other type libs, like `dom`.
+ *
+ * @category ES Modules
+ */
 declare interface ImportMeta {
-  /** A string representation of the fully qualified module URL. */
+  /** A string representation of the fully qualified module URL. When the
+   * module is loaded locally, the value will be a file URL (e.g.
+   * `file:///path/module.ts`).
+   *
+   * You can also parse the string as a URL to determine more information about
+   * how the current module was loaded. For example to determine if a module was
+   * local or not:
+   *
+   * ```ts
+   * const url = new URL(import.meta.url);
+   * if (url.protocol === "file:") {
+   *   console.log("this module was loaded locally");
+   * }
+   * ```
+   */
   url: string;
 
   /** A flag that indicates if the current module is the main module that was
@@ -20,12 +38,27 @@ declare interface ImportMeta {
    * ```
    */
   main: boolean;
+
+  /** A function that returns resolved specifier as if it would be imported
+   * using `import(specifier)`.
+   *
+   * ```ts
+   * console.log(import.meta.resolve("./foo.js"));
+   * // file:///dev/foo.js
+   * ```
+   */
+  resolve(specifier: string): string;
 }
 
-/** Deno supports user timing Level 3 (see: https://w3c.github.io/user-timing)
- * which is not widely supported yet in other runtimes.  These types are here
- * so that these features are still available when using the Deno namespace
- * in conjunction with other type libs, like `dom`. */
+/** Deno supports [User Timing Level 3](https://w3c.github.io/user-timing)
+ * which is not widely supported yet in other runtimes.
+ *
+ * Check out the
+ * [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance)
+ * documentation on MDN for further information about how to use the API.
+ *
+ * @category Performance
+ */
 declare interface Performance {
   /** Stores a timestamp with the associated name (a "mark"). */
   mark(markName: string, options?: PerformanceMarkOptions): PerformanceMark;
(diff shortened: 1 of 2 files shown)


lib/deno/lib.deno.ns.d.ts
```diff
@@ -1,13 +1,31 @@
-// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 
-// <reference no-default-lib="true" />
-// <reference lib="esnext" />
+/// <reference no-default-lib="true" />
+/// <reference lib="esnext" />
+/// <reference path="./lib.deno_net.d.ts" />
```
jsejcksn, line 5: Manually modified to use a relative path reference

evanw (changes requested):
This PR introduces type errors, and so will not be merged as-is (see CI for details).

evanw:
After looking into this a bit, it looks like Deno's type declarations have been changed into many files, which is a bit unwieldy to keep up to date. It also looks like I should be using the `deno types` command instead of checking these type declarations in at all.

evanw:
Ok I have switched esbuild over to using `deno types`. The CLI help for this command says the following:

```
$ deno help types
deno-types 
Print runtime TypeScript declarations.

  deno types > lib.deno.d.ts

The declaration file could be saved and used for typing information.

USAGE:
    deno types [OPTIONS]

OPTIONS:
    -h, --help
            Print help information

    -q, --quiet
            Suppress diagnostic output

        --unstable
            Enable unstable features and APIs
```

So this command is intended to be used like this. This way the type definitions should always match the minimum version of Deno that esbuild supports (as represented by the Deno version that esbuild runs against in CI) without needing to include Deno's API evolution in esbuild's git history.

jsejcksn:
@evanw Thanks for providing a bit of clarity.

I thought you might have intentionally included only the Deno namespace types in attempt to [prevent a type-checking conflict with the vanilla TS DOM library types](https://deno.land/manual@v1.31.3/advanced/typescript/configuration#using-the-lib-property) in a custom configuration. Again — there's no documentation for how the types were created or used... but after seeing the test failure, it is now more clear.

> After looking into this a bit, it looks like Deno's type declarations have been changed into many files, which is a bit unwieldy to keep up to date. It also looks like I should be using the `deno types` command instead of checking these type declarations in at all.

If the goal is type-check with all of the Deno types, then — as you observed — you can use the `deno types` command from Deno CLI to generate a bundled type declaration file. An identical file is also available as an artifact in every [release](https://github.com/denoland/deno/releases) — so if a certain version is desired, it might be simpler to download it instead of needing to install that specific version of Deno and generate it.

The URL format is like this:

```
https://github.com/denoland/deno/releases/download/{TAGNAME}/lib.deno.d.ts
```

So, for example, the file which corresponds to the [`v1.31.3`](https://github.com/denoland/deno/releases/tag/v1.31.3) release is at:

<pre><code><a href="https://github.com/denoland/deno/releases/download/v1.31.3/lib.deno.d.ts">https://github.com/denoland/deno/releases/download/v1.31.3/lib.deno.d.ts</a></pre></code>

<details>
<summary>
I used this GraphQL query and it looks like the file starts to appear in release <code>v0.39.0</code> and later:</summary>

```gql
query DenoDts($after: String) {
  repository(name: "deno", owner: "denoland") {
    releases(
      orderBy: {field: CREATED_AT, direction: ASC}
      first: 100
      after: $after
    ) {
      pageInfo {
        startCursor
        endCursor
      }
      nodes {
        releaseAssets(first: 100, name: "lib.deno.d.ts") {
          nodes {
            downloadUrl
          }
        }
        tagName
        createdAt
      }
    }
  }
}
```
</details>

Issue #2081, opened 2022-03-04
Provide esbuild-wasm build that does not require workers or dynamic code instantiation


lucacasonato:
Currently when using `esbuild-wasm` you have two options:

- either you enable `worker: true` in the setup options, which means your environment requires support for web workers,
- _or_ you don't enable workers, in which case the esbuild code will call `new Function(...)` with the code originally destined for the web worker

The latter case does not work on platforms like Deno Deploy or Cloudflare Workers, which limit dynamic JS eval.

It would be great if the second option (not using a worker), did not require the use of dynamic code evaluation. Could the source code always be a function in source, and then `.toString()`'d for the worker?

lucacasonato:
Thanks a lot @evanw!

PR #3056, opened 2023-04-13
pref(wasm): deduplicated web-worker code in browser.js

privatenumber:
## Problem

In the WASM distribution, a pretty large Web Worker function was being duplicated and creating bloat.

## Changes

Reuse Web Worker code by leveraging `toString()`.

File|Size before|Size after
-|-|-
`lib/browser.js` | `122 KB` | `95 KB`
`lib/browser.min.js` | `50 KB` | `40KB`
`esm/browser.js` | `118 KB` | `93 KB`
`esm/browser.min.js` | `48 KB` | `39 KB`

### DIFF
--- a/lib/npm/browser.ts
+++ b/lib/npm/browser.ts
@@ -78,13 +78,15 @@ const startRunningService = async (wasmURL: string | URL, wasmModule: WebAssembl
     terminate: () => void
   }
 
+  const webWorkerFunction = WEB_WORKER_FUNCTION;
+
   if (useWorker) {
     // Run esbuild off the main thread
-    let blob = new Blob([`onmessage=${WEB_WORKER_SOURCE_CODE}(postMessage)`], { type: 'text/javascript' })
+    let blob = new Blob([`onmessage=${webWorkerFunction.toString()}(postMessage)`], { type: 'text/javascript' })
     worker = new Worker(URL.createObjectURL(blob))
   } else {
     // Run esbuild on the main thread
-    let onmessage = WEB_WORKER_FUNCTION((data: Uint8Array) => worker.onmessage!({ data }))
+    let onmessage = webWorkerFunction((data: Uint8Array) => worker.onmessage!({ data }))
     worker = {
       onmessage: null,
       postMessage: data => setTimeout(() => onmessage({ data })),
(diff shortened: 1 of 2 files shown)


evanw:
This used to be how esbuild works. However, it was deliberately changed because esbuild (and likely some other tools too) [don't fully support code that uses `Function.prototype.toString`](https://esbuild.github.io/content-types/#function-tostring).

I recommend not using esbuild if you're worried about this level of bloat. This PR saves ~10 kb but the WebAssembly binary for esbuild is currently more than 10,000 kb, so this PR only reduces the size by about 0.1%.

privatenumber:
> This used to be how esbuild works. However, it was deliberately changed because esbuild (and likely some other tools too) [don't fully support code that uses Function.prototype.toString](https://esbuild.github.io/content-types/#function-tostring).

Ah right. I wonder if it could be the other way around then? Keep it as a string, and run it through `eval` or `new Function` for the main thread version.

> I recommend not using esbuild if you're worried about this level of bloat. This PR saves ~10 kb but the WebAssembly binary for esbuild is currently more than 10,000 kb, so this PR only reduces the size by about 0.1%.

It's true the total size savings are trivial. But in my usage, I'm bundling the JS part in the same chunk as my entry-point, and the WASM binary is loaded separately. So the reduction in the JS files had some notable impact.

evanw:
That also used to be how esbuild works (using `new Function` to construct the worker function). However, that prevented people from using esbuild in environments where dynamic JavaScript evaluation is forbidden, such as Cloudflare Workers or Deno Deploy: #2081. So esbuild was changed to work the way it does now to enable these use cases.

privatenumber:
Gotcha, sorry to make you revisit this problem. I underestimated how complex this is. I appreciate your patience in explaining this to me.

You may have already thought of this too, but what do you think of separating out the Web Worker/`worker_threads` versions into another file/entry-point?

```json5
"exports": {
	".": {
		"node": {
			"worker": "./lib/node/worker_threads.js",
			"default": "./lib/node/main.js"
		},
		"default": {
			"worker": "./lib/browser/web-worker.js",
			"default": "./lib/browser/main.js"
		}
	},
	"./main": {
		"node": "./lib/node/main.js"
		"default": "./lib/browser/main.js"
	},
	"./worker": {
		"node": "./lib/node/worker_threads.js"
		"default": "./lib/browser/web-worker.js"
	}
}
"

PR #3069, opened 2023-04-17
perf(linker): Fixes brute force chunk cycle detection 

### LINKED ISSUE #3068: Naive chunks' cycle detect assertion
## Context

I work on a custom inner loop bundle tool on top of esbuild, and I've noticed a regression on our build times. We work with the concept of "strategy", which is a way to coordinate more than one build. One of those is a "vendor strategy", which is analogous to vite's vendor build, which creates one build for user code, and another for vendor code. This build was taking a ridiculous amount of time (~470s on a sufficiently resourceful hardware, 64G 32vCPUs). 

## Investigation

After collecting some profiles, I've found that `linkerContext.enforceNoCyclicChunkImports` was behaving strangely:

Specially so when considering that its behavior is of an assert-like nature; that is, to ensure that the chunking algorithm behaved properly.

https://github.com/evanw/esbuild/blob/f0704baa38ef9a808b7e34cac75aa4ac285bb087/internal/linker/linker.go#L454-L478
 
It's a straightforward [cycle detection](https://en.wikipedia.org/wiki/Cycle_(graph_theory)#Cycle_detection) algorithm. However, given that our app is _huge_ and has large-enough dependency graph and connected-enough edge numbers, we have stress tested this algorithm and climbed the $N^2$ slope enough to be noticeable.

### But why so slow?

The worst-case time complexity of this algorithm can be broke down to:
- A DFS run: $O(N+E)$
- That is run for all chunks: $O(N * (N+E))$

where $N$ is the number of chunks, and $E$ is the number of edges on the graph.

In short, it's performing poorly because it is a brute force algorithm, that is being tested on a sufficiently large input space: $N > 2,500$. Breaking down what the algorithm does:

1. Iterates over all chunks and calls `validate(chunk)`
2. Compares the current chunk index with the indices in `path`, which is tracking the visited nodes.
3. Recursively calling validate for all `chunk.crossChunkImports`.

But because we are both iterating on all chunks _and_ recursively calling `validate` for all `chunk.crossChunkImports` without no mechanism to reduce the search space (i.e leave early), we are bound to do _lots_ of repetitive work. That is true to the amount of edges on the graph, because that would mean longer `validate` call stacks for each node that would be repeated over and over.

ruiconti:
Fixes #3068

## Approach

Given that the behavior is concerned only on _detecting_ a cycle rather than _identifying_ them, the simplest possible revision would be to improve current approach's shortcomings. Practically, this would mean

1. Memoize: store and use the results of validating a given node
2. Reduce the search space: skip nodes that have already been visited

I used an approach borrowed from [topological sorting](https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search), in which nodes are marked with a state (not-visited, visiting and visited), and if they've already been visited, we skip them. Furthermore, if we find a cycle, we halt. That gives us a linear execution time, rather than exponential.

## Results and reproducibility

It's clear that the impact of this would only be seen on some areas of the input space. Particularly, past a given graph complexity. Consuming the change from this fork, the vendor build which was taking up to ~400s, dropped to 4s, a **100x (anecdotal) improvement**. The app build's incremental times dropped from ~150s to ~30s, a 5x improvement.

Since this is a plain algorithmic revision, I believe that the point could be proven with a benchmarked test, and didn't see the need to generate a sufficiently complex npm dependency graph and run a full `esbuild build` on it.

Code Sandbox: https://codesandbox.io/p/sandbox/blue-http-750ebf?file=%2Flinker_test.go&selection=%5B%7B%22endColumn%22%3A6%2C%22endLineNumber%22%3A10%2C%22startColumn%22%3A6%2C%22startLineNumber%22%3A10%7D%5D

<pre>
PASS
BenchmarkLegacy100Sparse          100000             12931 ns/op             896 B/op          1 allocs/op
BenchmarkLegacy200Sparse             100          15657528 ns/op            1818 B/op          3 allocs/op
<b>BenchmarkLegacy300Sparse               1        19536221791 ns/op          28712 B/op        446 allocs/op</b>
BenchmarkRevision50Sparse         100000             12944 ns/op            2542 B/op          7 allocs/op
BenchmarkRevision100Sparse         50000             26291 ns/op            5279 B/op         11 allocs/op
BenchmarkRevision200Sparse         30000             55162 ns/op           10769 B/op         18 allocs/op
<b>BenchmarkRevision300Sparse         20000             98047 ns/op           20824 B/op         23 allocs/op</b>
BenchmarkRevision1000Sparse         3000            454412 ns/op           88313 B/op         67 allocs/op
</pre>

### DIFF
--- a/internal/linker/linker.go
+++ b/internal/linker/linker.go
@@ -445,41 +445,6 @@ func (c *linkerContext) mangleProps(mangleCache map[string]interface{}) {
 // can cause initialization bugs. So let's forbid these cycles for now to guard
 // against code splitting bugs that could cause us to generate buggy chunks.
 func (c *linkerContext) enforceNoCyclicChunkImports() {
-	var validate func(int, map[int]bool, map[int]bool) bool
-	inPath := make(map[int]bool)
-	visited := make(map[int]bool)
-
-	validate = func(chunkIndex int, visited, inPath map[int]bool) bool {
-		if visited[chunkIndex] {
-			return false
-		}
-		if inPath[chunkIndex] {
-			c.log.AddError(nil, logger.Range{}, "Internal error: generated chunks contain a circular import")
-			return true
-		}
-
-		inPath[chunkIndex] = true
-
-		for _, chunkImport := range c.chunks[chunkIndex].crossChunkImports {
-			if chunkImport.importKind != ast.ImportDynamic {
-				if validate(int(chunkImport.chunkIndex), visited, inPath) {
-					return true
-				}
-			}
-		}
-
-		visited[chunkIndex] = true
-		return false
-	}
-
-	for chunkIndex := range c.chunks {
-		if validate(chunkIndex, visited, inPath) {
-			break // leave early
-		}
-	}
-}
-
-func (c *linkerContext) enforceNoCyclicChunkImportsColor() {
 	var validate func(int, map[int]int) bool
 
 	// DFS memoization with 3-colors, more space efficient
--- a/internal/linker/linker.go
+++ b/internal/linker/linker.go
@@ -445,24 +445,24 @@ func (c *linkerContext) mangleProps(mangleCache map[string]interface{}) {
 // can cause initialization bugs. So let's forbid these cycles for now to guard
 // against code splitting bugs that could cause us to generate buggy chunks.
 func (c *linkerContext) enforceNoCyclicChunkImports() {
-	var validate func(int, []int)
-	validate = func(chunkIndex int, path []int) {
+	// var validate func(int, []int)
+	validate := func(chunkIndex int, path []int) {
 		for _, otherChunkIndex := range path {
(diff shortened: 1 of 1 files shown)


evanw:
Thanks for the improvement and for the detailed report.

However, note that encountering this problem likely means that esbuild is a poorly-suited tool for your use case. It's not good to have too many chunks at run-time and esbuild doesn't currently have the ability to control the number of chunks (as it's code splitting feature is still in a WIP state). So you probably want to look for another tool even after this improvement has landed.

MLoughry:
@evanw We currently use `esbuild` for dev innerloop builds, where the output is "good enough" and the build speed is much faster than our `webpack` dev builds. But we will certainly keep that in mind if/when we investigate bundlers aside from `webpack` for production builds.

PR #3091, opened 2023-04-28
fix(cli): should not hang for empty stdin

hyf0:
I want to see the output of running `esbuild --log-level=verbose` and found the esbuild hang forever.

After investigating, the hanging is caused by empty stdin.


evanw:
This PR breaks using stdin to provide input to esbuild. It causes esbuild to terminate early before you type anything. Without this PR, you can use esbuild without providing a file as input by typing `./esbuild [options]`, typing the input, and then ending stdin with Ctrl+D. What you are calling a "hang" is just esbuild waiting for input when you run esbuild in stdin mode.

hyf0:
Thanks for explaining.

Issue #3125, opened 2023-05-19
[Bug] ESBuild minify results in `ReferenceError: o is not defined` for a const defined in an if block


sarahfank:
# Description
A const is defined within an if block and then used inside a function that is also defined in the if block. After minification when running in production, the value of the const is not being replaced properly in the function and is resulting in `ReferenceError: o is not defined`.

# Reproduce
To reproduce this, create a new vite project unsing `npm create vite@latest` and replace the contents of `main.ts` with the following: 
```
function main() {
  const is_visible = true;
  if (is_visible) {
    const MY_CONST = 20;

    console.log('above function call', MY_CONST);   // this works

    function myFunction () {
      console.log(MY_CONST);    // o is not defined
    }

    myFunction();
  }
}

main();
```

When you build and serve the dist folder and open in the browser, you will see `ReferenceError: o is not defined`.

hyrious:
This issue involves a rollup tree shaking step, here's the reproduction steps without using vite:

```js
// main.ts = same as provided above

$ esbuild main.ts
function main() {
  const is_visible = true;
  if (is_visible) {
    let myFunction2 = function() {
      console.log(MY_CONST);
    };
    var myFunction = myFunction2;  // <- not used, will be shaked by rollup
    const MY_CONST = 20;
    console.log("above function call", MY_CONST);
    myFunction2();
  }
}
main();

$ esbuild main.ts | rollup
function main() {
  {
    let myFunction2 = function() {
      console.log(MY_CONST);  // <- constant folding bug!
    };
    const MY_CONST = 20;
    console.log("above function call", MY_CONST);
    myFunction2();
  }
}
main();

$ esbuild main.ts | rollup | esbuild --minify-syntax
function main() {
  {
    let myFunction2 = function() {
      console.log(MY_CONST);  // <- reference error
    };
    console.log("above function call", 20), myFunction2();
  }
}
main();
```

The steps corresponds to how vite build works:

1. A rollup() at the outside..
2. ..with an esbuild plugin that transpiles ts to js.
3. Do esbuild --minify to the result of 1.

hyrious:
@DragonLiParadise The issue is because esbuild is simulating the variable hoisting of function names and there's a bug in constant folding when a const name was used lexically before its difinition. So you could just don't write function body in arbitrary blocks (if-blocks, for-blocks, etc.)

jluxenberg:
[...]

PR #3150, opened 2023-06-08
test: failing tests for constant folding and TDZ issues

jluxenberg:
RE: https://github.com/evanw/esbuild/issues/3125

Spent a bit of time trying to fix these in `js_parser.go` but couldn't solve it. This PR has a few failing tests which illustrate the problem.

There are two issues:

(1) Constant propagation not working properly when the constant is declared after it is used

Consider this program:
```js
{ 
  const f = () => x; 
  const x = 0; 
  f();
}
```

The definition of `x` should be hoisted to the top of the enclosing block. Since it is constant, it is eligible for constant folding/propagation. The propagation should respect the Temporal Dead Zone; accessing `x` before it is initialized should raise a `ReferenceError: Cannot access 'x' before initialization`.

`esbuild` minifies this program to `(() => x)()`. I believe this is because `x` is correctly identified as a constant, but then isn't able to do the substitution into the function body because it is defined after the function that uses it.

(2) Dead statement removal doesn't respect the Temporal Dead Zone

Consider this program:
```js
{
  number; 
  const number=10;
}
```

This should crash with an error `ReferenceError: Cannot access 'number' before initialization`.

However, `esbuild` minifies it to something like `{const n=10}`, omitting the `number;` statement.

### DIFF
--- a/internal/js_parser/js_parser_test.go
+++ b/internal/js_parser/js_parser_test.go
@@ -2675,6 +2675,15 @@ func TestConstantFolding(t *testing.T) {
 	expectPrinted(t, "x = -0 == false", "x = true;\n")
 	expectPrinted(t, "x = 1 == true", "x = true;\n")
 	expectPrinted(t, "x = 2 == true", "x = false;\n")
+
+	expectPrintedMangle(t, `{ const x = 0; const f = () => x; f(); };`, "(() => 0)();\n")
+	expectPrintedMangle(t, `{ const f = () => x; const x = 0; f(); };`, "(() => 0)();\n")
+	expectPrintedMangle(t, `const f = () => x; const x = 0; f();`, "(() => 0)();\n")
+}
+
+func TestTemporalDeadZone(t *testing.T) {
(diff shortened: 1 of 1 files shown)


evanw:
Regarding (1), the issue #3125 was fixed and the fix already includes a test.

Regarding (2), this is because [bound identifiers are considered to be side-effect free](https://github.com/evanw/esbuild/blob/a7236e4d36f60f367c0583fba8204454a9cc6754/internal/js_ast/js_ast_helpers.go#L2078-L2081) to allow for important optimizations. It's true that this means esbuild can incorrectly remove TDZ checks such as in the case you presented. However, code that triggers TDZ checks is inherently broken anyway, and disabling common and important tree-shaking optimizations for a rare edge case that only appears in broken code doesn't seem like a good trade-off for a production-oriented bundler like esbuild. This is why I have deliberately made the decision to not respect the TDZ check here.

jluxenberg:
Thanks for the detailed explanation and fix for #3125!

PR #3194, opened 2023-06-25
lower async generator functions

### LINKED ISSUE #2780: Support transforming async generators like @babel/plugin-proposal-async-generator-functions does
It would be nice to be able to have esbuild transform async generators like the babel plugin `@babel/plugin-proposal-async-generator-functions` does when native async support is disabled, that is:
```ts
   esbuild: {
     supported: {
       "async-await": false,
     }
  }
```

Currently using async generator functions with async await disabled results in an error.

This is used as a stopgap by Angular's Zone.JS so they can track async generator functions as they have no way to track native async await in the browser ATM, so they rely on down transpiling them to generators.

See: https://github.com/angular/angular-cli/blob/67ffa8b40be04b0e671429063810fe5f52b2043a/packages/angular_devkit/build_angular/src/babel/presets/application.ts#L232-L239

If esbuild supported this natively, babel wouldn't be needed here.

evanw:
With this PR, esbuild will now transform `async` generator functions into normal generator functions when the configured target environment doesn't support them. These functions behave similar to normal generator functions except that they use the `Symbol.asyncIterator` interface instead of the `Symbol.iterator` interface and the iteration methods return promises. Here's an example (helper functions are omitted):

```js
// Original code
async function* foo() {
  yield Promise.resolve(1)
  await new Promise(r => setTimeout(r, 100))
  yield *[Promise.resolve(2)]
}
async function bar() {
  for await (const x of foo()) {
    console.log(x)
  }
}
bar()

// New output (with --target=es6)
function foo() {
  return __asyncGenerator(this, null, function* () {
    yield Promise.resolve(1);
    yield new __await(new Promise((r) => setTimeout(r, 100)));
    yield* __yieldStar([Promise.resolve(2)]);
  });
}
function bar() {
  return __async(this, null, function* () {
    try {
      for (var iter = __forAwait(foo()), more, temp, error; more = !(temp = yield iter.next()).done; more = false) {
        const x = temp.value;
        console.log(x);
      }
    } catch (temp) {
      error = [temp];
    } finally {
      try {
        more && (temp = iter.return) && (yield temp.call(iter));
      } finally {
        if (error)
          throw error[0];
      }
    }
  });
}
bar();
```

This is an older feature that was added to JavaScript in ES2018 but I didn't implement the transformation then because it's a rarely-used feature. Note that esbuild already added support for transforming `for await` loops (the other part of the [asynchronous iteration proposal](https://github.com/tc39/proposal-async-iteration)) a year ago, so support for asynchronous iteration should now be complete.

I have never used this feature myself and code that uses this feature is hard to come by, so this transformation has not yet been tested on real-world code. If you do write code that uses this feature, please let me know if esbuild's `async` generator transformation doesn't work with your code.

Fixes #2780

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -355,6 +355,67 @@ func Source(unsupportedJSFeatures compat.JSFeature) logger.Source {
 			})
 		}
 
+		// These help for lowering async generator functions
+		export var __await = function (promise, isYieldStar) {
+			this[0] = promise
+			this[1] = isYieldStar
+		}
+		export var __asyncGenerator = (__this, __arguments, generator) => {
+			var resume = (k, v, yes, no) => {
+				try {
+					var x = generator[k](v), isAwait = (v = x.value) instanceof __await, done = x.done
+					Promise.resolve(isAwait ? v[0] : v)
+						.then(y => isAwait
+							? resume(k === 'return' ? k : 'next', v[1] ? { done: y.done, value: y.value } : y, yes, no)
+							: yes({ value: y, done }))
+						.catch(e => resume('throw', e, yes, no))
+				} catch (e) {
+					no(e)
+				}
+			}
+			var method = k => it[k] = x => new Promise((yes, no) => resume(k, x, yes, no))
+			var it = {}
+			return generator = generator.apply(__this, __arguments),
+				it[Symbol.asyncIterator] = () => it,
+				method('next'),
+				method('throw'),
+				method('return'),
+				it
+		}
+		export var __yieldStar = value => {
+			var obj = value[__knownSymbol('asyncIterator')]
+			var isAwait = false
+			var method
+			var it = {}
+			if (obj == null) {
+				obj = value[__knownSymbol('iterator')]()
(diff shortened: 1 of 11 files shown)


fedellen:
Thanks for the fix and the write up here ❤️

internal/runtime/runtime.go
```diff
@@ -355,6 +355,67 @@ func Source(unsupportedJSFeatures compat.JSFeature) logger.Source {
...
+			var it = {}
+			return generator = generator.apply(__this, __arguments),
+				it[Symbol.asyncIterator] = () => it,
+				method('next'),
+				method('throw'),
+				method('return'),
+				it
+		}
+		export var __yieldStar = value => {
+			var obj = value[__knownSymbol('asyncIterator')]
```
tido64, line 386: @evanw: Hi, I'm currently debugging a crash we're seeing after bumping esbuild to a more recent version where the only diff is the addition of this polyfill. Is this line meant to throw if `asyncIterator` is not defined/supported by the VM? What is then the reason for check on line 390 below?

PR #3217, opened 2023-07-08
[deno] add support for aarch64-linux-android

secext2022:
see this issue: <https://github.com/denoland/fresh/issues/1421>


lib/deno/mod.ts
```diff
@@ -152,6 +152,7 @@ async function install(): Promise<string> {
     'aarch64-unknown-linux-gnu': '@esbuild/linux-arm64',
     'x86_64-apple-darwin': '@esbuild/darwin-x64',
     'x86_64-unknown-linux-gnu': '@esbuild/linux-x64',
+    'aarch64-linux-android': '@esbuild/android-arm64',
```
evanw, line 155: You have put this in the "platforms that Deno supports" section but AFAIK this is not one of the platforms that Deno supports. See for example https://deno.land/x/install@v0.1.8/install.sh?source=. So this should go in the other section instead.

secext2022: ok, moved

PR #3248, opened 2023-07-19
Update golang.org/x/sys v0.10.0 for CWE-266

madhavajay:
Hi,
I am getting this CWE when using esbuild in our containers.

https://security.snyk.io/vuln/SNYK-GOLANG-GOLANGORGXSYSUNIX-3310442

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,4 +4,4 @@ module github.com/evanw/esbuild
 // themselves for old OS versions. Please do not change this.
 go 1.13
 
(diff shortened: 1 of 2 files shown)


evanw:
Sorry but this PR prevents esbuild from building with Go 1.13, so I'm not going to take this PR. I'm keeping esbuild working with old versions of Go so that people can build esbuild for older versions of macOS (the Go team has removed support for older versions of macOS in newer Go versions).

File system access in esbuild is not security-sensitive so this alert you're getting doesn't seem relevant. I recommend disabling the alert.

PR #3249, opened 2023-07-19
feat: support CSS @starting-style rule

yisibl:
This at rule allow authors to start CSS transitions on first style update. That is, you can now make the transition take effect when the display property changes from none to block.

- Spec: https://drafts.csswg.org/css-transitions-2/#defining-before-change-style-the-starting-style-rule
- Shipped in Chrome 117: https://chromestatus.com/feature/4515377717968896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,22 @@
     }
     ```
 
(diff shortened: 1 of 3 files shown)


evanw:
Cool! Thanks for the PR. TIL about `@starting-style`.

PR #3260, opened 2023-07-20
Fix CommonOptions.define type

dr-vortex:
Fixes the type of CommonOptions.define to be consistent with the documentation:

> Replacement expressions must either be a JSON object (null, boolean, number, string, array, or object) or a single identifier

https://esbuild.github.io/api/#define


evanw:
The type annotations are correct. Please read the documentation more carefully. Specifically the code example immediately before the text you copied and pasted.

PR #3280, opened 2023-08-02
fix: quotes cannot be removed when the value of animation-name is a CSS Wide keywords

yisibl:
Continuing to improve: https://github.com/evanw/esbuild/commit/4973f9e45fbf9e5fcfadc7d6f193bddafe901267 and https://github.com/evanw/esbuild/commit/e77404a482479c1739e929e6caab36b8b6bb8868


evanw:
Thanks for the fix. However, I'm going to take a slightly different approach. Strings must still be converted to symbols even if they contain CSS keywords because local symbols must be renamed to avoid collisions.

PR #3299, opened 2023-08-09
fix(TsconfigRaw): `baseUrl` to be string

privatenumber:
- https://www.typescriptlang.org/tsconfig#baseUrl
- https://github.com/sindresorhus/type-fest/blob/main/source/tsconfig-json.d.ts#L709


evanw:
Thanks for the fix!

Issue #3341, opened 2023-08-25
esbuild tries to resolve wrong css file name in any node_module folder with an upper case .tsx-component


jonasem:
After upgrading esbuild 0.17 -> 0.19 we experienced a weird behavior where esbuild will try to resolve .css-files for components in node_modules even though they are not imported, just merely by beeing on the disk.

We recreated this in a tiny project only installing react@18 and esbuild@0.19 and adding a folder in node_modules like this:

![image](https://github.com/evanw/esbuild/assets/1006193/187f95e3-b7c1-4cd4-829e-1e26d8b23797)

The error output is:

`Could not read from file: /home/jonas/tmp/esbuild-bug/node_modules/foo/bar/A.css`

Note that it looks for a big-letter A.css, while we have a small one, and the kicker is, it's not ever imported anywhere. 

We tested this multiple times, just making a lower-Pascal-case css file matching an upper case .tsx file triggers this error.

evanw:
Please provide a way to reproduce the issue. You have not provided enough information here. Marking this issue as `unactionable` for now since I cannot do anything without a way to reproduce the issue.

jonasem:
Sure, you can try this repo I made to show the issue https://github.com/jonasem/esbuild-component-css-bug

hyrious:
You can enable `logLevel: 'verbose'` to see what actually happens in detail, then you will find this log:

```
⬥ [VERBOSE] Resolving import "./bar/AComponent" in directory "/node_modules/foo" of type "import-statement"
...
    Checking for file "AComponent"
    Checking for file "AComponent.jsx"
    Checking for file "AComponent.js"
    Checking for file "AComponent.css"
    Found file "AComponent.css"
```

So here are 2 things:

- when resolving extension-less imports, esbuild searches for `.css` files, this may be wrong
- esbuild's resolver is case-insensitive, this is ok. however it failed to work well on case-sensitive fs like on linux

jonasem:
Thank you for the insight into how to debug this issue. So I didn't expect this behavior, but it make sense when I see the error. Renaming the file to uppercase gave an error during runtime which also makes sense with the given verbose log.I hope I don't have to import typescript files by giving the full extension though, esbuils didn't do that before in this case.

hyrious:
[...]

PR #3352, opened 2023-08-28
Resolve css after ts in node_modules

jonasem:
Addresses #3341 

Typescript files in node_modules will currently resolve css files over ts files when importing without extension.


vprus:
Jonas, can you tell if this PR was ever merged? I'm running in exactly this issue, and it seems that your PR is closed without merging and the original issue is still open?

PR #3398, opened 2023-09-20
Mark bun: as external with platform=node

colinhacks:


evanw:
I'm not sure this specific direction (adding bun-specific stuff to esbuild's node target) makes sense because that target is for node, not bun.

Jarred-Sumner:
> I'm not sure this specific direction (adding bun-specific stuff to esbuild's node target) makes sense because that target is for node, not bun.

Bun is intended as a drop-in replacement for Node, so if anything in the `node` target does not work in Bun, it would be a bug in Bun (and not esbuild). A `bun` target would be more maintenance work for esbuild and would mean existing projects would need to change their esbuild configuration to fully support Bun.

ersinakinci:
Any update on this? Or does anyone have workarounds?

evanw:
I'm not sure if this will work for you, but FYI esbuild already has an API to mark things as external: https://esbuild.github.io/api/#external. Specifically `--external:bun:* --external:bun` will mark `bun` as external as well as anything starting with `bun:`.

PR #3424, opened 2023-10-05
update to Unicode 15.1

JLHwung:


JLHwung:
The Unicode 15.1 has marked U+30FB and U+FF65 as `Other_ID_Continue`

https://www.unicode.org/Public/15.1.0/ucd/PropList.txt

so they are now valid ES6+ identifiers again, though the browsers are not catching up yet.

@evanw Do you want to en-quote them in literal keys for browser compat? Or do you think we can move forward and just remove this test

https://github.com/evanw/esbuild/blob/02dae18a70b3d26dcb5b9599f02a9a36d88f809d/internal/js_parser/js_parser_test.go#L722-L734

Babel updated to Unicode 15.1 three weeks ago and no one ever complain that the katagana middle dot is not quoted anymore.

evanw:
I think they still have to be quoted. According to [this](https://www.unicode.org/L2/L2023/23160-utc176-properties-recs.pdf), Unicode 4.1 through Unicode 15 omitted these characters from ID_Continue by accident. Unicode 15.1 adds them back but they are still syntax errors in older browsers. For example, Firefox 118 throws `Uncaught SyntaxError: illegal character U+30FB` when trying to run the unquoted code.

JLHwung:
Thanks for the background update, it is very helpful.

On the other hand, U+30FB will be valid ID_Continue again, I think we should start accepting these two in IdentifierName, but quote them in literal object key for browser compatibility. Otherwise once browsers update to Unicode 15.1, it will seems like a bug when browser accepts U+30FB in ID_Continue but esbuild throws on them.

PR #3254, opened 2023-07-20
Update golang.org/x/sys v0.1.0 for CWE-266

madhavajay:
As per this PR:
https://github.com/evanw/esbuild/pull/3248

The minimum version that fixes the CWE is v0.1.0, will that work with the old version you are using?

https://pkg.go.dev/golang.org/x/sys?tab=versions
Screenshot 2023-07-20 at 12 56 58 pm

PR #3430, opened 2023-10-07
fix(sec): upgrade golang.org/x/sys to 0.1.0

ChengDaqi2023:
### What happened？
There are 1 security vulnerabilities found in golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8
- [CVE-2022-29526](https://www.oscs1024.com/hd/CVE-2022-29526)

### What did I do？
Upgrade golang.org/x/sys from v0.0.0-20220715151400-c0bba94af5f8 to 0.1.0 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### How can we automate the detection of these types of issues?
By using the [GitHub Actions](https://github.com/murphysecurity/actions) configurations provided by murphysec, we can conduct automatic code security checks in our CI pipeline.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,4 +4,4 @@ module github.com/evanw/esbuild
 // themselves for old OS versions. Please do not change this.
 go 1.13
 
-require golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8
+require golang.org/x/sys v0.1.0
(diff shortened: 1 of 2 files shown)


evanw:
Sorry but this PR prevents esbuild from building with Go 1.13, so I'm not going to take this PR. I'm keeping esbuild working with old versions of Go so that people can build esbuild for older versions of macOS (the Go team has removed support for older versions of macOS in newer Go versions).

File system access in esbuild is not security-sensitive so this alert you're getting doesn't seem relevant. I recommend disabling any alerts you're getting about this.

See also #3248 and #3254.

Issue #645, opened 2021-01-06
How to integrate react-refresh in esbuild to implement hmr ?


sessionboy:
I use esbuild to build my react application, no webpack, no rollup, only esbuild.It works well, everything looks great.
But I don't know how to implement hmr in esbuild.

```js
   service.build({
      platform: 'browser',
      format: 'esm',
      splitting: true,
      bundle: true,
      loader: {
        '.js': 'jsx',
        '.png': 'file',
        '.svg': 'file',
        '.jpg': 'file'
      },
      minify: false,
      entryPoints: [
        path.join(paths.appSrc, '/layout/entry-client.tsx')
      ],
      outdir: path.join(paths.appBuild, 'client'),
      sourcemap: true,
      metafile: path.join(paths.appBuild, 'client-metafile.json'),
      define: {
        'process.env.NODE_ENV': JSON.stringify('development'),
        'process.browser': "true",
      }
    })   
```
I checked the hmr implementation of vite. But the difference is that vite only injects the HMR code when the module is requested, not during the build.So this is easy to implement.

I used code splitting.

I need to inject hmr code during the first build process. Re-execute the build when the code is updated, and update the code dynamically.

This is not a simple matter.

I guess esbuild plugin should be able to implement it, but I need some guidance. How can i implement it ?

evanw:
Sorry, esbuild isn't designed for HMR. It's possible to use esbuild's transform API as part of another HMR tool (e.g. what Vite does) but this isn't something that esbuild's build API supports. I'm not sure if it can be done with an esbuild plugin because esbuild's plugin API deliberately doesn't let you substitute your own linker. I haven't ever implemented HMR before but I believe HMR requires a different type of linker. For example the [react-refresh documentation](https://github.com/facebook/react/issues/16604#issuecomment-528663101) says that integrating it requires your bundler to support HMR.

sessionboy:
> esbuild isn't designed for HMR.
I know this, I think hmr should be an additional feature of esbuild.
I don't want to use webpack or other bundler, I think the future will be no-bundler.
Many people expect esbuild to implement hmr. If esbuild implements hmr, then the no-bundler goal will be achieved immediately, this will make web go into the future.
Snowpack proposed the [esm-hmr spec](https://github.com/snowpackjs/esm-hmr), maybe this can help you implement hmr.

What's bad is that react-refresh currently only supports babel's transform. I haven't found any documentation or case of using esbuild's transform to implement react-refresh.

[...]

PR #3449, opened 2023-10-13
Add Fast Refresh support

ArnaudBarre:
I know this was previously [considered as out of scope](https://github.com/evanw/esbuild/issues/645), but after the implementation of JSX dev runtime, this is the last piece required to make esbuild the perfect tool for building React SPAs.

It would allow to get an instant dev-server start without the overhead of an external plugin & with HMR support (what I will work on it this lands!)

This is my first time writing go, I can iterate on it in the coming weeks!


internal/js_parser/js_parser.go
```diff
@@ -1733,6 +1741,297 @@ func (p *parser) importJSXSymbol(loc logger.Loc, jsx JSXImport) js_ast.Expr {
...
+	"useDebugValue":       true,
+}
+
+type reactRefreshHandle struct {
+	handle ast.Ref
+	id     string
+}
+
+func (p *parser) reactRefreshRegisterHandle(target ast.Ref, id string) js_ast.Stmt {
+	var handleName = "_c"
```
rtsao, line 1776: I think it would be good to handle scope collisions (like the SWC transform):

```js
const _c = 'collision';
const _c1 = 'collision1';
export function Component() {
  return <div></div>;
}

// becomes:

import { jsxDEV as o } from "react/jsx-dev-runtime";
const _c = 'collision';
const _c1 = 'collision1';
export function Component() {
    return /*#__PURE__*/ o("div", {}, void 0, false, {
        fileName: "input.jsx",
        lineNumber: 4,
        columnNumber: 10
    }, this);
}
_c2 = Component;
var _c2;
$RefreshReg$(_c2, "Component");
```

In the case of SWC, the integer suffix is incremented until a unique identifier is found in scope.

ZinoKader:
Really interested in using this if it lands. What's the maintainers' stance on adding fast refresh support?

internal/js_parser/js_parser.go
brandonxiang: @ArnaudBarre any update?

ArnaudBarre: Nothing plan for the moment until I get more positive feedback on this from either maintainers or the community. 16 reactions for something that would allow instant dev server for SPA in dev (because not other tools can do bundling with fast refresh without plugin overhead) feels not worth it publishing this as a fork

ArnaudBarre:
Rolldown will be stable in a near future and is built on top of OXC that has Fast Refresh support. I think this is a better way forward for building a fast React toolchain.
Thanks you @evanw for all you did, esbuild is what inspire to work on making JS tooling faster.

PR #3491, opened 2023-11-14
fix #3490: parse `import type from from`

### LINKED ISSUE #3490: Cannot parse `import type from from "foo"`
- [esbuild playground](https://esbuild.github.io/try/#dAAwLjE5LjUALS1sb2FkZXI9dHMAaW1wb3J0IHR5cGUgZnJvbSBmcm9tICJmb28i)
- related issue: https://github.com/microsoft/TypeScript/issues/56376

magic-akari:
- fix: #3490


evanw:
Thanks for the report and for the suggested fix. Heads up that I'm planning to fix this a bit differently to avoid backtracking.

PR #3520, opened 2023-11-30
add Safari compat data for arbitrary module namespace names

sapphi-red:
This PR adds Safari compat data for "arbitrary module namespace names" feature.


evanw:
Thanks for finding this. Hopefully it will be added to https://github.com/mdn/browser-compat-data/issues/18152 in the future but this is great for now.

PR #3534, opened 2023-12-09
pkg/api: fix Rebuild always write to output files

shuLhan:
On the first Rebuild, the oldHashes is empty, so the condition that checking between old and new hash will always false which makes the output files always written.

This changes fix this issue by moving the content comparison only if no old hash found.

### DIFF
--- a/pkg/api/api_impl.go
+++ b/pkg/api/api_impl.go
@@ -1578,7 +1578,14 @@ func rebuildImpl(args rebuildArgs, oldHashes map[string]string) (rebuildState, m
 							defer waitGroup.Done()
 							fs.BeforeFileOpen()
 							defer fs.AfterFileClose()
-							if oldHash, ok := oldHashes[result.AbsPath]; ok && oldHash == newHashes[result.AbsPath] {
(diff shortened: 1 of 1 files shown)


evanw:
You are trying to change something that is a deliberate behavior of esbuild. When you run the build, the build outputs are supposed to be newer than build inputs. Otherwise the build could look stale, which it's not supposed to be after a build. You have also not described your problem or given any context for why you are doing this. I'm closing this PR for both of these reasons.

shuLhan:
Hi @evanw, thanks for reviewing. Sorry for not describing the issue in more detail, I though the code was quite clear.

So, the issue was like these.

Given an input file A and build output X with modified time T, everytime we run esbuild, either one time or using api Context, the file X always changes with new modified time T2.

First build,

  A => X T

Second build,

  A => X T2

and so on.

This become an issue if the output file is committed to VCS, due to pipeline or some watch involve in the background while in development mode.

The expected output is if the content of A does not changes and X file is exist, esbuild should not create new file X with new modified time.

Thats what this PR trying to do.
Let me know if I need to create an issue first, so we can discuss this further.
Maybe the current fix is not the correct solution.

PR #3539, opened 2023-12-12
Provide API to create a custom esbuild CLI with plugins

edewit:
This is very important as it allows to create create a CLI version of esbuild with a custom set of Go Plugin (in our case we needed scss). Without it we have to fork esbuild and add it which is not very nice.

Signed-off-by: Erik Jan de Wit <erikjan.dewit@gmail.com>

### DIFF
--- a/pkg/cli/cli_impl.go
+++ b/pkg/cli/cli_impl.go
@@ -1127,7 +1127,7 @@ func addAnalyzePlugin(buildOptions *api.BuildOptions, analyze analyzeMode, osArg
 	buildOptions.Metafile = true
 }
 
-func runImpl(osArgs []string) int {
+func runImpl(osArgs []string, plugins []api.Plugin) int {
 	// Special-case running a server
 	for _, arg := range osArgs {
 		if arg == "--serve" ||
@@ -1278,6 +1278,8 @@ func runImpl(osArgs []string) int {
 			}
 		}
 
+		buildOptions.Plugins = plugins
+		// buildOptions.Plugins = append(buildOptions.Plugins, scssPlugin)
 		// Handle post-build actions with a plugin so they also work in watch mode
 		buildOptions.Plugins = append(buildOptions.Plugins, api.Plugin{
 			Name: "PostBuildActions",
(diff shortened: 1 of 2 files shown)


ia3andy:
Thanks for creating this PR @edewit, this will help a lot!

pkg/cli/cli_impl.go
```diff
@@ -1278,6 +1278,8 @@ func runImpl(osArgs []string) int {
 			}
 		}
 
+		buildOptions.Plugins = plugins
+		// buildOptions.Plugins = append(buildOptions.Plugins, scssPlugin)
```
ia3andy, line 1282: @edewit this can be removed right?

ia3andy (changes requested):

pkg/cli/cli_impl.go
ia3andy, line 1282:
```suggestion
```

ia3andy:
Hey @evanw, 

Could you consider getting this in, it would allow us to package our[ own CLI](https://github.com/mvnpm/esbuild) with plugins (scss) using normal GO import instead of forking the whole repository.

Cheers!

ia3andy:
@evanw ping!

Rattlyy:
bump :)

ia3andy:
@evanw any reason not to get this in?

evanw:
Hey, thanks for the bump. There's no reason to not get this in. I think something like this makes sense and I'm supportive of it. The only reason it hasn't gotten in yet is that I haven't put aside the time yet to do the PR (i.e. think about the approach, make any changes, land it, and ship it).

ia3andy:
Hey @evanw, that's a very good news! Thanks.

It seems this very basic PR is already enough for our use-case at least, do you have something else in mind?

evanw:
There are several approaches to this. In general `RunWith*` functions are not a great approach to extending APIs with many optional parameters, as there could be a combinatorial explosion of options (and therefore of functions). That's why esbuild's API generally takes structs instead of individual parameters. Other approaches could be to require you to separate parsing the API options from using them in this scenario, or to provide a callback to modify the CLI options after they are parsed but before they are used. However, it could also be ok to take the approach you're proposing as plugins are somewhat of a special case (one of the few API parameters that can't be configured with the CLI).

ia3andy:
Thanks!

PR #3560, opened 2023-12-21
feat: download tgz using the current environment's npm registry

Beace:
Sometimes we can only use a private npm registry for downloading. Can we use this method to modify the download source of esbuild？


evanw:
The code you are looking at is the fallback for when `npm` isn't installed. Before this code runs, esbuild runs `npm install` which is supposed to be the way this case is handled:

https://github.com/evanw/esbuild/blob/2aa166b623a0db059409e7ccf02ebc084fcb3abd/lib/npm/node-install.ts#L118

There are many complicated parts about how npm installs packages and it shouldn't be the job of esbuild's installer to replicate all of that complexity. Rather, esbuild should be delegating to npm to handle this stuff instead. If something isn't working about esbuild's installer, then the correct first step is to file a bug report that describes the problem and provides a way to reproduce it, not to propose a change without any context like this.

evanw:
Closing due to lack of a follow-up reply.

PR #3611, opened 2024-01-26
work around api deprecations in deno 1.40.x (#3609)

### LINKED ISSUE #3609: Deno: Usage of deprecated API with >= Deno 1.40
With the recent Deno 1.40 release some older APIs [were marked as deprecated](https://deno.com/blog/v1.40#deprecations-stabilizations-and-removals). Some of those are used within the deno esbuild adapter and print warning messages to the terminal.

```sh
Stack trace:
  at https://deno.land/x/esbuild@v0.19.11/mod.js:1888:45
  at eventLoopTick (ext:core/01_core.js:59:7)
  at async file:///Users/marvinh/dev/test/esbuild-deprecations/foo.ts:3:1

hint: Use `Deno.stderr` instance methods instead.
hint: It appears this API is used by a remote dependency. Try upgrading to the latest version of that dependency.

warning: Use of deprecated "Deno.isatty()" API. This API will be removed in Deno 2.

Stack trace:
  at https://deno.land/x/esbuild@v0.19.11/mod.js:1888:26
  at eventLoopTick (ext:core/01_core.js:59:7)
  at async file:///Users/marvinh/dev/test/esbuild-deprecations/foo.ts:3:1

hint: Use `Deno.stdin.isTerminal()`, `Deno.stdout.isTerminal()` or `Deno.stderr.isTerminal()` instead.
hint: It appears this API is used by a remote dependency. Try upgrading to the latest version of that dependency.

warning: Use of deprecated "Deno.run()" API. This API will be removed in Deno 2.

Stack trace:
  at https://deno.land/x/esbuild@v0.19.11/mod.js:1889:26
  at eventLoopTick (ext:core/01_core.js:59:7)
  at async file:///Users/marvinh/dev/test/esbuild-deprecations/foo.ts:3:1

hint: Use "Deno.Command()" API instead.
hint: It appears this API is used by a remote dependency. Try upgrading to the latest version of that dependency.
```

## Steps to reproduce

1. Upgrade Deno to 1.40.1 `deno upgrade`
2. Create a file `foo.ts` with these contents:
  ```ts
  import * as esbuild from "https://deno.land/x/esbuild@v0.19.11/mod.js";
  
  await esbuild.build({
    entryPoints: ["foo"],
    bundle: true,
    plugins: [
      {
        name: "foo-entry",
        setup(build) {
          build.onResolve({ filter: /foo/ }, (args) => {
            return { path: args.path, namespace: "foo" };
          });
  
          build.onLoad({ filter: /.*/, namespace: "foo" }, () => {
            return { contents: "export default 42;", loader: "ts" };
          });
        },
      },
    ],
  });
  
  esbuild.stop();
  ```
  3. Run `deno run -A foo.ts`

evanw:
Deno 1.40.0 introduced run-time warnings about certain APIs that esbuild uses. With this PR, esbuild will work around these run-time warnings by using newer APIs if they are present and falling back to the original APIs otherwise. This should avoid the warnings without breaking compatibility with older versions of Deno.

Unfortunately, doing this introduces a breaking change. The newer child process APIs lack a way to synchronously terminate esbuild's child process, so calling `esbuild.stop()` from within a Deno test is no longer sufficient to prevent Deno from failing a test that uses esbuild's API (Deno fails tests that create a child process without killing it before the test ends). To work around this, esbuild's `stop()` function has been changed to return a promise, and you now have to change `esbuild.stop()` to `await esbuild.stop()` in all of your Deno tests.

Fixes #3609

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+* Work around API deprecations in Deno 1.40.x ([#3609](https://github.com/evanw/esbuild/issues/3609))
+
+    Deno 1.40.0 introduced run-time warnings about certain APIs that esbuild uses. With this release, esbuild will work around these run-time warnings by using other APIs if they are present, and falling back to the original APIs otherwise. This should avoid the warnings without breaking compatibility with older versions of Deno.
+
+    Unfortunately, doing this introduces a breaking change. The other child process APIs seem to lack a way to synchronously terminate esbuild's child process, which causes Deno to now fail tests with a confusing error about a thing called `op_spawn_wait` in Deno's internals. To work around this, esbuild's `stop()` function has been changed to return a promise, and you now have to change `esbuild.stop()` to `await esbuild.stop()` in all of your Deno tests.
+
 ## 0.19.12
 
 * The "preserve" JSX mode now preserves JSX text verbatim ([#3605](https://github.com/evanw/esbuild/issues/3605))
--- a/lib/deno/mod.ts
+++ b/lib/deno/mod.ts
@@ -191,6 +191,8 @@ type SpawnFn = (cmd: string, options: {
   read(): Promise<Uint8Array | null>
   close(): Promise<void> | void
   status(): Promise<{ code: number }>
+  unref(): void
+  ref(): void
 }
 
 // Deno ≥1.40
@@ -232,6 +234,8 @@ const spawnNew: SpawnFn = (cmd, { args, stdin, stdout, stderr }) => {
       await child.status
     },
     status: () => child.status,
+    unref: () => child.unref(),
+    ref: () => child.ref(),
   }
 }
 
@@ -272,6 +276,8 @@ const spawnOld: SpawnFn = (cmd, { args, stdin, stdout, stderr }) => {
       child.close()
     },
     status: () => child.status(),
+    unref: () => { },
+    ref: () => { },
   }
 }
 
@@ -327,12 +333,20 @@ const ensureServiceIsRunning = (): Promise<Service> => {
       })
       readMoreStdout()
 
+      let refCount = 0
+      child.unref() // Allow Deno to exit when esbuild is running
+
+      const refs: common.Refs = {
+        ref() { if (++refCount === 1) child.ref(); },
+        unref() { if (--refCount === 0) child.unref(); },
+      }
+
       return {
         build: (options: types.BuildOptions) =>
           new Promise<types.BuildResult>((resolve, reject) => {
             service.buildOrContext({
               callName: 'build',
-              refs: null,
+              refs,
               options,
               isTTY,
               defaultWD,
@@ -344,7 +358,7 @@ const ensureServiceIsRunning = (): Promise<Service> => {
           new Promise<types.BuildContext>((resolve, reject) =>
             service.buildOrContext({
               callName: 'context',
-              refs: null,
+              refs,
               options,
               isTTY,
               defaultWD,
@@ -355,7 +369,7 @@ const ensureServiceIsRunning = (): Promise<Service> => {
           new Promise<types.TransformResult>((resolve, reject) =>
             service.transform({
               callName: 'transform',
-              refs: null,
+              refs,
               input,
               options: options || {},
               isTTY,
@@ -388,7 +402,7 @@ const ensureServiceIsRunning = (): Promise<Service> => {
           new Promise((resolve, reject) =>
             service.formatMessages({
               callName: 'formatMessages',
-              refs: null,
+              refs,
               messages,
               options,
               callback: (err, res) => err ? reject(err) : resolve(res!),
@@ -398,7 +412,7 @@ const ensureServiceIsRunning = (): Promise<Service> => {
           new Promise((resolve, reject) =>
             service.analyzeMetafile({
               callName: 'analyzeMetafile',
-              refs: null,
+              refs,
               metafile: typeof metafile === 'string' ? metafile : JSON.stringify(metafile),
               options,
               callback: (err, res) => err ? reject(err) : resolve(res!),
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+* Work around API deprecations in Deno 1.40.x ([#3609](https://github.com/evanw/esbuild/issues/3609))
+
+    Deno 1.40.0 introduced run-time warnings about certain APIs that esbuild uses. With this release, esbuild will work around these run-time warnings by using other APIs if they are present, and falling back to the original APIs otherwise. This should avoid the warnings without breaking compatibility with older versions of Deno.
+
+    Unfortunately, doing this introduces a breaking change. The other child process APIs seem to lack a way to synchronously terminate esbuild's child process, which causes Deno to now fail tests with a confusing error about a thing called `op_spawn_wait` in Deno's internals. To work around this, esbuild's `stop()` function has been changed to return a promise, and you now have to change `esbuild.stop()` to `await esbuild.stop()` in all of your Deno tests.
```
evanw, line 9: It's unfortunate that this is a breaking change. I think this is significant enough to require a breaking change release for esbuild.

lib/deno/mod.ts
```diff
@@ -178,63 +178,142 @@ interface Service {
...
+      write: bytes => writer.write(bytes),
+      close: () => writer.close(),
+    },
+    stdout: {
+      read: () => reader.read().then(x => x.value || null),
+      close: () => reader.cancel(),
+    },
+    close: async () => {
+      // Note: This can throw with EPERM on Windows (happens in GitHub Actions)
+      child.kill()
```
evanw, line 224: Trying to kill the child process on Windows throws `PermissionDenied: Access is denied. (os error 5)`. But not killing the process on Windows causes Deno to fail tests with `1 async operation to op_spawn_wait was started in this test, but never completed`. So it seems like maybe there's no way to get tests to pass on Windows with Deno's new APIs? I'm not sure how to proceed here. Maybe this is a bug in Deno itself?

dbushell: in `stopService` you may need to await `stdin.close()` since `writer.close()` returns a promise? (same for `stdout` maybe though [Deno docs](https://deno.land/api@v1.40.1?s=Deno.Command) say only the "stdin `WritableStream` needs to be closed manually") - sorry I don't have Windows on hand to test.

I think the first test fail because the child process is spawned within `fn` but closed outside in `finally { esbuildNative.stop() }`. Seems like Deno test is being overly critical/protective there as it's working as intended. I don't know enough to resolve that issue sadly.

lucacasonato: This looks like a race between you calling `stdin.close()` and `child.kill()`. The esbuild binary probably exits by itself when you close it's stdin to prevent runaway processes. Thus, when you call `stdin.close()` the process may have already exited by the time you call `child.kill()` (and then that call fails because the process is already gone).

You can rely on the fact that esbuild shuts down when you close `stdin`: just make sure you close `stdin` during shutdown, and then just wait for the process to actually have exited with `await child.status`.

CHANGELOG.md
lucacasonato: > which causes Deno to now fail tests with a confusing error about a thing called `op_spawn_wait` in Deno's internals

Unfortunately there is an error mapping missing in the internals which is why you got this error. It should have said:

```
An async operation to wait for a subprocess to exit was started in this test, but never completed. This is often caused by not awaiting the result of a `Deno.Process#status` call.
```

Will be fixed in Deno 1.40.2

lucacasonato: Also the reason we don't have synchronous process termination, is because subprocess termination is by definition not synchronous. When you send a signal with the `kill()` syscall, the process may intercept it and decide to not exit. If it doesn't exit immediately, you can't just hang the parent process forever.

What the old Deno APIs, and Node's API do to emulate synchronous exit is to just "forget" about the process after `close` / `kill`, which means that if the subprocess doesn't actually shut down, you have a zombie process that will get orphaned when the parent process exits. This new API avoids this footgun - it's modeled after the Rust API for subprocesses.

lib/shared/types.ts
```diff
@@ -673,4 +673,4 @@ export let version: string
 // Unlike node, Deno lacks the necessary APIs to clean up child processes
 // automatically. You must manually call stop() in Deno when you're done
 // using esbuild or Deno will continue running forever.
```
lucacasonato, line 675: Also FYI, the new API has `Deno.ChildProcess#ref` and `Deno.ChildProcess#unref` methods that should let most users avoid calling this API (unless they are running in a test with sanitization, like you are here).

I suggest you call `process.unref()` like you do in Node, and then most users don't have to call this method at all anymore :)

evanw: Yes, that comment is old. Thanks for mentioning it. I just became aware of `unref` while doing this yesterday and still need to look into it.

Although thinking about the bigger picture, I suppose perhaps this whole `deno.land/x/esbuild` library is unnecessary now that Deno has since added support for npm packages. It feels like I should just deprecate and abandon it, and then tell people to import from `npm:esbuild` instead.

lucacasonato: That is certainly an option - however in that case a user may still want to support the explicit shutdown of the underlying esbuild process.

evanw: Yes, that's true. Calling `esbuild.stop()` from `npm:esbuild` currently prints the following:

```
Warning: Not implemented: ChildProcess.prototype.disconnect
```

I'm not sure why though as esbuild never calls `disconnect` and there is no stack trace associated with the warning.

lib/deno/mod.ts
evanw: That seems to have worked. Thanks for the suggestion!

lib/shared/types.ts
evanw: Note: Calling `unref` seems to have caused a regression: #3682. I'm removing the call to `unref` to fix the regression: #3685.

PR #3647, opened 2024-02-13
fix: types should be read-only

Zamiell:
Today I discovered that ESBuild would not take in my immutable `entryPoints` array, which is a bug. So, I fixed it in this PR, as well as the surrounding data structures that were also incorrectly marked as mutable.

More info: In typescript libraries it is idiomatic to take in read-only data structures. That way, end-users do not have to cast their immutable data structures to be mutable simply to satisfy the TypeScript compiler (which is both ugly and reduces safety in using the library).

In case it isn't obvious, this PR is backwards compatible in that end-users can now use either mutable OR immutable arrays.

### DIFF
--- a/lib/shared/types.ts
+++ b/lib/shared/types.ts
@@ -20,9 +20,9 @@ interface CommonOptions {
   /** Documentation: https://esbuild.github.io/api/#global-name */
   globalName?: string
   /** Documentation: https://esbuild.github.io/api/#target */
-  target?: string | string[]
+  target?: string | readonly string[]
   /** Documentation: https://esbuild.github.io/api/#supported */
-  supported?: Record<string, boolean>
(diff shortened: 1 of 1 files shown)


evanw:
Sorry but esbuild's build options object is intentionally mutable. Quoting from [esbuild's documentation](https://esbuild.github.io/plugins/#build-options):

> Plugins can access the initial build options from within the setup method. This lets you inspect how the build is configured _as well as modify the build options before the build starts._

Changing the types to be immutable is incorrect. Note that this means casting immutable data structures to be mutable to satisfy the TypeScript compiler is also incorrect, because it could mean that your "immutable" data structures end up being mutated by esbuild plugins.

I'm closing this PR as I do not plan to make this change.

Issue #3666, opened 2024-02-23
add an optional cache and `exclude` option to the `onResolve` plugin hook


sod:
As the angular-cli right now fully commits to esbuild, the `onResolve` hook reveals itself as a pretty stark performance hinderance to the otherwise astonishing speed of esbuild.

In my mind this is how the API could look like to improve:

```ts
build.onResolve({ filter: /^[^./]/, exclude: /^(@angular|rxjs)\// } }, async (args) => {
    if (/* some condition */) {
        return { path: args.path, external: true, cachable: true };
    }

    return { path: args.path, cachable: true };
});
```

1. The `exclude` option could be a way around the missing lookahead feature of go-regex. To stamp out imports we don't want to look at
2. The opt-in `cachable: true` could tell esbuild to not ask for the same information again, which would be nice for watcher performance

Would you be open to this idea and if yes, would you appreciate a PR contribution?

sod:
Maybe the cachable is a bit much. But the `exclude` option could go a long way.

PR #3667, opened 2024-02-24
feat #3666 add optional exclude option for plugin `onLoad` and `onResolve` hooks

sod:
This allows plugins that are very broadly applied to user code, to ignore files and folders it knows not to touch, like `node_modules`, static files or shared folders.

As the cost of calling a plugin is not 0, especially `onResolve` hooks, this could help improve performance of plugins.

See #3666 

used like:

```js
const myPlugin = {
  name: 'my-plugin',
  setup(build) {
    build.onResolve({ filter: /.js$/, exclude: /^(static|my-lib)\// }, args => {
    })
    build.onLoad({ filter: /.js$/, exclude: /\/node_modules\// }, args => {
    })
  }
}
```

the exclude is optional and the filter-only syntax stil works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,22 @@
     })(Foo || {});
     ```
 
+* Add `exclude?: RegExp` option to `onLoad` on `onResolve` hooks
+
+    Allows plugins to stamp out files and folders to not execute the hook.
+
+    ```js
+    const myPlugin = {
+      name: 'my-plugin',
+      setup(build) {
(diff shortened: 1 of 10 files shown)


sod:
@evanw whats your stance on this? Should I rebase, change something or drop? :)

sod:
I consider no response as a decline. For the sake of not cluttering esbuild PR tab, I close this.

DamienCassou:
I would need such a feature.

Issue #938, opened 2021-03-09
Bundling fails without read permission to parent directories


mitschabaude:
Hi, I am trying to use esbuild on an Ubuntu VPS where my `$HOME` directory is `/usr/home/<USER>`, but I don't have read permission to the parent folder `/usr/home`.

This causes bundling to fail with `Cannot read directory ... permission denied`

The error can be reproduced on any machine by mimicking the permissions layout:

```sh
sudo mkdir forbidden && sudo mkdir forbidden/repo
sudo chmod o-rw forbidden && sudo chmod a+w forbidden/repo
cd forbidden/repo
npm init -y
npm i esbuild react
echo 'import "react"' | npx esbuild --bundle
```
Output:
```sh
 > error: Cannot read directory "..": permission denied

 > <stdin>:1:7: error: Could not resolve "react" (mark it as external to exclude it from the bundle)
```

From my (ignorant) viewpoint it seems unnecessary to read anything above the package root when resolving node modules, so I thought esbuild could possibly prevent tripping up in this situation. (E.g. Webpack bundles without errors in the same situation.)

heyheyhello:
@kzc Don't use --outfile=`tty` and it'll be silent.

```
$ esbuild --version
0.9.0
$ esbuild 0.js --minify
console.log(2);
```

kzc:
@heyheyhello I prefaced my comment with:

> Here's a contrived example for demonstration purposes:

heyheyhello:
Yeah and your contrived example is great to show the `../../../../../` weirdness ✅ but if you care about esbuild's stdout for unix-pipe reasons then it should be good about that by default?

kzc:
It was a side note based on the previous command. Here's a simpler example:

Compare the length of this command:

```
$ esbuild 0.js --outfile=out.js

  out.js  24b 

⚡ Done in 4ms
``

to this one:
``
$ esbuild 0.js --outfile=out.js --log-level=error
$ 
```

Once anyone has used esbuild once they never need to see the verbose output again and will have to type `--log-level=error` in all their build scripts for future projects to not clutter the console or their log files.  Is it a big deal?  No, it's not.  It's just a minor annoyance. I just prefer more succinct commands and explicit opt-ins for nonessential information.

evanw:
> This causes bundling to fail with `Cannot read directory ... permission denied`

[...]

PR #3671, opened 2024-02-28
fix: use the built in way for Go to check permissions (fix for Windows and cross platform)

arizchang1:
instead of explicitly checking EACCES and EPERM, Go has a built in way to check permissions (Windows fix and is cross-platform)

fixes the Windows issue reported by cuzfrog: https://github.com/evanw/esbuild/issues/938

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -3,6 +3,7 @@ package resolver
 import (
 	"errors"
 	"fmt"
+	fs_real "io/fs"
 	"path"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


evanw:
Do you have a way to reproduce the issue? Ideally changes like this would have test coverage. This is especially important because I don't have a Windows machine and I never use Windows myself.

arizchang1:
Yes I reproduced it following this https://github.com/evanw/esbuild/issues/938#issuecomment-1676923150
essentially we can make a dummy directory structure like test/test2/test3/test.ts, where the current user has access to test3 but not test2, and then run `esbuild.exe --bundle test.ts` from test3. I think it's the same issue that was fixed in that thread, but this is just a better cross-platform way to check permissions

SoftexNL:
This is unfortunately a problem for remix/vite also, can't be built on a windows server either

ulspider:
@arizchang1 or @evanw, is it possible to test on linux this fix in order to merge this fix ? I can test this fix on windows if you want

PR #3674, opened 2024-03-01
Add OpenBSD arm64

### LINKED ISSUE #3665: OpenBSD arm64 support
I'd love an arm64 version for OpenBSD, looks like there is an arm64 version for FreeBSD so hopefully the effort is minimal?

ikmckenz:
Resolves #3665 

Tested working on OpenBSD on an M1 Apple chip.


npm/esbuild/package.json
rnmeow, line 1: seems like an extra indent tab was added

ikmckenz: I just match the existing indentation of this file

rnmeow: > I just match the existing indentation of this file

The line you added line has two tabs while there should be only one.

image

ikmckenz: Looks like because the `.editorconfig` file only currently includes js and ts files for spaces, with the rest using tabs, my editor inserted two tabs, instead of four spaces. I edited that config file, then reformatted my offending file and confirmed it replaced the tabs with spaces, and all looks good now. 
I think the reason I didn't notice it is my editor and GitHub were showing tabs to equal two spaces of width, perhaps you have different settings configured for displaying tabs. Thank you!

rnmeow (approved):

ikmckenz:
Updated for the 0.21.4 release

PR #3679, opened 2024-03-05
fix #2388: allow consuming types without dom types

### LINKED ISSUE #2388: Cannot find namespace 'WebAssembly'
During the compilation using `tsc` with the esbuild API I encountered following error:
```
node_modules/esbuild/lib/main.d.ts:584:16 - error TS2503: Cannot find namespace 'WebAssembly'.
```

I had to `skipLibCheck` in TypeScript so that I can compile without any errors.

image

remcohaszing:
Currently the TypeScript types for esbuild depend on the DOM types. However, esbuild itself does not depend on the DOM.

This change uses type augmentation to make the type error go away, while keeping backwards support for use with the DOM types.

Closes #2388

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,10 +107,13 @@ test-old-ts: platform-neutral | require/old-ts/node_modules
 node-unref-tests: | scripts/node_modules
 	node scripts/node-unref-tests.js
 
-lib-typecheck: lib-typecheck-node lib-typecheck-deno
+lib-typecheck: lib-typecheck-node lib-typecheck-node-nolib lib-typecheck-deno
 
 lib-typecheck-node: | lib/node_modules
 	cd lib && node_modules/.bin/tsc -noEmit -p tsconfig.json
+
+lib-typecheck-node-nolib: | lib/node_modules
+	cd lib && node_modules/.bin/tsc -noEmit -p tsconfig-nolib.json
 
 lib-typecheck-deno: lib/deno/lib.deno.d.ts | lib/node_modules
 	cd lib && node_modules/.bin/tsc -noEmit -p tsconfig-deno.json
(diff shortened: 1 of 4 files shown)


evanw:
Ok, I can give this a try, but I reserve the right to revert this if it causes other problems. It looks like avoiding `dom` types also requires `URL` to be defined, so I'll add that too.

Seems kind of weird for using esbuild to be polluting the global type scope, but I guess that's just how TypeScript works. It's too bad TypeScript doesn't have some sort of `if typeof WebAssembly !== 'undefined'` for types.

remcohaszing:
> I reserve the right to revert this if it causes other problems

Of course you do! You are the maintainer :heart: 

> It looks like avoiding `dom` types also requires `URL` to be defined, so I'll add that too.

Nice catch! This is less apparent, as people often use the `dom` types or `@types/node`, both if which specify URL.

> Seems kind of weird for using esbuild to be polluting the global type scope, but I guess that's just how TypeScript works. It's too bad TypeScript doesn't have some sort of `if typeof WebAssembly !== 'undefined'` for types.

You can, but it’s quite hacky. ([playground link](https://www.typescriptlang.org/play?noLib=true&filetype=d.ts#code/C4TwDgpgBAYghgGwQIzgYwNZQLxQHYQBuEATgFBkD0lUAlnsKQGbrQCqASgDJQDeVNKEKgALAPYBnYHjgBbCAC4oUkvQDmAbgFQAvhWpQAAsAkBaWmrxiSEMqEhQAkg1IyEnHrjh4QUCAA9GPAATCSgPKAB+WEQUdCwlDzIgA))

```ts
type Fallback = never

// @ts-ignore
type InternalURL = any extends URL ? Fallback : URL
```

This only works from handwritten type definition files, not from TypeScript source files.

PR #3683, opened 2024-03-06
fix some comments

cuinix:


internal/js_parser/ts_parser_test.go
```diff
@@ -2739,7 +2739,7 @@ func TestTSImportEqualsInNamespace(t *testing.T) {
 func TestTSTypeOnlyImport(t *testing.T) {
 	expectPrintedTS(t, "import type foo from 'bar'; x", "x;\n")
 	expectPrintedTS(t, "import type foo from 'bar'\nx", "x;\n")
-	expectPrintedTS(t, "import type from from 'bar'; x", "x;\n")
```
lydell, line 2742: It is valid to name a type `from` and import it – this was intentional: https://github.com/evanw/esbuild/commit/08b4607d4cc83768c2fb7677ae90ae974c4658a3

(Also, this is not a comment – it’s a test.)

Issue #3692, opened 2024-03-11
Enable picking truly random port in serve mode


ptrdom:
Currently there is no way to tell esbuild to pick a truly random port in serve mode - if `port` is unset or set to `0` then port in `8000-8009` range gets picked up. Ideally if `port` is set to `0` then this would just be passed to Go `listen` implementation and it would pick a free random port.

SillyHooman:
So I just have to modify this right:
`if serveOptions.Port == 0 {
		// Default to picking a "800X" port
			if result, err := net.Listen(network, net.JoinHostPort(host, ":0")); err == nil {
				listener = result
				break
		}
	}`
This will allow us to use random port can anybody check if my understanding is true

ptrdom:
It is somewhat tricky - I think we want to preserve default behavior of keeping the port in familiar range of `8000-8009`, but also allow passing `0`. Problem here comes mainly from the fact that Go does not have nullable variables, so `0` expressed both user setting port to `0` and user not setting the port. The code likely needs a separate variable to track whether user has set the port or not.

SillyHooman:
Thanks for the feedback, should i pr my changes or leave it
PS: am a beginner

ptrdom:
Not sure, @evanw could you comment on this issue? Need your confirmation before exploring solutions.

SillyHooman:
Hey yo i did this pr #3693 and i also tested locally which was confusing to setup and test but it has the random functionality

evanw:
I can do this but making the port nullable in the Go API is a breaking change, so it'll have to wait until a breaking change release to do it. Adding the `breaking` label for now to track this.

SillyHooman:
K

evanw:
> Ideally if `port` is set to `0` then this would just be passed to Go `listen` implementation and it would pick a free random port.

I'm not sure that this is necessary. What is the reason to have a random port from the full range instead of a random port from a narrow range? I'm not sure why it would be ideal for esbuild to do that. What happens now because esbuild doesn't do that?

ptrdom:
> > Ideally if `port` is set to `0` then this would just be passed to Go `listen` implementation and it would pick a free random port.
> 
> I'm not sure that this is necessary. What is the reason to have a random port from the full range instead of a random port from a narrow range? I'm not sure why it would be ideal for esbuild to do that. What happens now because esbuild doesn't do that?

[...]

PR #3693, opened 2024-03-12
Random Port #3692

SillyHarish:
Now when port is not specified or is set to 0, the serve method will choose a random port instead of just between 8000 - 8009


SillyHooman:
PS: I'm new to this whole GitHub Thing

Issue #3682, opened 2024-03-06
esbuild inside Deno.test() leaks resources


AlexJeffcott:
When using esbuild inside a `Deno.test()` esbuild leaks at least one pending resource which Deno reports as a test error. Calling esbuild's `await stop()` does not work.

Deno version: 1.41.1 (latest)

esbuild version: 0.21.1 (latest)

A minimal reproducible example is this test which I expect to pass:

```Typescript
#!/usr/bin/env -S deno test -A

import { assertEquals } from 'https://deno.land/std@0.218.0/assert/mod.ts'
import * as esbuild from 'https://deno.land/x/esbuild@v0.20.1/mod.js'

Deno.test('should transform', async () => {
  const result = await esbuild.transform('const test: boolean = true', { loader: 'ts' })
  assertEquals(result.code, 'const test = true;\n')
  await esbuild.stop()
})

Deno.test('should transform again', async () => {
  const result = await esbuild.transform('const test: boolean = true', { loader: 'ts' })
  assertEquals(result.code, 'const test = true;\n')
  await esbuild.stop()
})
```

Will output:

```shell
error: Promise resolution is still pending but the event loop has already resolved.
```

The same problem applies to the npm: version, albeit intermittently:

```typescript
#!/usr/bin/env -S deno test -A

import { assertEquals } from 'https://deno.land/std@0.218.0/assert/mod.ts'
import * as esbuild from 'npm:esbuild@0.20.1'

Deno.test('should transform', async () => {
  const result = await esbuild.transform('const test: boolean = true', { loader: 'ts' })
  assertEquals(result.code, 'const test = true;\n')
  await esbuild.stop()
})

Deno.test('should transform again', async () => {
  const result = await esbuild.transform('const test: boolean = true', { loader: 'ts' })
  assertEquals(result.code, 'const test = true;\n')
  await esbuild.stop()
})
```

Will output 90% of the time:

```shell
error: Leaks detected:
  - A child process was started during the test, but not closed during the test. Close the child process by calling `proc.kill()` or `proc.close()`.
  - An async operation to wait for a subprocess to exit was started in this test, but never completed. This is often caused by not awaiting the result of a `Deno.Process#status` call.
To get more details where leaks occurred, run again with the --trace-leaks flag.
```

evanw:
It looks like this isn't caught by esbuild's current tests because esbuild's tests do something like this, which for some reason doesn't cause a problem with Deno:

```js
[...]

PR #3701, opened 2024-03-15
Re-enable `unref` in Deno

lucacasonato:
This re-enables unref-by-default in Deno, and fixes the event loop starvation issue by explicitly re-ref'ing the esbuild sub-process when a user calls `stop()`.

Background: https://github.com/evanw/esbuild/issues/3682#issuecomment-1999391689

### DIFF
--- a/lib/deno/mod.ts
+++ b/lib/deno/mod.ts
@@ -192,6 +192,8 @@ type SpawnFn = (cmd: string, options: {
   read(): Promise<Uint8Array | null>
   close(): Promise<void> | void
   status(): Promise<{ code: number }>
+  unref(): void
+  ref(): void
 }
 
 // Deno ≥1.40
@@ -209,6 +211,11 @@ const spawnNew: SpawnFn = (cmd, { args, stdin, stdout, stderr }) => {
     write: bytes => writer.write(bytes),
     read: () => reader.read().then(x => x.value || null),
     close: async () => {
+      // Ref the child process again, so that a user calling `close()` can await
+      // the returned promise without the event loop starving because there are
+      // no more ref'd async tasks.
+      child.ref()
+
       // We can't call "kill()" because it doesn't seem to work. Tests will
       // still fail with "A child process was opened during the test, but not
       // closed during the test" even though we kill the child process.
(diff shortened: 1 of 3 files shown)


lib/deno/mod.ts
```diff
@@ -209,6 +211,11 @@ const spawnNew: SpawnFn = (cmd, { args, stdin, stdout, stderr }) => {
     write: bytes => writer.write(bytes),
     read: () => reader.read().then(x => x.value || null),
     close: async () => {
+      // Ref the child process again, so that a user calling `close()` can await
+      // the returned promise without the event loop starving because there are
+      // no more ref'd async tasks.
+      child.ref()
```
lucacasonato, line 217: This is the only new code. All other code is just a revert of 116f63ec1d063b5f75c4b3ba959da8c7598c495b.

lib/shared/types.ts
```diff
@@ -664,20 +664,24 @@ export let version: string
...
 // within a Deno test. Deno fails tests that create a child process without
 // killing it before the test ends, so you have to call this function (and
-// await the returned promise) in every Deno test that uses esbuild.
+// await the returned promise) in every Deno test that starts esbuild.
+//
+// You may also start esbuild once at the top level of your test suite (by
+// calling `initialize()`) instead of starting and stopping the esbuild process
+// for every test. This will not interfere with the resource sanitizer, and will
+// improve the efficiency of your tests because the esbuild process can be
+// reused between tests.
```
lucacasonato, line 684: I have added this paragraph to explain that you may also start esbuild outside of a `Deno.test` call to avoid the resource sanitizer flagging it.

lucacasonato:
@evanw Could you take a look at this?

marvinhagemeister (approved):

gmzacharydovel:
Just a heads up, I wouldn't be surprised if error about not being able to initialize esbuild bin folder are related to this because things are running when they shouldn't be running. It looks like some kind of race condition in that case.

https://github.com/denoland/deno/issues/29645

marvinhagemeister:
@gmzacharydovel the issue you liked to is an issue with `deno install`. It's not an issue with esbuild.

gmzacharydovel:
@marvinhagemeister I want to believe that, but it is happening on a already installed project.  Furthermore, the error disappears/appears in ways that make it looks like a race condition.

I appreciate that being raised and I'll try to keep my eye out if Deno is trying to do some kind of hidden install process, but I have my doubts.

Issue #3751, opened 2024-05-06
CVE-2023-45288


eXaminator:
It seems there is a CVE in Go: https://avd.aquasec.com/nvd/cve-2023-45288

Trivvy finds this CVE in our project using `esbuild` through `tsx`:
![image](https://github.com/evanw/esbuild/assets/2758908/d991270d-7816-4cfa-b4b1-b4ba2e1830e3)

evanw:
This vulnerability is irrelevant in a development server that is never exposed to the internet. See esbuild’s [documentation about the development server](https://esbuild.github.io/api/#serve):

> Note that this web server is intended to only be used in development. _Do not use this in production._

and the FAQ entry about [noise from vulnerability scanners](https://esbuild.github.io/faq/#old-go-version):

> These reports are benign and should be ignored. … Later versions of Go have dropped support for certain older platforms that I want esbuild to be able to run on (e.g. older versions of macOS).

evanw:
I'm closing this issue as it's irrelevant, as described above.

coolbeatz71:
I'm facing the same issue. Any fix?

PR #3774, opened 2024-05-21
fix CVE-2023-45288: Upgrade Golang to `v1.22.2`

coolbeatz71:
### What happened？
There are high 1 security vulnerabilities found in Golang `1.20.12`
 
* [CVE-2023-45288](https://github.com/advisories/GHSA-4v7x-pqxf-cx7m)
 
### What did I do？
Upgrade golang to `v1.22.2` on the github workflow. Basically addressing this [issue](https://github.com/evanw/esbuild/issues/3751)
 
### What did you expect to happen？
Ideally, no insecure library should be used.
 
### How can we automate the detection of these types of issues?
By using the [GitHub Actions](https://github.com/murphysecurity/actions) configurations provided by murphysec, we can conduct automatic code security checks in our CI pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,13 +2,13 @@ name: CI
 
 on:
   push:
-    branches: ['*']
+    branches: ["*"]
   pull_request:
-    branches: ['*']
+    branches: ["*"]
   workflow_dispatch:
 
 permissions:
-  contents: read  #  to fetch code (actions/checkout)
+  contents: read #  to fetch code (actions/checkout)
 
 jobs:
   esbuild-platforms:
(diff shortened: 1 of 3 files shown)


evanw:
This vulnerability is irrelevant in a development server that is never exposed to the internet. See esbuild’s [documentation about the development server](https://esbuild.github.io/api/#serve):

> Note that this web server is intended to only be used in development. _Do not use this in production._

and the FAQ entry about [noise from vulnerability scanners](https://esbuild.github.io/faq/#old-go-version):

> These reports are benign and should be ignored. … Later versions of Go have dropped support for certain older platforms that I want esbuild to be able to run on (e.g. older versions of macOS). 

I'm closing this issue as it's irrelevant, as described above.

PR #3784, opened 2024-05-26
fix: use standardized units in text outputs

n-peugnet:
The small `b` was misleading as it usually means "bit", whereas here we have "bytes". Also, the prefixes `k`, `m` and `g` did not match any existing convention, here as we are dividing by 1024, we should use `Ki`, `Mi` and `Gi` to prevent any confusion.

See: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units


n-peugnet:
@evanw what do you think about this?

evanw:
I've never heard someone say "kibibyte" and I would be very confused if someone had. I've only ever heard people use "kilobyte". So for me it would cause more confusion, not less.

n-peugnet:
Of course, the most important part of my proposition was the capital `B` instead of the lowercase one. But this lead to also modifying the prefix.

Here we have 3 possibilities as explained in the Wikipedia article that I linked:

1. The SI prefix (`kB`, `MB`, `GB`, etc.), but we would then have to divide by 1000 instead of 1024 to be conformant with the standard.
2. The JEDEC definition (`KB`, `MB`, `GB`) while keeping the division with 1024, but [as stated in the definition](https://www.jedec.org/standards-documents/dictionary/terms/mega-m-prefix-units-semiconductor-storage-capacity):
   
   > The definitions of kilo, giga, and mega based on powers of two are included only to reflect common usage. IEEE/ASTM SI 10‑1997 states "This practice frequently leads to confusion and is deprecated."

   Indeed, for values other `kB`/`KB`, how can we know if the standard used is the SI or the JEDEC, and thus adivision by 1000 or 1024 ?

3. The IEC standard (`KiB`, `MiB`, `GiB`, etc.) which leaves no ambiguity and matches the current behavior.

Sure you could go for option 2. but it is deprecated and IEC standard is becoming more and more common, for example [Webpack](https://webpack.js.org/guides/getting-started/) (https://github.com/webpack/webpack/commit/9732e371c20a808fce845009dff8c0f946288325) and Git are using it but there are a lot more.

n-peugnet:
Also, as you seem to be using MacOS, it seems the Finder uses neither of these standards/conventions (JEDEC-like prefix with base 10 divisions). [As you can see](https://unix.stackexchange.com/questions/558390/why-does-finder-and-terminal-report-different-file-sizes-for-files-and-folders), the size of a `MB` or `GB` is far from universal. This is why IMO the IEC standard is the least confusing of all.

PR #3897, opened 2024-08-27
Wait until process exit in esbuild.stop()

lucacasonato:
This changes `esbuild.stop()` to only resolve the returned promise when the child process has actually exited.

This is useful for when you have some resource sanitization in place that checks that the child process has actually been closed.


marvinhagemeister (approved):

nicolo-ribaudo:
Ping? :)

I just spent a while trying to figure out what was keeping my process alive longer than expected, and I tracked it down to the promise returned by `esbuild.stop()` resolving before that ESBuild actually stops.

PR #3914, opened 2024-09-16
Provide a flag to auto-prefix local css classes when minified

chowey:


ziemkowski:
@chowey Any reason to limit this to minified classnames? Local classname collission is still an issue without minification if you use common names such as `.btn`, `.row`, `.box`, `.container`, etc. which is easy to run into with CSS Modules.

chowey:
@ziemkowski If I understand you correctly, your requirement should already work.

So long as your CSS files are correctly identified as CSS modules, local classnames (when minification is disabled) should _not_ be an issue. esbuild already renames them.

See [this example](https://esbuild.github.io/try/#YgAwLjI0LjAALS1idW5kbGUgLS1vdXRkaXI9LgBlAGVudHJ5LmpzAGltcG9ydCBhIGZyb20gIi4vYS5tb2R1bGUuY3NzIjsKaW1wb3J0IGIgZnJvbSAiLi9iLm1vZHVsZS5jc3MiOwoKY29uc29sZS5sb2coYS5idG4sIGIuYnRuKTsAAGEubW9kdWxlLmNzcwAuYnRuIHsKICBjb2xvcjogcmVkOwp9AABiLm1vZHVsZS5jc3MALmJ0biB7CiAgY29sb3I6IGJsdWU7Cn0). Taking these two files:

```css
.btn {
  color: red;
}
```

```css
.btn {
  color: blue;
}
```

they successfully bundle into this one file:

```css
/* a.module.css */
.a_btn {
  color: red;
}

/* b.module.css */
.b_btn {
  color: blue;
}
```

chowey:
@evanw is there anything needed to make this acceptable? Such as docs? Or is it a no-go?

mhsdesign:
Thank you for building this. I come around every know and then and really would need that feature as well.

Can anything be done to get a decision here? I assume there are tests missing for a new feature to be merged?

PR #3917, opened 2024-09-19
fix running esbuild cli with deno

Joshix-1:
The problem:
```sh
$ deno --version
deno 1.46.3 (stable, release, x86_64-unknown-linux-gnu)
v8 12.9.202.5-rusty
typescript 5.5.2
$ deno run -A https://deno.land/x/esbuild@v0.23.1/mod.js > /dev/null
error: Uncaught (in promise) TypeError: stdin is not piped
  const writer = child.stdin.getWriter();
                       ^
    at ChildProcess.get stdin (ext:runtime/40_process.js:243:13)
    at spawnNew (https://deno.land/x/esbuild@v0.23.1/mod.js:1712:24)
    at https://deno.land/x/esbuild@v0.23.1/mod.js:1872:3
    at eventLoopTick (ext:core/01_core.js:169:7)
```

Relevant deno docs: https://docs.deno.com/api/deno/~/Deno.Command
This means getWriter/getReader can only be called if stdin/stdout are piped.

### DIFF
--- a/lib/deno/mod.ts
+++ b/lib/deno/mod.ts
@@ -203,11 +203,12 @@ const spawnNew: SpawnFn = (cmd, { args, stdin, stdout, stderr }) => {
     stdout,
     stderr,
   }).spawn()
-  const writer = child.stdin.getWriter()
-  const reader = child.stdout.getReader()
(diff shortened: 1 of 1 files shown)


evanw:
Thanks for the report. It looks like Deno broke this API in Deno version 1.31.0 as far as I can tell. This code was working fine with Deno version 1.30.0.

Issue #3957, opened 2024-10-30
 CLI setting the watch and serve mode did not write to the file system


ghost:
The CLI setting the watch and serve mode did not **write to the file system**
```sh
esbuild test.js --outdir=dist --watch --servedir=dist
```

> By default **the CLI** and JavaScript APIs **write to the file system** and the Go API doesn't. 

https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/pkg/cli/cli_impl.go#L1036-L1041
https://github.com/evanw/esbuild/blob/d34e79e2a998c21bb71d57b92b0017ca11756912/pkg/cli/cli_impl.go#L1435-L1439

PR #3959, opened 2024-10-31
cli: set write option to true by default

laowudui:
- Updated the parseOptionsForRun function to set options.Write to true
- This change ensures that the CLI will write output files by default


laowudui:
#3957

laowudui:
## test
```sh
touch test.js
./esbuild test.js --outdir=dist --watch --servedir=dist
```

PR #4024, opened 2025-01-01
Use the native DecompressionStream API to gunzip the .tgz archive

quentinadam:
Deno natively supports the Web [Compression Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API).

A [`new DecompressionStream("gzip")`](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream) instance can be used to gunzip the `.tgz` archive instead of depending on the `denoflate` package from denoland.

An additional benefit of removing this dependency is that the Deno code then becomes dependency-free which paves the way to support publishing it to [JSR](https://jsr.io/).

### DIFF
--- a/lib/deno/mod.ts
+++ b/lib/deno/mod.ts
@@ -1,7 +1,6 @@
 import type * as types from "../shared/types"
 import * as common from "../shared/common"
 import * as ourselves from "./mod"
-import * as denoflate from "https://deno.land/x/denoflate@1.2.1/mod.ts"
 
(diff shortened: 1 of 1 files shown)


evanw:
I'm closing this PR as esbuild's Deno API has been removed now that https://deno.land/x has been frozen. Deno now has support for installing npm packages, so in theory you can now just install `npm:esbuild` with Deno instead.

Issue #3620, opened 2024-01-31
CSS transforms can break nested selectors


nstepien:
https://esbuild.github.io/try/#YgAwLjIwLjAALS1zdXBwb3J0ZWQ6bmVzdGluZz1mYWxzZQBlAGVudHJ5LmNzcwBkaXYgewogID4gaW5wdXQsCiAgPiBsYWJlbCA+IGlucHV0IHsKICAgIG1hcmdpbjogMWVtOwogIH0KfQ

Esbuild transforms this:
```css
div {
  > input,
  > label > input {
```
into this:
```css
div > :is(input, label > input) {
```
Looks good right? No!
The following
```css
div > :is(label > input)
```
actually means "select an `input` that is an immediate child of `label` AND is also an immediate child of `div`", which is just not possible.

~I use esbuild through Vite, but I couldn't figure out a way to avoid this issue without disabling `build.cssMinify`.
Is there an esbuild option to never transform nested selectors?~

~In Vite I've tried~
```js
  build: {
    target: ['chrome121'],
  },
  esbuild: {
    minifyIdentifiers: false,
    minifySyntax: false,
    minifyWhitespace: false,
    supported: {
      nesting: true
    }
  },
```
~without success, the broken selectors are still using `:is()`.~

nstepien:
I found that `build.cssTarget: []` prevents nested selectors from being transformed!

nstepien:
https://esbuild.github.io/try/#YgAwLjIwLjAALS1zdXBwb3J0ZWQ6bmVzdGluZz1mYWxzZQBlAGVudHJ5LmNzcwA6Oi13ZWJraXQtc2Nyb2xsYmFyLXRodW1iIHsKICAmOndpbmRvdy1pbmFjdGl2ZSB7CiAgICBiYWNrZ3JvdW5kOiByZWQ7CiAgfQogICY6aG92ZXIgewogICAgYmFja2dyb3VuZDogYmx1ZTsKICB9IAp9

This also looks broken.

nstepien:
[...]

PR #4037, opened 2025-01-13
(CSS) Fix nesting lowering with complex sub selectors

tim-we:
This PR fixes the issue described in https://github.com/evanw/esbuild/issues/3620.

When nesting is lowered (for example when the target is Chrome 117) the current version of esbuild will transform this:
```css
.parent {
  > .a,
  > .b1 > .b2 {
    color: red;
  }
}
```
into this ([playground link](https://esbuild.github.io/try/#dAAwLjI0LjIALS1sb2FkZXI9Y3NzIC0tdGFyZ2V0PWNocm9tZTExNwAucGFyZW50IHsKICA+IC5hLAogID4gLmIxID4gLmIyIHsKICAgIGNvbG9yOiByZWQ7CiAgfQp9))
```css
.parent > :is(.a, .b1 > .b2) {
  color: red;
}
```
which is wrong because `.parent > :is(.b1 > .b2)` is semantically different from `.parent > .b1 > .b2`.

With the change in this PR the output will instead be this:
```css
.parent > .a,
.parent > .b1 > .b2 {
  color: red;
}
```

I have added a test for this as well. The fix is in the first pass of the `lowerNestingInRuleWithContext` method. When a nested complex selector is sufficiently complex (has another combinator) the transformation will be disabled.

### DIFF
--- a/internal/css_parser/css_nesting.go
+++ b/internal/css_parser/css_nesting.go
@@ -151,6 +151,15 @@ func (p *parser) lowerNestingInRuleWithContext(rule css_ast.Rule, context *lower
 				} else if commonLeadingCombinator.Byte != combinator.Byte {
 					canUseGroupDescendantCombinator = false
 				}
+
+				// If the complex selector contains multiple combinators (e.g., "& «COMBINATOR» «something» «COMBINATOR» «something»"),
+				// we cannot safely transform it into "parent :is(...selectors)". See https://github.com/evanw/esbuild/issues/3620.
+				for _, subSelector := range sel.Selectors[2:] {
+					if subSelector.Combinator.Byte != 0 {
+						canUseGroupDescendantCombinator = false
+						break
+					}
+				}
 			}
 
 			// Are all children of the form "&«something»"?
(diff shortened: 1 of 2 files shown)


sanoojes:
I think its not fixed 

config
```
{
  bundle: true,
  platform: "browser",
  globalName: "Lucid",
  sourcemap: false,
  minify: true,
  entryPoints: [ "styles/app.css" ],
  outfile: "src/user.css",
  supported: { nesting: false },
  plugins: []
}
```

After build
image

Source
image

tim-we:
Hello @sanoojes, this is PR that did not get accepted, the issue was solved in a different way. If you think this is not fixed either comment on #3620 or create a new issue.
However looking at the code you posted I don't understand the issue, the transformation looks ok? Please state clearly what you think is wrong, it is very much not obvious from those screenshots.

sanoojes:
image

ohh sorry to comment here 
also the transformation i want is something like in the screenshot above as the sass complies
i do not want that :is() selectors any way to fix it ?
i comment on issue #3620 when am free 
i will provide details and issues soon as i get free

also apologies if i made mistakes 
English its not my first language

tim-we:
This PR and the issue #3620 is about the correctness of the transformation. If I understand you correctly you just don't want any [`:is()` pseudo-class function](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) in your output. This is unrelated to this PR and the issue.

I wonder why you want that as `:is()` seems to be supported by all major browser for more than 10 years. To disable that transformation you just have to pick a very old browser as a target, like Firefox 77 or Chrome 87 (those are from 2011).

If you change the browser version you can see the effect:
https://esbuild.github.io/try/#dAAwLjI1LjEALS10YXJnZXQ9ZmlyZWZveDc3IC0tbG9hZGVyPWNzcwAuYSwgLmIgewogIC5jIHsKICAgIGNvbG9yOiByZWQ7CiAgfQp9

PR #4082, opened 2025-02-16
Fix emitting real mapping sources when null mapping comes first

### LINKED ISSUE #4080: Source mappings may now contain `null` entries in 0.24.1 causes errors generate non spec compliant map?
The mentioned change causes errors when using `@ampproject/remapping`, likely because it does not comply with the specification. The original issue was reported https://github.com/ampproject/remapping/issues/200), but it was later clarified that the behavior aligns with the spec.  

> `rootSources`, derived from the map's `sources` field, contains 16 items. However, `segment[1]` is `16`, meaning `rootSources[16]` results in `undefined`. Passing `undefined` to `originalPositionFor` triggers an error.  
>  
> This is not an issue with remapping itself; rather, it's invalid for a source map to reference a 17th source (index 16) when only 16 sources exist. According to the [[source map specification](https://tc39.es/ecma426/#decode-source-map-mappings)](https://tc39.es/ecma426/#decode-source-map-mappings) (9.a.iii.14):  
>  
> > If any of `sourceIndex`, `originalLine`, or `originalColumn` are less than 0, or if `sourceIndex` is greater than or equal to `sources`’s [[size](https://infra.spec.whatwg.org/#list-size)](https://infra.spec.whatwg.org/#list-size), [[optionally report an error](https://tc39.es/ecma426/#optionally-report-an-error)](https://tc39.es/ecma426/#optionally-report-an-error).

### LINKED ISSUE #4107: Sourcemap of script containing the `ai` library produces invalid sourcemap
Repro
```js
import { streamText } from "ai";
console.log(streamText);
```

ai at version `4.1.50`

```
npx esbuild repro.js --bundle --sourcemap --outfile=dist/repro.js
```

The generated dist/repro.js.map fails to parse on a variety of tools. Eg.
https://evanw.github.io/source-map-visualization/
https://www.npmjs.com/package/source-map-explorer
https://crates.io/crates/sourcemap

Errors are usually of this variety
```
The "mappings" field of the imported source map contains invalid data. Invalid VLQ data at index 147599: Original source index 124 is invalid (there are 123 sources).
```

[repro.js.map.txt](https://github.com/user-attachments/files/19115585/repro.js.map.txt)

jridgewell:
This fixes an ordering bug when a null mapping and a real mapping share the same `sourceIndex`. If a null mapping comes first, then we'll assign it a `sourceIndexToSourcesIndex` but skip recording its source files in the `items` array because of the `isNullEntry` check. When we find the real mapping later in `results`, we'll skip recording its source files because of the map-`ok` check.

The simplest fix I can think of is flipping the order just a bit. We'll skip assigning `sourceIndexToSourcesIndex` for null mappings, so that when a real mapping comes up we'll record its sources.

I'm not sure how to generate a unit test case for this, but it resolves the issue found in https://github.com/angular/angular-cli/issues/29465#issue-2807308693.

Fixes #4080
Fixes #4107

### DIFF
--- a/internal/linker/linker.go
+++ b/internal/linker/linker.go
@@ -6963,13 +6963,13 @@ func (c *linkerContext) generateSourceMapForChunk(
 	items := make([]item, 0, len(results))
 	nextSourcesIndex := 0
 	for _, result := range results {
+		if result.isNullEntry {
+			continue
+		}
 		if _, ok := sourceIndexToSourcesIndex[result.sourceIndex]; ok {
 			continue
 		}
 		sourceIndexToSourcesIndex[result.sourceIndex] = nextSourcesIndex
-		if result.isNullEntry {
-			continue
-		}
 		file := &c.graph.Files[result.sourceIndex]
 
 		// Simple case: no nested source map
@@ -7057,7 +7057,10 @@ func (c *linkerContext) generateSourceMapForChunk(
 	for _, result := range results {
 		chunk := result.sourceMapChunk
 		offset := result.generatedOffset
-		sourcesIndex := sourceIndexToSourcesIndex[result.sourceIndex]
+		sourcesIndex, ok := sourceIndexToSourcesIndex[result.sourceIndex]
+		if !ok {
+			continue
+		}
 
 		// This should have already been checked earlier
 		if chunk.ShouldIgnore {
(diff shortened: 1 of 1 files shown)


meriturva:
No chance that this PR will be merged?
Thank you!

evanw:
Thanks for the PR! I'll take care of reducing the repro and adding a test case. The linked Angular issue is very helpful as I'm able to reproduce the problem myself. A minimal repro is pretty complicated and may involve at least seven files.

It looks like the issue happens when the same source file emits multiple discontiguous parts in the output file and source maps are skipped for the first part. This can happen when esbuild automatically generates a separate `__export(exports, { ... })` part for the file to handle ESM namespace exports, for example.

Looks like this also fixes #4107.

PR #4092, opened 2025-02-21
Fix lowering standard decorators on a class with inheritance causing field initializers before super call with useDefineForClassFields: false

### LINKED ISSUE #4087: Private field incorrectly initialized before the super call
> Sorry that I had to close the prior issue which had no responses (https://github.com/evanw/esbuild/issues/3939) and open this as this bug is blocking us and our framework users (Serenity) from updating esbuild to 0.24 or later:

This is an issue introduced in 0.24 as 0.23.1 does not have this problem. Probably introduced with the fix to https://github.com/evanw/esbuild/issues/3913

- [ESBuild Playground Link](https://esbuild.github.io/try/#dAAwLjI0LjAAewogIGxvYWRlcjogJ3RzJywKICBmb3JtYXQ6ICdpaWZlJywKICB0YXJnZXQ6ICdlczIwMTUnLAogIHRzY29uZmlnUmF3OiB7CiAgICBjb21waWxlck9wdGlvbnM6IHsKICAgICAgdXNlRGVmaW5lRm9yQ2xhc3NGaWVsZHM6IGZhbHNlCiAgICB9CiAgfSwKfQBmdW5jdGlvbiB0ZXN0RGVjb3JhdG9yKGE6IHN0cmluZykgewogIHJldHVybiBmdW5jdGlvbiAodGFyZ2V0OiBGdW5jdGlvbiwgX2NvbnRleHQ/OiBhbnkpIHsKIH0KfQoKY2xhc3MgQSB7CiAgIGNvbnN0cnVjdG9yKCkgewogICB9Cn0KCkB0ZXN0RGVjb3JhdG9yKCJhIikKY2xhc3MgQiBleHRlbmRzIEEgewogIHByaXZhdGUgbTE6IGFueTsKCiAgY29uc3RydWN0b3IoKSB7CiAgICBzdXBlcigpOwogIH0KCn0KCm5ldyBCKCk7)

```js
{
  loader: 'ts',
  format: 'iife',
  target: 'es2015',
  tsconfigRaw: {
    compilerOptions: {
      useDefineForClassFields: false
    }
  }
}
```

```typescript
function testDecorator(a: string) {
  return function (target: Function, _context?: any) { }
}

class A {
   constructor() {
   }
}

@testDecorator("a")
class B extends A {
  private m1: any;

  constructor() {
    super();
  }

}

new B();
```

The generated constructor for the derived B class with a private field includes a field initializer before the super call which results in the  following error in browser console if you copy paste the generated code:

```
VM17:64 Uncaught ReferenceError: Must call super constructor in derived class 
before accessing 'this' or returning from derived constructor
    at new B (<anonymous>:64:7)
    at <anonymous>:71:3
    at <anonymous>:72:3
```

```js
class B extends (_a = A) {
    constructor() {
      this.m1 = void 0;
      super();
    }
  }
```

In addition to producing invalid code, I think this behavior should only be applied to properties with decorators themselves (not when only the class  itself has a decorator). Otherwise this will be a breaking change for those using useDefineForClassFields: false. 

TypeScript does not do this for classes with only class decorators:

https://www.typescriptlang.org/play/?target=2#code/GYVwdgxgLglg9mABFApgZygERROAnAQynwAooC8BzFKALkQLAE8AaRXMVADzoeYEpEAbwBQAXxEiIAGwJo0iAILCRiNRwx4Q0UoNFrEEiSIACqDNlyFieKbPmIAQohQ8UYACYLl+xAFsARnpGJgBuSXUETW0bEj1VAzQQAAcUPDjwtWNjMBQAdycMoA

But only for individual properties with decorators:

https://www.typescriptlang.org/play/?target=2#code/GYVwdgxgLglg9mABFApgZygERROAnAQynwAooC8BzFKALkQLAE8AaRXMVADzoeYEpEAbwBQAXxEiIAGwJo0iAILCRiNRwx4Q0UoNFrEEiVNnzEAIUQoeKMABMFy-YgACqDNlyFieVYgAOeDAAbkQoiAC2AIz0jEwA3JLqCJraPiR6fmpoIP4oeBmJasbGYCgA7haFQA

furkanevran:
Allow omitting the field initializers if the class have a decorator and the field doesn't since esbuild doesn't initialize it with extra initializers and only set it to it's default value when useDefineForClassFields is false

Fixes #4087

### DIFF
--- a/internal/bundler_tests/bundler_tsconfig_test.go
+++ b/internal/bundler_tests/bundler_tsconfig_test.go
@@ -2875,6 +2875,10 @@ func TestTsconfigDecoratorsUseDefineForClassFieldsFalse(t *testing.T) {
 		files: map[string]string{
 			"/Users/user/project/src/entry.ts": `
 				class Class {
+					foo
+				}
+				class ClassExtends extends Class {
+					bar
 				}
 				class ClassMethod {
 					foo() {}
@@ -2882,18 +2886,24 @@ func TestTsconfigDecoratorsUseDefineForClassFieldsFalse(t *testing.T) {
 				class ClassField {
 					foo = 123
 					bar
+					baz
 				}
 				class ClassAccessor {
 					accessor foo = 123
 					accessor bar
 				}
 				new Class
+				new ClassExtends
 				new ClassMethod
 				new ClassField
 				new ClassAccessor
 			`,
 			"/Users/user/project/src/entrywithdec.ts": `
 				@dec class Class {
+					foo
+		        }
+				@dec class ClassExtends extends Class {
+					bar
 				}
 				class ClassMethod {
 					@dec foo() {}
@@ -2901,14 +2911,21 @@ func TestTsconfigDecoratorsUseDefineForClassFieldsFalse(t *testing.T) {
 				class ClassField {
 					@dec foo = 123
 					@dec bar
+					baz
+				}
+				class ClassFieldExtends extends Class {
+					@dec bar
+					baz
 				}
 				class ClassAccessor {
 					@dec accessor foo = 123
 					@dec accessor bar
 				}
 				new Class
+				new ClassExtends
 				new ClassMethod
 				new ClassField
+				new ClassFieldExtends
 				new ClassAccessor
 			`,
 			"/Users/user/project/src/tsconfig.json": `{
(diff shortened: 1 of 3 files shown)


furkanevran:
I could also alternatively replace the condition on:

https://github.com/evanw/esbuild/blob/v0.25.0/internal/js_parser/js_parser_lower_class.go#L608-L610

to 
```go
class.ShouldLowerStandardDecorators || (result.lowerAllInstanceFields && class.ExtendsOrNil.Data != nil)
```

PR #4094, opened 2025-02-23
fix CORS with live reloading

hamdouni:
Just add the permissive header to allow cross origin request, so live reloading works when serving files with another http server.

Use case : a Go application with an http server, serving static files (including JS files) and esbuild server for watching files modification. If the JS file is not served by the esbuild server, requests to /esbuild are blocked with "CORS Missing Allow Origin".

I presume allowing this in development mode may be acceptable.


ZuBB:
@hamdouni it used to be that way, but I feel it's not coming back

see this commit for more details https://github.com/evanw/esbuild/commit/de85afd65edec9ebc44a11e245fd9e9a2e99760d

hamdouni:
Got it. 
I'm not found of having a proxy just for dev.
I will try to find another way to have 2 servers on localhost with different ports.
Thanks 👍🏾

Issue #3439, opened 2023-10-10
Request: support "indexed" source maps


seanstrom:
Hello 👋 

I have a request for functionality related to how esbuild supports source maps at the moment. And in this case I would to request some additional support for source maps with the `sections` field, a.k.a ["indexed" source maps](https://github.com/tc39/source-map-spec/blob/main/source-map-rev3.md#index-map-supporting-post-processing)

The reason for this request is because I'm using some tooling that compiles to JavaScript and provides a source map with this indexed source map format. And at the moment, these source maps are not supported by esbuild: 
https://github.com/evanw/esbuild/blob/02dae18a70b3d26dcb5b9599f02a9a36d88f809d/internal/js_parser/sourcemap_parser.go#L38-L41

---

But I'm hoping that it would be possible to "flatten" these indexed source maps in a way that is similar to [Webpack's approach seen here](https://github.com/webpack-contrib/source-map-loader/blob/5e265904edf5c04d8dc209755926441a2732fc1e/src/utils.js#L38-L75). What are your thoughts? Let me know what you think 🙏

evanw:
I have done a lot of work with source map infrastructure over the years and I've only ever encountered one indexed source map (by chance, on a random website). So I don't actually really have any test cases to go by. It'd be helpful for you to provide a way to reproduce your issue here (i.e. providing specific code with specific build instructions).

seanstrom:
Yup yup sure thing 👍 

Here's a minimal example of my specific use case of compiling from ClojureScript to JavaScript: https://github.com/seanstrom/cljs-dev-template/tree/minimal-sourcemaps-debug-esbuild

I've included some instructions inside the README of that branch, should be as simple as setting the local environment to have some dependencies like Java, Clojure, and Node. And then running some NPM scripts for compiling the source code and bundling the outputs with esbuild. If you need specific version numbers for the environment dependencies, you can reference the `.tool-versions` file.

If you need any help with setting up the project, please let me know 🙏

jridgewell:
> I have done a lot of work with source map infrastructure over the years and I've only ever encountered one indexed source map

They're definitely rare. Turbopack is the only build tool I know of that generates them purposefully (it interacts with our caching infrastructure gracefully), but we're not currently used in any npm libraries that esbuild would then have to consume. Maybe in the future they'll be more popular.

seanstrom:
[...]

PR #4109, opened 2025-03-07
feat: support index source map

clyfish:
fix https://github.com/evanw/esbuild/issues/3439


evanw:
Looks good to me, and works for the test case in #3439 (although that test case only uses a single section). I'll move forward with this then. Thanks for the implementation!

Issue #4131, opened 2025-04-08
Bug: Memory Leak with watch mode `--watch=true`


mxschmitt:
Reproduction:

1. `git clone https://github.com/microsoft/playwright && cd playwright`
2. `npm ci`
3. `npx esbuild --log-level=verbose "packages/playwright-core/src/**/*.ts" --outdir=packages/playwright-core/lib --sourcemap=linked --watch=true --platform=node --format=cjs`
4. Go to `packages/playwright-core/src/client/page.ts` and modify the file header 20 times, between each time save the file and look at the memory consumption.

This can be used as well:

```ts
import fs from "fs"
for (let i = 0; i < 100; i++) {
  fs.appendFileSync("packages/playwright-core/src/client/page.ts", "\n")
  await new Promise((resolve) => setTimeout(resolve, 50))
}
```

and aborted after a few seconds.

Expected: It stays around 50-200 MB
Actual: It grows up to 10-40 GB

Version: 0.25.0 - also tested with 0.25.2

I was able to reproduce on macOS and Linux (Ubuntu amd64).

mxschmitt:
Investigation:

- We are doing `--watch=true`
- Internally if `--watch` is not passed, GC is disabled as per here:
  - https://github.com/evanw/esbuild/blob/44e746965d783646f97daf3d0617ff816727e7fb/cmd/esbuild/main.go#L316
- Changing it from `--watch=true` to `--watch` fixes it

evanw:
Thanks for figuring this out, and for the PR. Quite a dumb bug. I'll include the fix for this in the next release.

PR #4132, opened 2025-04-08
fix(cli): --watch normalisation with=true

mxschmitt:
This aligns it with how the internal CLI parser accepts `--watch=true` which was implemented in https://github.com/evanw/esbuild/commit/ce04765e63fda0ec8d39dc0265970a8396abecdd.

Before `--watch=true` was not correctly identified as "Watch Mode" and disabled the GC as a drive by. This caused memory leaks.

Resolves https://github.com/evanw/esbuild/issues/4131

### DIFF
--- a/cmd/esbuild/main.go
+++ b/cmd/esbuild/main.go
@@ -222,11 +222,11 @@ func main() {
 			// This may be a reasonable thing to do in a short-lived VM where all
 			// processes in the VM are only started once and then the VM is killed
 			// when the processes are no longer needed.
-			if arg == "--watch" {
-				isWatch = true
-			} else if arg == "--watch=forever" {
+			if arg == "--watch=forever" {
 				arg = "--watch"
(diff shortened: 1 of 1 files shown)


cmd/esbuild/main.go
```diff
@@ -300,7 +300,7 @@ func main() {
 			for _, arg := range osArgs {
 				if !strings.HasPrefix(arg, "-") {
 					nonFlagCount++
-				} else if arg == "--serve" || arg == "--watch" || strings.HasPrefix(arg, "--serve=") {
+				} else if arg == "--watch" || strings.HasPrefix(arg, "--watch=") || arg == "--serve" || strings.HasPrefix(arg, "--serve=") {
```
JoostK, line 303: This seems inaccurate for `=false` options (it already was for `serve`, now also for `watch`)

Issue #3980, opened 2024-11-18
ESM dynamic import to glob imports, throw synchronous error


suguanYang:
### Expected behavior
the transformed Import Calls should follow the ecma specification, which should be a Rejected Promise.

### current behavior:
```
const main = async () => {
   const num = 1;
   await import(`./file${num }`).catch(err => {
         // can not capture module not fould error
   });

   try {
      await import(`./file${num }`)
   } catch (err) {
     // this could
   }

}
```
the transformed code:
```
  var __glob = (map) => (path) => {
    var fn = map[path];
    if (fn) return fn();
    throw new Error("Module not found in bundle: " + path); // synchronously throws an error.
  };
  // import("./file*") in entry.js
  var globImport_file = __glob({
    "./file.js": () => Promise.resolve().then(() => (init_file(), file_exports))
  });
  var main = async () => {
    const num = 3;
    await globImport_file(`./file${num}`).catch((err) => {
      if (err.code === "ERR_MODULE_NOT_FOUND") {
        console.log(err);
      }
    });
    try {
      await globImport_file(`./file${num}`);
    } catch (err) {
    }
  };
```

### Possible Workaround:
add a new glob for dynamic import:
```
var __globImport = map => path => {
	var fn = map[path]
	if (fn) return fn()
	return Promise.reject('Module not found')
}
```

PR #4153, opened 2025-04-22
fix: use __glob_dynamic for dynamic import

suguanYang:
This fix https://github.com/evanw/esbuild/issues/3980 by provides a different glob function: `__glob_dynamic` which return a rejected promise.


suguanYang:
Actually the current synchronous  behavior is similar to nodejs, should we respect the specification or the host platform?

PR #4163, opened 2025-04-30
fix: handle `__proto__` as a computed property in exports and add tests for special names

### LINKED ISSUE #4162: Special export name `__proto__` is not exported
See playground:
https://esbuild.github.io/try/#dAAwLjI1LjMALS1mb3JtYXQ9Y2pzAGV4cG9ydCBjb25zdCBfX3Byb3RvX18gPSAxMjM7

magic-akari:
- Closes: #4162


magic-akari:
We do not need to modify the code related to `cjs-module-lexer`, as `cjs-module-lexer` can parse `{ __proto__ }` but fails to parse `{ ['__proto__']: null }`.

PR #4171, opened 2025-05-05
dev server: simple support for CORS requests

### LINKED ISSUE #4125: Make CORS configurable
https://github.com/evanw/esbuild/commit/de85afd65edec9ebc44a11e245fd9e9a2e99760d seems to have overcorrected when it came to fixing a cors vulnerability. Some level of configurability would be greatly appreciated.

evanw:
Starting with version 0.25.0, esbuild's development server is no longer configured to serve cross-origin requests. This was a deliberate change to prevent any website you visit from accessing your running esbuild development server. However, this change prevented (by design) certain use cases such as "debugging in production" by having your production website load code from `localhost` where the esbuild development server is running.

To enable this use case, esbuild is adding a feature to allow [Cross-Origin Resource Sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) (a.k.a. CORS) for [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests). Specifically, passing your origin to the new `cors` option will now set the `Access-Control-Allow-Origin` response header when the request has a matching `Origin` header. Note that this currently only works for requests that don't send a preflight `OPTIONS` request, as esbuild's development server doesn't currently support `OPTIONS` requests.

Some examples:

* **CLI:**

    ```
    esbuild --servedir=. --cors-origin=https://example.com
    ```

* **JS:**

    ```js
    const ctx = await esbuild.context({})
    await ctx.serve({
      servedir: '.',
      cors: {
        origin: 'https://example.com',
      },
    })
    ```

* **Go:**

    ```go
    ctx, _ := api.Context(api.BuildOptions{})
    ctx.Serve(api.ServeOptions{
      Servedir: ".",
      CORS: api.CORSOptions{
        Origin: []string{"https://example.com"},
      },
    })
    ```

The special origin `*` can be used to allow any origin to access esbuild's development server. Note that this means any website you visit will be able to read everything served by esbuild.

Fixes #4125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,46 @@
 
 ## Unreleased
 
+* Add simple support for CORS to esbuild's development server ([#4125](https://github.com/evanw/esbuild/issues/4125))
+
+    Starting with version 0.25.0, esbuild's development server is no longer configured to serve cross-origin requests. This was a deliberate change to prevent any website you visit from accessing your running esbuild development server. However, this change prevented (by design) certain use cases such as "debugging in production" by having your production website load code from `localhost` where the esbuild development server is running.
+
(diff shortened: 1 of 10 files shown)


YangYongAn:
good job! there is no longer to need express to proxy !
image

mbrevda:
Thanks!

Issue #4164, opened 2025-04-30
Nested @media transformed incorrectly for pseudo elements


EECOLOR:
I have the following `css`:

```
.this-works1.test {
  background-color: hotpink;

  @media screen and (min-width: 768px) {
    background-color: yellow;
  }
}

.this-works2 {
  &.test {
    background-color: hotpink;

    @media screen and (min-width: 768px) {
      background-color: yellow;
    }
  }
}

.this-does-not-work1::before {
  background-color: hotpink;

  @media screen and (min-width: 768px) {
    background-color: yellow;
  }
}

.this-does-not-work2 {

  &::before {
    background-color: hotpink;

    @media screen and (min-width: 768px) {
      background-color: yellow;
    }
  }
}
```

I am using the following code to transform it:

```
const { code, map, warnings } = await esbuild.transform(cssSource, {
        sourcefile: `${prefix}.css`,
        sourcemap: 'external',
        loader: 'local-css',
        target: ['chrome94', 'opera79', 'edge94', 'firefox92', 'safari15']
      })
```

The result is the following:

```
.prefix_this-works1.prefix_test {
  background-color: hotpink;
}
@media screen and (min-width: 768px) {
  .prefix_this-works1.prefix_test {
    background-color: yellow;
  }
}
.prefix_this-works2.prefix_test {
  background-color: hotpink;
}
@media screen and (min-width: 768px) {
  .prefix_this-works2.prefix_test {
    background-color: yellow;
  }
}
.prefix_this-does-not-work1::before {
  background-color: hotpink;
}
@media screen and (min-width: 768px) {
   {
    background-color: yellow;
  }
}
.prefix_this-does-not-work2::before {
  background-color: hotpink;
}
@media screen and (min-width: 768px) {
   {
    background-color: yellow;
  }
}
```

As you can see it created these rules:

```
@media screen and (min-width: 768px) {
   {
    background-color: yellow;
  }
}
```

Here the selector inside of `@media` is missing.

I was expecting something more like this:

```
@media screen and (min-width: 768px) {
  .prefix_this-does-not-work2::before {
    background-color: yellow;
  }
}
```

---

I am not sure if this is an easy fix. I did not dig into the source code too much (and have no experience in Go). On top of that I am not sure if we can simply switch the target to browsers that support nesting.

In any case, if it is an easy fix all good. If it is not easy, please give me a few pointers that would allow me to do the work and submit a pull request (if we need to actually support older browsers).

[...]

PR #4177, opened 2025-05-08
Allow media queries nested in pseudo element selectors

EECOLOR:
Fixes: https://github.com/evanw/esbuild/issues/4164

Please note that I used AI to create the solution. From what I can tell from my limited understanding of Go this solution makes sense. What I can not determine:

- Is the function I added the best way of achieving this result?
- Is this in line with your style of writing Go?
- Should the function be in that specific file?
- Did I miss any test cases for subtle interaction with other features?

This change fixes the following use case:

`input`
```
a {
  &::before {
    @media screen {
      ...
    }
  }
}
```

`before`
```
@media screen {
  {
    ...
  }
}
```

`after`
```
@media screen {
  &::before {
    ...
  }
}
```

### DIFF
--- a/internal/css_parser/css_nesting.go
+++ b/internal/css_parser/css_nesting.go
@@ -47,7 +47,7 @@ func (p *parser) lowerNestingInRule(rule css_ast.Rule, results []css_ast.Rule) [
 			// Note: This is only for the parent selector list that is used to
 			// substitute "&" within child rules. Do not filter out the pseudo
 			// element from the top-level selector list.
-			if !sel.UsesPseudoElement() {
+			if !sel.UsesPseudoElement() || !containsNestingSelectors(r.Rules) {
 				parentSelectors = append(parentSelectors, css_ast.ComplexSelector{Selectors: substituted})
 			}
 		}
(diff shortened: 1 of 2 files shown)


bronisMateusz:
Annoying bug, I also have the same problem in my project, so it would be great to see a fix finally implemented. Thanks @EECOLOR for explaining the source of the problem.

bronisMateusz:
I started to think about it, and I think EsBuild is working properly here though, looking at how nesting should be used: [https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting#nested_declarations_rule](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting#nested_declarations_rule)

The errors lie in our notation, not in the way EsBuild works. LightningCSS behaves similarly when parsing files.

I change code from this:
```css
.navbar__container {
    position: absolute;
    display: grid;

    &::before {
        position: absolute;
        content: "";

        @media (min-width: 64rem) {
            inset-block-start: 2rem;
            inset-inline-end: 3.625rem;
        }
    }
}
```

to this:
```css
.navbar__container {
    position: absolute;
    display: grid;

    @media (min-width: 64rem) {
        &::before {
            position: absolute;
            content: "";
        }
    }

    &::before {
        position: absolute;
        content: "";
    }
}    
```

And build looks fine.

evanw:
I'm closing this because I believe it was fixed by the fix for #4265, which was released in [version 0.25.10](https://github.com/evanw/esbuild/releases/tag/v0.25.10).

EECOLOR:
@evanw Thanks! As soon as I have time I will verify it. https://github.com/evanw/esbuild/issues/4164 has a more complete test case, so if it indeed fixes that problem I will close that ticket as well.

I will let you know if it does not solve the problem.

PR #4200, opened 2025-06-08
io/fs.FS for input files.

### LINKED ISSUE #3954: Support using a `fs.FS` for input and output files
I am building a deep integration with ESBuild to my Go application. Due to how it works, some of the files are embedded, and some are on the filesystem, and all access to files happen through the [`fs.FS`](https://pkg.go.dev/io/fs#FS) interface.

In essence, I would like to be able to pass an `fs.FS` interface to all of the different methods and functions and for ESBuild to use it for all file accesses. Another separate `fs.FS` should be available to be passed for the output (as the [`embed.FS`](https://pkg.go.dev/embed#FS) is read-only).

It _should_ be a relatively straightforward change, but as I don't know the codebase, I cannot be sure.

firelizzard18:
Closes #3954. Adds `InputFS io/fs.FS` which allows the caller to provide a filesystem that input files will be read from.

Though it actually uses `internal/io_fs` because importing `io/fs` would break compatibility with versions earlier than Go 1.16. The shim for earlier versions is not intended to be implementable; if a caller wishes to pass in an `io/fs.FS`, that caller should use Go 1.16 or newer.

### DIFF
--- a/internal/fs/io_fs.go
+++ b/internal/fs/io_fs.go
@@ -1,0 +1,342 @@
+package fs
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/evanw/esbuild/internal/io_fs"
+	"golang.org/x/sys/unix"
+)
+
+func FromFS(fs io_fs.FS) FS {
+	ret := &ioFS{FS: fs}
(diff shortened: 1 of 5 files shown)


hulyaafatih-rgb (approved):

firelizzard18:
I resolved the conflict

Issue #3310, opened 2023-08-11
Better support for disabling sourcemaps of `node_modules` as plugins can't cover all cases.


AshHeskes:
Referencing [this issue](https://github.com/evanw/esbuild/issues/1685) which discusses disabling sourcemaps for `node_modules`. @evanw quite kindly provided a code snippet as a plugin that would filter out sourcemaps for node modules by basically appending an inline sourcemap containing nothing to each file matching the filter.

While this snippet does work in the majority of cases where the imported file is a flavour of JS that supports such comments. It doesn't for example work for `require('something.json')` which is valid in node > 16 (perhaps earlier). If you append the empty inline sourcemap you get an error during bundling that comments are not supported in JSON. Which is also correct. 

As a quick example project:

```md
+ app
  - index.js (`require('some-lib')`)
+ node_modules
  + some-lib
    - index.js (`require('./some-json.json')` - valid)
    - some-json.json (`{ my-prop: 'foo' }` - valid )
```

With an esbuild config of (using the snippet provided by @evanw):

```js
const esbuild = require('esbuild');
const { readFile } =  require('node:fs/promises');

esbuild.build( {
  entryPoints : [ 'src/index.js' ],
  bundle : true,
  outdir : './dist',
  format : 'cjs',
  platform: 'node'
  plugins: [
    {
      name: 'excludeNodeModulesFromSourceMaps',
      setup(build) {
        build.onLoad({ filter: /node_modules/ }, async (args) => {
          return {
            contents: `${await readFile(
              args.path,
              'utf8'
            )}\n//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIiJdLCJtYXBwaW5ncyI6IkEifQ==`,
            loader: 'default'
          };
        });
      }
    }
  ]
} );
```

Will error due to the fact you can't have a comment in `json`. Which means you have to filter out `json` imports from `excludeNodeModulesFromSourceMaps`. Which means `node_modules` importing `json` will still be included in sourcemaps. 

That might not sound like a big deal and normally I would agree. However this method of excluding sourcing maps via a comment doesn't only apply to `json` it can apply to anything that doesn't use `//` as a comment. Like a text file, or a binary but it seems insane to handle on a case by case basis. So I think there needs to be a better mechanism to say don't sourcemap dependencies  Than just appending an inline empty sourcemap comment. 

[...]

PR #4217, opened 2025-06-24
feat: add exclude-sourcemap option to 

krypciak:
Implements #3310  

Let me know if something is wrong or if I should add more documentation :+1:


mackuba:
Oh yeah that would be very nice to have!

PR #4226, opened 2025-07-02
add more CSS `KnownDeclarations`

OverflowCat:


evanw:
What is the intended effect of this PR?

Edit: I'm asking because you have no tests and there's nothing in the description that motivates this change.

PR #4262, opened 2025-08-10
fix: enable constant folding in case of pure expressions

### LINKED ISSUE #4259: Constant folding failure
```js
const a = 256 * 2;
```
minify to
```js
const a=512;
```
(https://esbuild.github.io/try/#dAAwLjI1LjgALS1taW5pZnkAY29uc3QgYSA9IDI1NiAqIDI7)

But
```js
0;
const a = 256 * 2;
```
minify too
```js
const a=256*2;
```
(https://esbuild.github.io/try/#dAAwLjI1LjgALS1taW5pZnkAMDsKY29uc3QgYSA9IDI1NiAqIDI7), where constant folding fails.

viveknathani:
fixes #4259

Incase of pure expression statements, we can let constant folding happen.

All tests are passing. No impact on expressions with side effects or other statement types.

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -10277,6 +10277,24 @@ func (p *parser) keepClassOrFnSymbolName(loc logger.Loc, expr js_ast.Expr, name 
 	}}
 }
 
+// isSimplePureExpr returns true for expressions that are side-effect free
+func isSimplePureExpr(expr js_ast.Expr) bool {
+	switch e := expr.Data.(type) {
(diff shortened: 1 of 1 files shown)


evanw:
First, these statements are not side-effect free. For example `1+2n` throws an exception which is a side effect.

Second, I'm confused about why this is relevant to handle. Such statements are inherently dead code and useless. What's a real-world example where this code change of yours is relevant?

viveknathani:
Aah, my bad!

Oversight from my end on the `1+2n` scenario.

Thanks for pointing it out! Closing the PR.

PR #4267, opened 2025-08-21
Keep the global names unchanged

GG323:


GG323:
Keep the global names unchanged

PR #4295, opened 2025-10-05
fix #4286: use `Uint8Array.fromBase64` if present

### LINKED ISSUE #4286: Update esbuild's `__toBinary` helper to use `Uint8Array.fromBase64` if available
I am implementing something where it would be useful to have a base64-to-uint8array decoder at runtime. Since I'm already loading a few assets with the 'binary' loader, esbuild includes its own fast base64 decoder. However when I try to access it using its name `__toBinary` esbuild renames it to `__toBinary2` because it thinks my `__toBinary` is something else in the global scope and so it should not clobber that. I could just paste esbuild's implementation into my own files, but esbuild doesn't (can't) remove the second copy which just seems a little dumb. Is there any way to allow module code to reference the same functions that esbuild has when the same thing needs to be done? If not, can one be added?

__toBinary getting renamed:
https://esbuild.github.io/try/#YgAwLjI1LjkALS1sb2FkZXI6LmJpbj1iaW5hcnkgLS1idW5kbGUgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IHggZnJvbSAiLi9mb28uYmluIjsKY29uc29sZS5sb2coX190b0JpbmFyeSgiYWFhYSIpKTsKY29uc29sZS5sb2coeCk7AABmb28uYmluAGFhYWE

__toBinary included twice:
https://esbuild.github.io/try/#YgAwLjI1LjkALS1sb2FkZXI6LmJpbj1iaW5hcnkgLS1idW5kbGUgLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IHggZnJvbSAiLi9mb28uYmluIjsKaW1wb3J0IF9fdG9CaW5hcnkgZnJvbSAiLi90b0JpbmFyeS5qcyIKY29uc29sZS5sb2coX190b0JpbmFyeSgiYWFhYSIpKTsKY29uc29sZS5sb2coeCk7AABmb28uYmluAGFhYWEAAHRvQmluYXJ5LmpzAHZhciB0YWJsZSA9IG5ldyBVaW50OEFycmF5KDEyOCk7CmZvciAodmFyIGkgPSAwOyBpIDwgNjQ7IGkrKykgdGFibGVbaSA8IDI2ID8gaSArIDY1IDogaSA8IDUyID8gaSArIDcxIDogaSA8IDYyID8gaSAtIDQgOiBpICogNCAtIDIwNV0gPSBpOwoKZXhwb3J0IGRlZmF1bHQgKGJhc2U2NCkgPT4gewogICAgdmFyIG4gPSBiYXNlNjQubGVuZ3RoLCBieXRlcyA9IG5ldyBVaW50OEFycmF5KChuIC0gKGJhc2U2NFtuIC0gMV0gPT0gIj0iKSAtIChiYXNlNjRbbiAtIDJdID09ICI9IikpICogMyAvIDQgfCAwKTsKICAgIGZvciAodmFyIGkyID0gMCwgaiA9IDA7IGkyIDwgbjsgKSB7CiAgICAgICAgdmFyIGMwID0gdGFibGVbYmFzZTY0LmNoYXJDb2RlQXQoaTIrKyldLCBjMSA9IHRhYmxlW2Jhc2U2NC5jaGFyQ29kZUF0KGkyKyspXTsKICAgICAgICB2YXIgYzIgPSB0YWJsZVtiYXNlNjQuY2hhckNvZGVBdChpMisrKV0sIGMzID0gdGFibGVbYmFzZTY0LmNoYXJDb2RlQXQoaTIrKyldOwogICAgICAgIGJ5dGVzW2orK10gPSBjMCA8PCAyIHwgYzEgPj4gNDsKICAgICAgICBieXRlc1tqKytdID0gYzEgPDwgNCB8IGMyID4+IDI7CiAgICAgICAgYnl0ZXNbaisrXSA9IGMyIDw8IDYgfCBjMzsKICAgIH0KICAgIHJldHVybiBieXRlczsKfTs

evanw:
With this change, esbuild's `binary` loader will now use the new [`Uint8Array.fromBase64`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64) function unless it's unavailable in the configured target environment. If it's unavailable, esbuild's previous code for this will be used as a fallback.

Fixes #4286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
     console.log('size:', width + '\xD7' + height)
     ```
 
+* Use `Uint8Array.fromBase64` if available ([#4286](https://github.com/evanw/esbuild/issues/4286))
+
+    With this release, esbuild's `binary` loader will now use the new [`Uint8Array.fromBase64`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64) function unless it's unavailable in the configured target environment. If it's unavailable, esbuild's previous code for this will be used as a fallback.
+
 ## 0.25.10
 
 * Fix a panic in a minification edge case ([#4287](https://github.com/evanw/esbuild/issues/4287))
--- a/compat-table/src/index.ts
+++ b/compat-table/src/index.ts
@@ -53,6 +53,7 @@ export const jsFeatures = {
   ExportStarAs: true,
   ForAwait: true,
   ForOf: true,
+  FromBase64: true,
   FunctionNameConfigurable: true,
   FunctionOrClassPropertyAccess: true,
   Generator: true,
--- a/compat-table/src/mdn.ts
+++ b/compat-table/src/mdn.ts
@@ -18,6 +18,7 @@ const supportedEnvironments: Record<string, Engine> = {
 const jsFeatures: Partial<Record<JSFeature, string>> = {
   ClassStaticBlocks: 'javascript.classes.static.initialization_blocks',
   ExportStarAs: 'javascript.statements.export.namespace',
+  FromBase64: 'javascript.builtins.Uint8Array.fromBase64',
   ImportAssertions: 'javascript.statements.import.import_assertions',
   ImportAttributes: 'javascript.statements.import.import_attributes',
   ImportMeta: 'javascript.operators.import_meta',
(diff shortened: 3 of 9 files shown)


bakkot:
Might be worth also having the fallback implementation check for `fromBase64` at runtime. That way users who need support for older browsers can still get the native (and likely faster) implementation for their consumers who are on newer browsers.

PR #4296, opened 2025-10-06
improve keepNames behavior for profiling tools

sophiebits:
In my esbuild app with keepNames: true, minify: false I noticed that for declarations `let foo = function() {}` the function shows in V8 profiles as anonymous. This is due to the transformation to `let foo = __name(function() {}, 'foo')` which prevents the NamedEvaluation on the anonymous function which typically happens natively (eg: https://tc39.es/ecma262/#sec-variable-statement-runtime-semantics-evaluation for VariableDeclaration).

Evidently V8 keeps an internal idea of the function name distinct from the .name property. To work around this, we transform these instead to `let foo = __firstValue({ foo: function() {} })` which takes advantage of the name assignment in https://tc39.es/ecma262/#sec-runtime-semantics-propertydefinitionevaluation. With this fix, these functions show names correctly in the Chrome profiler.

### DIFF
--- a/internal/bundler_tests/snapshots/snapshots_default.txt
+++ b/internal/bundler_tests/snapshots/snapshots_default.txt
@@ -2749,102 +2749,154 @@ function foo(fn2 = function() {
 }) {
 }
 __name(foo, "foo");
-var fn = /* @__PURE__ */ __name(function() {
-}, "fn");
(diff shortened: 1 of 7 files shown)


sophiebits:
@evanw I would appreciate a review here if you have some time 🙏

PR #4307, opened 2025-10-21
refactor: remove obsolete // +build tag

vastonus:
From Go 1.17, the preferred syntax for build constraints is `//go:build`,
which replaces the old `// +build` form. The old style is now considered
deprecated but still supported for backward compatibility.

This change removes the obsolete `// +build xxx` line, keeping only the
modern `//go:build xxx` directive.

More info: https://github.com/golang/go/issues/41184 and https://go.dev/doc/go1.17#build-lines

Design Doc / Proposal：
https://go.dev/design/draft-gobuild

### DIFF
--- a/cmd/esbuild/main_wasm.go
+++ b/cmd/esbuild/main_wasm.go
@@ -1,5 +1,4 @@
 //go:build js && wasm
-// +build js,wasm
 
 package main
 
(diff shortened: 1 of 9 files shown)


evanw:
This breaks the build. [Support for Go 1.13 is deliberate](https://github.com/evanw/esbuild/blob/6b7c4f2dcbcaa1238bca0f7a4a1d95918296c82e/go.mod#L3-L5) so people can build esbuild themselves for old OS versions.

Ari4ka (approved):
****

evanw:
I'm closing this PR as it breaks the build.

PR #4313, opened 2025-10-30
feat: add CSS `@view-transition`

yisibl:
MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/@view-transition


yisibl:
It is my honor to be part of your project.

Issue #4311, opened 2025-10-30
Go Version 1.23.12 Contains Multiple CVEs – Request for Upgrade


ceesvanegmond:
Hi,

Starting today, we've identified several CVE vulnerabilities in the currently used Go version 1.23.12. Specifically, there are 10 CVEs, all originating from the `stdlib` library.
Would it be possible to upgrade to a newer Go version to address these issues?

Will post the CVE's in a seperate comment
Thanks!

ceesvanegmond:
```
========================================================================================================
Total: 10 (HIGH: 10, CRITICAL: 0)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬────────────────┬────────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version  │                           Title                            │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼────────────────┼────────────────────────────────────────────────────────────┤
│ stdlib  │ CVE-2025-47912 │ HIGH     │ fixed  │ v1.23.12          │ 1.24.8, 1.25.2 │ [net/url: insufficient validation of bracketed IPv6        │
│         │                │          │        │                   │                │ hostnames]                                                 │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2025-47912                 │
│         ├────────────────┤          │        │                   │                ├────────────────────────────────────────────────────────────┤
│         │ CVE-2025-58183 │          │        │                   │                │ [archive/tar: unbounded allocation when parsing GNU sparse │
│         │                │          │        │                   │                │ map]                                                       │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2025-58183                 │
│         ├────────────────┤          │        │                   │                ├────────────────────────────────────────────────────────────┤
│         │ CVE-2025-58185 │          │        │                   │                │ [encoding/asn1: pre-allocating memory when parsing DER     │
│         │                │          │        │                   │                │ payload can cause memory exhaustion]                       │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2025-58185                 │
│         ├────────────────┤          │        │                   │                ├────────────────────────────────────────────────────────────┤
│         │ CVE-2025-58186 │          │        │                   │                │ [net/http: lack of limit when parsing cookies can cause    │
│         │                │          │        │                   │                │ memory exhaustion]                                         │
[...]

Issue #4301, opened 2025-10-11
update go version


bhaktraj:
Fix: Security Update - Go 1.23 Vulnerability

Go version 1.23 has been identified to contain security vulnerabilities.

**Action Required:**
Please update your Go environment to version 1.24.4 immediately to patch these vulnerabilities and ensure system security.

This commit serves as a notification and reminder for all users and services relying on Go 1.23.

evanw:
This issue doesn't have enough information. I'm closing this issue as a duplicate of #4311 which describes the reasons for the upgrade.

PR #4317, opened 2025-10-31
update Go version to 1.24.9 for CVE fixes

KennethWhite:
This PR aims to update Go from 1.23.12 to 1.24.9 to address vulnerabilities, as noted in issues #4311 and #4301.

This is my first time contributing, feedback appreciated if I am missing something.


evanw:
Thanks for taking a stab at this. Wanted to give you a heads up that I'm currently considering taking a different approach to this which is more involved, but will hopefully make this type of issue less likely to come up in the future. So I may not take a PR like this.

KennethWhite:
Looking at the build error I am unsure how to resolve the issue. It appears the CI runner cannot find the specified version of Go, however, this should be cached/downloaded by the setup-go action. 

> Error: ENOENT: no such file or directory, open '/opt/hostedtoolcache/go/1.24.9/x64/misc/wasm/wasm_exec.js'

Its a bit tough to try to diagnose the issue further without the ability to run the pipelines.

As a side note: Some of the current CI actions are out of date, but not the cause of this issue and beyond the scope of what I wanted this PR to address.

evanw:
I'm closing this as esbuild has been [updated to Go 1.25](https://github.com/evanw/esbuild/pull/4318) as of [version 0.27.0](https://github.com/evanw/esbuild/releases/tag/v0.27.0).

PR #4319, opened 2025-11-03
enable trusted publishing

### LINKED ISSUE #4281: Consider Enabling Trusted Publishing
Trusted publishing is intended to be a safer alternative to API tokens for publishing packages. This is done by using short lived tokens between a CI provider and NPM.
NPM docs https://docs.npmjs.com/trusted-publishers

It looks like publishes happen for this repo from a local machine and not CI, so that would need to change to use this feature. The publish job looks like it also has more than just NPM publishes, so there's more complexity there too.

Is this something that should be enabled for this repo?

evanw:
GitHub and npm are recommending that maintainers for packages such as esbuild switch to [trusted publishing](https://docs.npmjs.com/trusted-publishers). With this change, a VM on GitHub will now build and publish all of esbuild's packages to npm instead of me. In theory.

Unfortunately there isn't really a way to test that this works other than to do it live. So the next release will be that live test. Hopefully it's uneventful and is exactly the same as the previous one (well, except for the green provenance attestation checkmark on npm that happens with trusted publishing).

The basic idea in this change is that every pushed commit that changes `version.txt` will trigger a new release attempt. This is explained in detail in the new `RUNBOOK.md` file. I did it that way because it's similar to how releases currently work and it seemed simple enough. 

Releasing esbuild involves coordinating between three separate repos. Previously the release process would commit and push to these repos as part of the release process. But doing that from a GitHub workflow doesn't seem straightforward, so I've changed these into a polling workflow on the other two repos. This seems ok for now because it should be fine that they don't update simultaneously with the release. They will update with at most a 6 hour delay.

Fixes #4281


hyrious:
Last week the Node.js 24 has become LTS with npm v11 included. So I guess `npm install -g npm@11.5.1` is not needed?

Here's a working case [in one of my packages](https://github.com/hyrious/lbq/blob/main/.github/workflows/npm-publish.yml). I'm using pnpm there, for npm it is much simpler (just `npm publish` will be enough).

evanw:
I also assumed that would be the case, but I wanted to be extra explicit here since releasing esbuild is important and I don't want it to break by accident. [The docs](https://docs.npmjs.com/trusted-publishers) say version 11.5.1 so I used that exact version since I don't think there's a reason not to. The idea is that this could guard against some unknown future npm regression that breaks publishing.

PR #4326, opened 2025-11-10
fix: deno release url wrong comment

ysknsid25:
If we want programmer  to refer Deno release note, Shouldn't we show [Deno LTS Release page](https://docs.deno.com/runtime/fundamentals/stability_and_releases/)?


evanw:
Sounds good! Thanks for pointing this out.

Issue #4329, opened 2025-11-14
Build crashing when metafile size > 512MB


elbywan:
👋 Kudos for this awesome build tool!

#### In a nutshell

When bundling very large projects with the `metafile: true` flag the build crashes with the following error.

```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at TextDecoder.decode (node:internal/encoding:447:16)
    at decodeUTF8 (/Users/xxx/dev/web-ui/.yarn/cache/esbuild-npm-0.25.10-f26f7be387-a8e4d33d7e.zip/node_modules/esbuild/lib/main.js:188:35)
    at visit (/Users/xxx/dev/web-ui/.yarn/cache/esbuild-npm-0.25.10-f26f7be387-a8e4d33d7e.zip/node_modules/esbuild/lib/main.js:99:16)
    at visit (/Users/xxx/dev/web-ui/.yarn/cache/esbuild-npm-0.25.10-f26f7be387-a8e4d33d7e.zip/node_modules/esbuild/lib/main.js:114:43)
    at decodePacket (/Users/xxx/dev/web-ui/.yarn/cache/esbuild-npm-0.25.10-f26f7be387-a8e4d33d7e.zip/node_modules/esbuild/lib/main.js:126:15)
    at handleIncomingPacket (/Users/xxx/dev/web-ui/.yarn/cache/esbuild-npm-0.25.10-f26f7be387-a8e4d33d7e.zip/node_modules/esbuild/lib/main.js:651:18)
    at Socket.readFromStdout (/Users/xxx/dev/web-ui/.yarn/cache/esbuild-npm-0.25.10-f26f7be387-a8e4d33d7e.zip/node_modules/esbuild/lib/main.js:581:7)
    at Socket.emit (node:events:524:28)
    at Socket.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:561:12) {
  code: 'ERR_STRING_TOO_LONG'
```

#### Context

We are currently using `esbuild` in our company to build the application as part of a huge monorepo (>10M LOC).
The `metafile` flag  is a requirement for us as we use specific build plugins to process the data.

The codebase is constantly growing and we faced the error recently.

From what I could gather this is caused by:
- The maximum size for a string in the v8 engine is 512MB ([defined here](https://chromium.googlesource.com/v8/v8.git/+/main/src/objects/string.h#519))
- The node process is storing the `metafile` value into a string before calling `JSON.parse()` - exceeding the threshold

[...]

PR #4349, opened 2025-12-02
compact metafile JSON blob

das7pad:
- Closes https://github.com/evanw/esbuild/issues/4329

This PR is compacting the metafile JSON blob contents to cut down the size and stay under the 512MiB string size limit in Node.js.

The tests have been extended to ensure that the metafile is properly compacted. The test snapshots use a pretty-printed blob for easier debugging.

See https://github.com/evanw/esbuild/issues/4329#issuecomment-3602112443 for alternatives considered.


das7pad:
Superseded by https://github.com/evanw/esbuild/commit/1ca56dc65155b0d887904c683cd43f7618ae621e

PR #4361, opened 2025-12-11
fix: allow subpath imports that start with `#/`

hybrist:
This aligns with the upcoming behavior in node that also started landing in other bundlers (Vite, webpack).

See: https://github.com/nodejs/node/pull/60864


evanw (approved):
Looks great. Thanks for the PR. Really appreciate the test cases!

Issue #3484, opened 2023-11-10
add option to local-css to prefix minified classnames


mpech:
link of original issue which was resolved: https://github.com/evanw/esbuild/issues/3295

Our problem is the following
- we have library A bundled by esbuild
- client B which depends on library A, also bundled **after** by esbuild
=> when B builds, B's stylesheet clashes with A's stylesheet, because we do not specify A as entrypoint, we just get the built files from A

What we would appreciate
- enable customizing classname like 'A[classnamehash]' (this would be the best, we could prefix our classes by libraryname)
- either specify A's stylesheet when building B to esbuild (it may be impractical, because if B is actually not a client but a library too, and if B uses A as peerdeps, while bundling B we will avoid clashes from A, but when building the client, maybe the classnames generated of A will have changed by then, which forces to have strict versionning)

mhsdesign:
Yes i agree. I currently have encountered the same problem. Im building a javascript plugin and it seems that another plugin already uses esbuild which causes css classes to overlap randomly. A simple prefix like `my-vendor` would be perfect.

lightningcss  also allows to declare a prefix via their "pattern" option: https://lightningcss.dev/css-modules.html#custom-naming-patterns

zouloux:
Hi, this is also an issue on some projects here.
I didn't find a way to intercept renaming through a plugin.
If anybody has a temporary solution, it would be helpful. In the meanwhile, we will disable the minify option.

mhsdesign:
You can use another plugin for css modules - like mine ^^ https://github.com/mhsdesign/esbuild-plugin-lightningcss-modules its based on lightningcss

markplindsay:
@evanw, my team develops a browser extension, and I recently switched us to esbuild for the dramatic speed improvement. Thank you very much for writing this software.

Our extension coexists in the DOM on many websites, and it is subject to these sites' global styling. I was troubleshooting a visual bug this week, and I noticed that Reddit uses a classname simply named `.a`. (Perhaps it is even generated by esbuild.)

Unfortunately, this conflicts with an `.a` classname generated by our esbuild during local-css's minification process,  and it messes up our extension's styling when it is rendered on Reddit.

We're currently working around this by not minifying our extension, but introducing an optional prefix for generated local-css classnames as configuration would solve this problem for us 100%.

kachurun:
[...]

PR #4391, opened 2026-02-10
feat: add deterministic local CSS naming option

joelmoss:
This feature introduces a new option `DeterministicLocalCSSNaming` that, when true, ensures that local CSS class names are generated in a deterministic manner. This is particularly useful for caching and consistent builds across different environments.

When `MinifyIdentifiers` is enabled, the local name is generated as the original CSS class name, and a hash of the absolute source file path, separated by an underscore (eg. `<original-class-name>_<hash-of-absolute-path>`).

When `MinifyIdentifiers` is disabled, the local name is appended with a CSS safe version of the relative source file path (eg.
`<original-class-name>_<hash-of-absolute-path>_<relative-source-file-path>`) to aid in development.

May address https://github.com/evanw/esbuild/issues/3484?


mhsdesign:
Hi thank you! using a hash would fix my problem from https://github.com/evanw/esbuild/issues/3484 indeed if the hash contains the package name of the npm package which it does due to the absolute path.

Now i have a question: As this change attempts to make bundles deterministic that two setups produce the same minified code, isnt it counterproductive to have the full absolute path being used for building the hash? Especially when people do not use a standardised docker container where their source code resides on the exact same path but work locally? Can we  instead use some kind of relative path + _if applicable_ the npm package name?

joelmoss:
The absolute path is used, because sometimes assets are located outside of the project. In which case, relative paths will not work.

mhsdesign:
paths could still be calculated via relative traversal `../../../some/outside.css`

joelmoss:
Yes, but then you may as well use absolute paths, and they are more readable.

PR #4397, opened 2026-02-19
remove the --trace flag

frances720:


frances720:
I'm so sorry!!! I don't know why the github default is always to create PR in the upstream directory. If I could delete these I would. Really sorry!

Issue #4401, opened 2026-02-20
Exception in downleveled code when concurrently polling an async generator that wraps another generator


ovcharik:
An exception is thrown in the generated downleveled code when an async generator is processed concurrently.

```ts
async function* inner() { yield 1; yield 2; }
async function* outer() { yield* inner(); }
const gen = outer();
Promise.all([gen.next(), gen.next()]).then(console.log).catch(console.error);
```

https://esbuild.github.io/try/#dAAwLjI3LjMALS10YXJnZXQ9ZXMyMDE1AGFzeW5jIGZ1bmN0aW9uKiBpbm5lcigpIHsgeWllbGQgMTsgeWllbGQgMjsgfQphc3luYyBmdW5jdGlvbiogb3V0ZXIoKSB7IHlpZWxkKiBpbm5lcigpOyB9CmNvbnN0IGdlbiA9IG91dGVyKCk7ClByb21pc2UuYWxsKFtnZW4ubmV4dCgpLCBnZW4ubmV4dCgpXSkudGhlbihjb25zb2xlLmxvZykuY2F0Y2goY29uc29sZS5lcnJvcik7

<img width="877" height="493" alt="Image" src="https://github.com/user-attachments/assets/909d9c52-bc88-495f-9003-08eb7faba89c" />

PR #4416, opened 2026-03-13
fix: Handle non-awaited async yield*

2767mr:
This PR updates the `__asyncGenerator` and `__yieldStar` helpers to allow `.next()` calls that are not awaited.

Previously, the `method` in `__yieldStar` incorrectly assumed that, once invoked, it would not be called again until the `resume` callback in `__asyncGenerator` had resolved its promise.

The helper class `__yieldStarResult` is now used to mark the incoming value as already resolved.

For #4401.


2767mr:
Closing in favor of #4417

Issue #4430, opened 2026-03-30
Support Custom Media Queries


pereorga:
When designing documents that use media queries, the same media query may be used in multiple places, such as to qualify multiple [@import](https://www.w3.org/TR/css-cascade-5/#at-ruledef-import) statements. Repeating the same media query multiple times is an editing hazard; an author making a change must edit every copy in the same way, or suffer from difficult-to-find bugs in their CSS.

(_From https://drafts.csswg.org/mediaqueries-5/#custom-mq_)

This is currently supported by Lightning CSS.

joeda1:
I agree that would be nice, but its still a draft not not yet supported by browsers: https://caniuse.com/?search=custom-media
I'm using a postcss plugin since many years, it works well with esbuild: https://www.npmjs.com/package/postcss-custom-media

PR #4431, opened 2026-03-30
feat(css): lower @custom-media rules

pereorga:
Closes https://github.com/evanw/esbuild/issues/4430

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.27.5
+## Unreleased
 
 * Lower CSS `@custom-media` rules ([#4430](https://github.com/evanw/esbuild/issues/4430))
 


evanw:
Thanks for the PR, and thank you very much for adding tests.

However, I'm guessing this should also work across files when bundling is enabled. Is that correct? CSS `@import` sort of behaves like a C++ `#include` where the source is evaluated inline, so I think that would select the last `<extension-name>` in the expanded import order as the ultimate substitution value for the custom media query.

I'm also not sure whether it's too early to add something like this to esbuild or not. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@custom-media) has this feature listed as experimental, and it only has one hidden prototype implementation so far (in Firefox).

pereorga:
Thanks for taking the time to look at this.

You are right. I can't believe I missed that, cross-file is the important use case.

Regarding adding support for it in esbuild, I see why it may be too early, [although there is some usage and demand for it](https://github.com/search?q=%22%40custom-media%22&type=code). I guess some developers like to use it, even if it is risky.

Anyway, since the current approach won't work for multiple files, I'll close this PR.

PR #4467, opened 2026-05-19
Make __esm helper preserve init errors across calls

### LINKED ISSUE #4461: `__esm` helper silently swallows init errors after the first throw
### Version

esbuild `0.25.4`. The pattern below has been present in the helper for a long time and reproduces on every recent version I checked.

### Summary

The `__esm` helper used to lazy-initialize ESM-namespace modules surfaces a thrown init error **only on the very first call**. All subsequent calls return `undefined` silently, so downstream code throws a misleading `TypeError: Cannot read properties of undefined (reading '<some-export>')` and the real init error is gone.

Root cause: `fn = 0` is sequenced **before** the call to the underlying init function. If the init throws, `fn` has already been cleared, so the next time the helper runs, the `fn && ...` short-circuit skips the init entirely and returns the still-unset `res = undefined`.

### Helper as emitted

```js
var __esm = (fn, res) => function __init() {
  return fn && (res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0)), res;
};
```

The order of operations on the call line:

1. `__getOwnPropNames(fn)[0]` — read the first property name of `fn`
2. `fn[...]` — fetch the init function
3. `(0, ...)` — strip `this`
4. `(fn = 0)` — assignment passed as the argument; **`fn` is now `0` before the call returns**
5. `(...)` — call the init function with `0` as the arg
6. If init throws, control jumps out **with `fn` already 0**

After step 6, `res` is still `undefined`. Every later call returns `undefined`.

### Minimal repro

```js
// boom.js
console.log("evaluating boom.js");
throw new Error("init failed (the real cause)");
export const value = 42;

// consumer.js
import { value } from "./boom.js";
export function getValue() { return value; }

// entry.js
import { getValue } from "./consumer.js";
export default {
  fetch() {
    try { return new Response(String(getValue())); }
    catch (e) { return new Response("err: " + e.message); }
  },
};
```

Bundle:

```sh
esbuild entry.js --bundle --format=esm --target=es2022 --outfile=out.js
```

Run twice on a runtime that keeps the module cache (Workers, Deno Deploy, Lambda warm starts, a long-lived Node server). First call: real `Error: init failed (the real cause)`. Second call: `Cannot read properties of undefined (reading 'value')` — the original error has been swallowed.

### Why this hurts in practice

Amplified hugely on long-lived single-isolate runtimes. One isolate handles many requests, so a cold-start init failure produces one log line with the real error, then N log lines with the symptom error. Short log retention or reactive (rather than always-on) tailing means the real cause is unrecoverable without a redeploy.

OpenNext / `@opennextjs/cloudflare` users are particularly exposed because their middleware bundle wraps Next.js's `_ENTRIES` registry through `__esm`; an init error blanks the registry and every later request gets `Cannot read properties of undefined (reading 'default')`.

### Proposed fix

Capture the thrown error in a closure and re-throw on every subsequent call:

```js
var __esm = (fn, res, err) => function __init() {
  if (err) throw err;
  if (fn) {
    try {
      res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0);
    } catch (e) {
      err = e;
      throw e;
    }
  }
  return res;
};
```

One extra closed-over variable per wrapped module. Behavior change: a failed init becomes permanently sticky on the same module instance — which is what users almost always want (re-running an init that threw is rarely safe; side effects may have run).

### Workaround

A post-build string-replace patch on the bundled output. Reference: https://github.com/teamvirtus/virtus-web/blob/main/scripts/patch-opennext-bundle.js

Happy to send a PR if you'd like.

LeSingh1:
Closes #4461.

The `__esm` helper sequenced `fn = 0` before the call to the underlying init function:

```js
var __esm = (fn, res) => function __init() {
  return fn && (res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0)), res
}
```

If init threw, `fn` was already 0 by then. Subsequent calls hit the `fn && ...` short-circuit and returned the still-unset `res` (`undefined`). The real init error was visible on the first call only; every later call returned `undefined` silently, and downstream code threw a misleading `TypeError: Cannot read properties of undefined (reading '<some-export>')`.

Reproducer (vanilla `esbuild@0.25.x`):

```js
// boom.js
console.log("evaluating boom.js");
throw new Error("init failed (the real cause)");
export const value = 42;

// entry.js
let logs = [];
async function callOnce(label) {
  try { logs.push(label + ": value=" + (await import("./boom.js")).value); }
  catch (e) { logs.push(label + ": error=" + e.message); }
}
(async () => { await callOnce("call1"); await callOnce("call2"); await callOnce("call3"); console.log(logs.join("\n")); })();
```

```
$ esbuild entry.js --bundle --format=cjs --target=es2022 --outfile=out.js
$ node out.js
evaluating boom.js
call1: error=init failed (the real cause)
call2: value=undefined
call3: value=undefined
```

After this change `call2` and `call3` re-throw the same `Error: init failed (the real cause)`.

This matters on long-lived single-isolate runtimes (Workers, Deno Deploy, Lambda warm starts, a Node server). One isolate handles many requests; a cold-start init failure produces one log line with the real cause, then N log lines with a `Cannot read properties of undefined` downstream symptom. Short log retention or reactive (rather than always-on) tailing means the real cause is unrecoverable without a redeploy. OpenNext / `@opennextjs/cloudflare` users are particularly exposed because their middleware bundle wraps the `_ENTRIES` registry through `__esm`.

Fix: capture the thrown error in a closure variable and re-throw on every later call.

```js
var __esm = (fn, res, err) => function __init() {
  if (err) throw err
  if (fn) try { res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0) } catch (e) { err = e; throw e }
  return res
}
```

Behavior change: a failed init is now permanently sticky on the same module instance. Re-running an init that threw is rarely safe because side effects (top-level statements that ran before the throw) may have already executed, so re-throwing is the spec-correct surface — matches what the V8/SpiderMonkey-emitted ESM `state === errored` does for native modules.

Same change applied to `__esmMin` for parity. `__esmMin` stays compact (one `if` per line, no extra braces beyond what the catch needs).

Verified:
- The vanilla 0.25.x output reproduces the bug; the patched output throws the original error on every call.
- `go test ./internal/runtime/... ./internal/bundler_tests/...` — both packages clean.
- No bundler snapshots change (helpers aren't snapshotted in this repo's `internal/bundler_tests/snapshots/`).

Not changed in this PR (separate, related shape): `__commonJS` assigns `mod = {exports: {}}` *before* the underlying `cb` runs, so a thrown CJS init similarly produces an empty `mod.exports` on later calls. Happy to follow up if you'd like.

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -170,10 +170,23 @@ func Source(unsupportedJSFeatures compat.JSFeature) logger.Source {
 		// This is for lazily-initialized ESM code. This has two implementations, a
 		// compact one for minified code and a verbose one that generates friendly
 		// names in V8's profiler and in stack traces.
-		export var __esm = (fn, res) => function __init() {
-			return fn && (res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0)), res
+		//
+		// A thrown init error is captured and re-thrown on every later call, so a
+		// failed cold-start surfaces the real cause on subsequent requests instead
+		// of being silently replaced with "Cannot read properties of undefined" on
+		// long-lived runtimes (Workers, Deno Deploy, Lambda warm starts, a Node
+		// server). Re-running an init that threw is rarely safe; side effects may
+		// have run.
+		export var __esm = (fn, res, err) => function __init() {
+			if (err) throw err
+			if (fn) try { res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0) } catch (e) { err = e; throw e }
+			return res
 		}
-		export var __esmMin = (fn, res) => () => (fn && (res = fn(fn = 0)), res)
+		export var __esmMin = (fn, res, err) => () => {
+			if (err) throw err
+			if (fn) try { res = fn(fn = 0) } catch (e) { err = e; throw e }
+			return res
+		}
 
 		// Wraps a CommonJS closure and returns a require() function. This has two
 		// implementations, a compact one for minified code and a verbose one that


Big621:
Thanks

PR #4488, opened 2026-06-30
css: fix green/blue channel swap in oklch gamut mapping

chatman-media:
`gamut_mapping_xyz_to_srgb`'s binary search calls `srgb_to_oklab(clipped_r, clipped_b, clipped_g)` when measuring how far the clipped color has drifted from the original, swapping the green and blue arguments. The very next line computes the comparison point with the correct `r, g, b` order, so the JND check ends up comparing against the wrong OKLab point and can settle on a less accurate clipped chroma.

Most existing test colors didn't catch this because they either stay in-gamut or happen to have green ≈ blue, making the swap a no-op. With channels far enough apart it's visible: `.a{color:oklch(0.65 0.3 50)}` targeting chrome90 lowers to `#de6800` instead of the correct `#e56300`.

Added two oklch cases with separated g/b to `TestColorFunctions` and fixed six stale hex values in `TestGradient` that were asserting the buggy output.

### DIFF
--- a/internal/css_parser/css_color_spaces.go
+++ b/internal/css_parser/css_color_spaces.go
@@ -437,7 +437,7 @@ func gamut_mapping_xyz_to_srgb(x F64, y F64, z F64) (F64, F64, F64) {
 		}
 
 		clipped_r, clipped_g, clipped_b := clip(r), clip(g), clip(b)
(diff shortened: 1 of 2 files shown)


evanw:
Thanks for fixing this!

PR #4493, opened 2026-07-06
fix #4484: actually skip writing the file when refusing to overwrite an input file

### LINKED ISSUE #4484: Bug: esbuild overwrites the input file despite throwing "Refusing to overwrite input file"
### Description

When `outdir` points to the directory containing the entry point, `esbuild` throws an error:

`Build failed with 1 error: error:`
`Refusing to overwrite input file error`.

However, it still destructively overwrites the original input file with the minified version before/during throwing the error.

### Reproduction

**Environment**

    OS: Windows
    esbuild version: 0.28.1

Create the following files and run `node esbuild.js`.

**sources/test.js**

    function test() { console.log( 1 + 1 === 2 ); }

**esbuild.js**

    import fs from 'fs';
    import path from 'path';
    import esbuild from 'esbuild';

    (async function run() {
      const basePath = process.cwd();
      const sourcePath = path.resolve(basePath, "sources");
      const sourceFile = path.resolve(basePath, "sources/test.js");

      try {
        console.log("Initial source size:", fs.statSync(sourceFile).size, 'bytes');

        await esbuild.build({
          logLevel: 'silent',
          entryPoints: [ sourceFile ],
          outdir: sourcePath,
          bundle: false,
          minify: true,
          write: true,
        });

      } catch (err) {

        console.log(err.message);
        console.log("Source size after failure:", fs.statSync(sourceFile).size, 'bytes');

      }
    })();

### Actual Behavior

The build fails as expected, but the input file is modified anyway:

    Initial source size: 47 bytes
    Build failed with 1 error:
    error: Refusing to overwrite input file "sources/test.js" (use "allowOverwrite: true" to allow this)
    Source size after failure: 33 bytes

The content of `sources/test.js` becomes minified:

    function test(){console.log(!0)}

### Expected Behavior

`esbuild` should abort the write operation and leave the original input file completely untouched if `allowOverwrite` is not set to true.

Zelys-DFKH:
Fixes #4484

When the output path resolves to an input file, esbuild logs "Refusing to overwrite input file" but writes the file anyway, so the input still gets clobbered. The error is raised inside Compile, and the write phase only checks for errors from before compilation, so the conflicting entry goes through.

This drops the conflicting entry from outputFiles at the point where the error is logged, using the same in-place compaction as the duplicate-output check below it. Companion outputs derived from the same path (.map and .LEGAL.txt, the only two appended by the linker) are dropped with it so they aren't written on their own. The suffixes are compared lowercase because canonicalFileSystemPathForWindows lowercases the keys on both sides.

Thanks to @noskill24 for the report and @cschanhniem for tracing the root cause.

The new test covers the sourcemap and legal comments cases and fails without the fix.

### DIFF
--- a/internal/bundler/bundler.go
+++ b/internal/bundler/bundler.go
@@ -3103,6 +3103,7 @@ func (b *Bundle) Compile(log logger.Log, timer *helpers.Timer, mangleCache map[s
 					sourceAbsPaths[absPathKey] = sourceIndex
 				}
 			}
+			rejectedAbsPaths := make(map[string]bool)
 			for _, outputFile := range outputFiles {
 				absPathKey := canonicalFileSystemPathForWindows(outputFile.AbsPath)
 				if sourceIndex, ok := sourceAbsPaths[absPathKey]; ok {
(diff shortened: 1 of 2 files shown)


evanw:
Thanks for the PR, but I'm going to try fixing this a different way. I believe the intended behavior is to not write out any output files when there's a build error, and these refusal log messages are build errors. So really there shouldn't be any output files at all.

Issue #4503, opened 2026-08-04
app crashes in our CI env


ZuBB:
node 24.18.0 (not sure if this is important)
esbuild @0.28.1
env: default runner in GHE, seems a default 64bit linux.

```log
2026-08-04T08:40:22.6099516Z fatal error: all goroutines are asleep - deadlock!
2026-08-04T08:40:22.6119243Z 
2026-08-04T08:40:22.6119445Z goroutine 1 [chan receive]:
2026-08-04T08:40:22.6120276Z github.com/evanw/esbuild/internal/helpers.(*ThreadSafeWaitGroup).Wait(...)
2026-08-04T08:40:22.6121862Z 	github.com/evanw/esbuild/internal/helpers/waitgroup.go:36
2026-08-04T08:40:22.6122446Z main.runService.func2()
2026-08-04T08:40:22.6122953Z 	github.com/evanw/esbuild/cmd/esbuild/service.go:114 +0x53
2026-08-04T08:40:22.6123481Z main.runService(0x1)
2026-08-04T08:40:22.6123962Z 	github.com/evanw/esbuild/cmd/esbuild/service.go:160 +0x637
2026-08-04T08:40:22.6124452Z main.main()
2026-08-04T08:40:22.6124896Z 	github.com/evanw/esbuild/cmd/esbuild/main.go:252 +0xb16
2026-08-04T08:40:22.6125254Z 
2026-08-04T08:40:22.6125818Z goroutine 7 [chan receive]:
2026-08-04T08:40:22.6126218Z main.runService.func1()
2026-08-04T08:40:22.6126720Z 	github.com/evanw/esbuild/cmd/esbuild/service.go:98 +0x45
2026-08-04T08:40:22.6127280Z created by main.runService in goroutine 1
2026-08-04T08:40:22.6127824Z 	github.com/evanw/esbuild/cmd/esbuild/service.go:97 +0x1ce
2026-08-04T08:40:22.6128183Z 
2026-08-04T08:40:22.6128470Z goroutine 8 [chan receive]:
2026-08-04T08:40:22.6129098Z main.(*serviceType).sendRequest(0x1d93f1a4b050, {0xa10be0, 0x1d93f4c025d0})
2026-08-04T08:40:22.6130414Z 	github.com/evanw/esbuild/cmd/esbuild/service.go:192 +0x12b
2026-08-04T08:40:22.6130893Z main.runService.func3()
[...]

PR #4506, opened 2026-08-04
fix service deadlock after failed build setup

ZuBB:
disclaimer: whole thing has been researched and coded with help of AI tools

### DIFF
--- a/cmd/esbuild/service.go
+++ b/cmd/esbuild/service.go
@@ -639,6 +639,12 @@ func (service *serviceType) handleBuildRequest(id uint32, request map[string]int
 	}
 
 	activeBuild := service.createActiveBuild(key)
+	shouldDestroyActiveBuild := true
+	defer func() {
+		if shouldDestroyActiveBuild {
+			service.destroyActiveBuild(key)
+		}
+	}()
 
 	hasOnEndCallbacks := false
 	if plugins, ok := request["plugins"]; ok {
@@ -769,6 +775,7 @@ func (service *serviceType) handleBuildRequest(id uint32, request map[string]int
 		// Keep the build alive until "dispose" has been called
 		activeBuild.disposeWaitGroup.Add(1)
 		activeBuild.ctx = ctx
+		shouldDestroyActiveBuild = false
 
 		return encodePacket(packet{
 			id: id,
@@ -781,8 +788,6 @@ func (service *serviceType) handleBuildRequest(id uint32, request map[string]int
 
 	result := api.Build(options)
 	response := resultToResponse(result)
-
-	service.destroyActiveBuild(key)
 
 	return encodePacket(packet{
 		id:    id,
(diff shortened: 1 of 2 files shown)


ZuBB:
this PR addressed what is reported in https://github.com/evanw/esbuild/issues/4503

I had to go this way as it started to appear in local setup as well

evanw:
Thanks very much for sending this. This is very helpful.

(for myself) It appears to be a real issue. I can't reproduce it myself in an unmodified end-to-end test because the Go child process does not normally keep the Node parent process alive. But I was able to reproduce it by hacking the JavaScript shim like this:

```patch
diff --git a/lib/npm/node.ts b/lib/npm/node.ts
index 1a6cb72d..5c5e31d3 100644
--- a/lib/npm/node.ts
+++ b/lib/npm/node.ts
@@ -300,15 +300,15 @@ let ensureServiceIsRunning = (): Service => {
   let refCount = 0
   child.unref()
   if (stdin.unref) {
-    stdin.unref()
+    // stdin.unref()
   }
   if (stdout.unref) {
-    stdout.unref()
+    // stdout.unref()
   }
 
   const refs: common.Refs = {
     ref() { if (++refCount === 1) child.ref(); },
-    unref() { if (--refCount === 0) child.unref(); },
+    unref() { if (--refCount === 0) child.unref(), stdin.end(); },
   }
 
   longLivedService = {
```

and then running the API like this:

```js
require('esbuild').context({
  mangleProps: /x(?=y)/,
}).then(
  success => console.log({ success }),
  failure => console.log({ failure }),
)
```

That prints the following:

```
✘ [ERROR] The "mangle props" setting is not a valid Go regular expression: x(?=y)

fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
github.com/evanw/esbuild/internal/helpers.(*ThreadSafeWaitGroup).Wait(...)
	github.com/evanw/esbuild/internal/helpers/waitgroup.go:36
main.runService.func2()
	github.com/evanw/esbuild/cmd/esbuild/service.go:114 +0x8c
main.runService(0x1)
	github.com/evanw/esbuild/cmd/esbuild/service.go:160 +0x560
main.main()
	github.com/evanw/esbuild/cmd/esbuild/main.go:252 +0xb5c

goroutine 7 [chan receive]:
main.runService.func1()
	github.com/evanw/esbuild/cmd/esbuild/service.go:98 +0x40
created by main.runService in goroutine 1
	github.com/evanw/esbuild/cmd/esbuild/service.go:97 +0x188

goroutine 8 [chan receive]:
main.(*serviceType).sendRequest(0x7fccbfd3d020, {0x103348fc0, 0x7fccbfd3d110})
	github.com/evanw/esbuild/cmd/esbuild/service.go:192 +0x134
main.runService.func3()
	github.com/evanw/esbuild/cmd/esbuild/service.go:125 +0x44
created by main.runService in goroutine 1
	github.com/evanw/esbuild/cmd/esbuild/service.go:122 +0x2a8
```

The fix proposed here does fix this end-to-end deadlock.

PR #4510, opened 2026-08-07
fix #4508: preserve lowered assignment bindings

### LINKED ISSUE #4508: minify + target≤es2020: lowered `||=` on a write-only local emits assignment to an undeclared variable (runtime ReferenceError)
## Summary

With `--minify` and a target that requires lowering logical assignment operators (`es2020` or below), esbuild mis-compiles a **write-only** local variable used with `||=`: the `let` declaration is removed as dead code and the read side is folded to `void 0`, but the lowered assignment survives — renamed to a fresh identifier that is **never declared**. In ESM strict mode this throws `ReferenceError` at runtime.

Reproduces on `0.25.12` and on `0.28.1` (latest at time of writing). Does **not** reproduce at `target=es2021`+ (no `||=` lowering) or without minification.

## Minimal reproduction

`in.mjs`:
```js
export class H {
  requestMode(e, i) {
    let r;
    (P => (P[P.NOT_RECOGNIZED = 0] = "NOT_RECOGNIZED", P[P.SET = 1] = "SET"))(r ||= {});
    return i ? this.a() : this.b();
  }
  a() { return 1; }
  b() { return 2; }
}
```

```
$ esbuild in.mjs --minify --format=esm --target=es2020
```

## Actual output (broken)

```js
class u{requestMode(E,e){return(r=>(r[r.NOT_RECOGNIZED=0]="NOT_RECOGNIZED",r[r.SET=1]="SET"))(void 0||(s={})),e?this.a():this.b()}a(){return 1}b(){return 2}}export{u as H};
```

`void 0 || (s = {})` — `s` is assigned but has no declaration anywhere. Running it:

```
$ node --input-type=module -e "import { H } from './broken.mjs'; new H().requestMode(12, false);"
ReferenceError: s is not defined
```

## Expected

One of:
- keep the binding: `let t; ... (t || (t = {}))` (what esbuild correctly emits when the variable is read elsewhere), or
- drop the entire dead construct: `(P => ...)({})`.

For comparison, `--target=es2021` (no `||=` lowering) is correct:

```js
class u{requestMode(s,r){let t;return(e=>(e[e.NOT_RECOGNIZED=0]="NOT_RECOGNIZED",e[e.SET=1]="SET"))(t||={}),r?this.a():this.b()}...}
```

## Root-cause hypothesis

At `target≤es2020` the lowering pass rewrites `r ||= {}` to `r || (r = {})`. Then the minify DCE pass sees `let r` where `r` is never *read* (write-only): it folds the read `r ||` to `void 0 ||` and deletes the `let r` declaration, but keeps the assignment `r = {}` (a side effect) — and symbol renaming then mints a fresh identifier (`s`) for a binding that no longer exists. The two passes disagree about whether the variable is dead.

The trigger requires all three:
1. `target ≤ es2020` (so `||=` is lowered into read + write),
2. `--minify` (DCE + rename),
3. the variable is **write-only** after the `||=` (any later read of `r` keeps the declaration and the bug disappears — verified by adding `return r` / `r.SET` reads).

## Real-world impact

This is the root cause of https://github.com/xtermjs/xterm.js/issues/5800. `@xterm/xterm@6.0.0`'s published `lib/xterm.mjs` (itself esbuild-minified) contains exactly this pattern in `InputHandler.requestMode` — a `const enum` compiled to an IIFE whose enum object is write-only:

```js
requestMode(e,i){let r;(P=>(P[P.NOT_RECOGNIZED=0]="NOT_RECOGNIZED",...))(r||={});let n=...}
```

Every Vite production build (default `build.target: 'modules'` → es2020; default esbuild minify) that bundles `@xterm/xterm@6.0.0` emits the broken `void 0 || (i = {})` form. The first DECRQM query (`CSI ? Ps $ p`, sent on startup by vim, nvim, htop, less, opencode, …) throws inside the parser, which permanently stalls xterm's write queue — rendering freezes while keyboard input keeps flowing to the PTY, so the app looks "half alive". Several downstream projects hit this independently (Pane, Wails apps, our Tauri 2 app on Windows/WebView2, Vite 5 & 6).

I verified the minimal Vite repro: a fresh `vite@6.0.3` project with only `@xterm/xterm@6.0.0` produces a `dist` bundle containing `requestMode(e,t){(g=>...)(void 0||(i={}))}` — byte-identical failure shape to the isolated CLI repro above. Rollup alone does not corrupt the pattern; esbuild alone does.

## Environment

- esbuild 0.25.12 (broken), 0.28.1 (broken)
- Windows 11 x64, Node 22 (platform-independent; also reported on macOS/Linux downstream)

wanxiankai:
Fixes #4508.

Lowering an assignment operator duplicates an identifier into separate read and write nodes, but the generated second reference wasn't included in the symbol use count. When single-use inlining later simplified the read side of a lowered `||=`, it could also remove the declaration while leaving the write behind.

This change records the generated identifier use and adds a regression test for the enum-IIFE pattern from the issue.

Tests:
- `go test ./internal/... ./pkg/...`
- `go vet ./cmd/... ./internal/... ./pkg/...`
- `node scripts/plugin-tests.js`
- `node scripts/js-api-tests.js`
- `node scripts/register-test.js`
- `node scripts/node-unref-tests.js`

### DIFF
--- a/internal/js_parser/js_parser_lower.go
+++ b/internal/js_parser/js_parser_lower.go
@@ -849,6 +849,9 @@ func (p *parser) lowerAssignmentOperator(value js_ast.Expr, callback func(js_ast
 		}
 
 	case *js_ast.EIdentifier:
+		// The assignment target was already recorded as one usage when it was
+		// visited. Lowering it generates a second usage, so record that too.
+		p.recordUsage(left.Ref)
 		return callback(
 			js_ast.Expr{Loc: value.Loc, Data: &js_ast.EIdentifier{Ref: left.Ref}},
 			value,
--- a/internal/js_parser/js_parser_lower_test.go
+++ b/internal/js_parser/js_parser_lower_test.go
@@ -214,6 +214,12 @@ class Foo {
 _x = new WeakMap();
 `)
 
+	// Lowering generates an additional symbol use that must keep the declaration
+	expectPrintedMangleTarget(t, 2020, `let r;
+(P => (P[P.NOT_RECOGNIZED = 0] = "NOT_RECOGNIZED", P[P.SET = 1] = "SET"))(r ||= {});`, `let r;
+((P) => (P[P.NOT_RECOGNIZED = 0] = "NOT_RECOGNIZED", P[P.SET = 1] = "SET"))(r || (r = {}));
+`)
+
 	expectPrintedTarget(t, 2021, "a ||= b", "a ||= b;\n")
 	expectPrintedTarget(t, 2021, "a.b ||= c", "a.b ||= c;\n")
 	expectPrintedTarget(t, 2021, "a().b ||= c", "a().b ||= c;\n")


internal/js_parser/js_parser_lower.go
```diff
@@ -849,6 +849,9 @@ func (p *parser) lowerAssignmentOperator(value js_ast.Expr, callback func(js_ast
 		}
 
 	case *js_ast.EIdentifier:
+		// The assignment target was already recorded as one usage when it was
+		// visited. Lowering it generates a second usage, so record that too.
+		p.recordUsage(left.Ref)
```
evanw, line 854: (for myself) This seems like a reasonable fix. Hypothetically compound assignment operators are already a read and a write, so an alternative fix could be to just always give compound assignment operators a use count of two. But that seems less consistent with how the rest of the code handles the use count than this approach.

internal/js_parser/js_parser_lower_test.go
```diff
@@ -212,6 +212,12 @@ class Foo {
   }
 }
 _x = new WeakMap();
+`)
+
+	// Lowering generates an additional symbol use that must keep the declaration
+	expectPrintedMangleTarget(t, 2020, `let r;
+(P => (P[P.NOT_RECOGNIZED = 0] = "NOT_RECOGNIZED", P[P.SET = 1] = "SET"))(r ||= {});`, `let r;
+((P) => (P[P.NOT_RECOGNIZED = 0] = "NOT_RECOGNIZED", P[P.SET = 1] = "SET"))(r || (r = {}));
```
evanw, line 220: This test doesn't make sense since it already passes on main without the fix (i.e. it wouldn't have caught this regression).

evanw (commented):
Thanks for the PR, but I'm going to fix this myself instead given that it's a one-line change and needs different tests.

PR #4511, opened 2026-08-08
escape `<!--` and `-->` for inline-script safety

### LINKED ISSUE #4494: `inline-script` should escape HTML comment brackets in string/regex/template literals
When the `inline-script` syntax feature is enabled (the default), ESBuild attempts to escape string/template literals that contain `</script>`, so that the output can be safely included in an inline script tag.

However, there's another case that can cause issues for inline script tags: if an inline script contains the character sequences `<!--` and `<script>` in that order, then the HTML parser will treat the closing `</script>` tag as part of the script (basically ignoring it). (There can be ~arbitrary text in between the `<!--` and the `<script>`.) This is called the [script data double escaped state](https://html.spec.whatwg.org/multipage/parsing.html#script-data-double-escaped-state), and you can observe it by viewing an HTML file with the following contents:

```html
<script>
  console.log("<!-- <script>");
</script>
This text is not displayed, because it's still considered inside the script tag.
```

(Note that GitHub's syntax highlighting does not correctly handle this case.)

ESBuild does not currently escape either `<!--` or `<script>`: https://esbuild.github.io/try/#dAAwLjI4LjEAAGNvbnNvbGUubG9nKCI8L3NjcmlwdD4gPCEtLSA8c2NyaXB0PiIp

For comparison, [Terser](https://try.terser.org/) avoids this case by escaping the angle brackets in string literals that look like the start/end of HTML comments:

```js
console.log("<!--", "<script>", "-->");
// Becomes:
console.log("\x3c!--","<script>","--\x3e");
```

(These sequences can also appear in regular expression literals, but it doesn't look like Terser does anything about that.)

MsfPablo:
Fixes #4494.

When the `inline-script` feature is enabled (the default), esbuild escapes `</script` in string/template literals so output can be embedded in an inline `<script>` tag. The HTML parser's script data state machine also treats `<!--` and `-->` specially:

- `<!--` enters the ["script data escaped"](https://html.spec.whatwg.org/multipage/parsing.html#script-data-escaped-state) state
- a later `<script` in that state enters the ["script data double escaped"](https://html.spec.whatwg.org/multipage/parsing.html#script-data-double-escaped-state) state, where `</script>` no longer ends the element
- `-->` exits the escaped state

So a string like `"<!-- <script>"` defeats the existing `</script` escaping. This PR escapes `<!--` as `\x3c!--` and `-->` as `--\x3e` in string/template content, mirroring [Terser's `inline_script` option](https://github.com/terser/terser), which the issue references:

```js
console.log("<!--", "<script>", "-->");
// Becomes:
console.log("\x3c!--","<script>","--\x3e");
```

`<script>` (without a slash) is left alone, matching Terser: it is only dangerous after a `<!--`, which is now escaped, and in the normal script-data state `<script` does nothing special.

Tagged template literals such as `String.raw\`...\`` emit raw content verbatim, so the printer cannot escape them in place. The existing `</script` lowering (which turns such templates into `__template` calls) is extended to also cover `<!--` and `-->`, letting the same escapes apply to the raw content while preserving the runtime value (the `\x3c`/`\x3e` escapes are source-level only; the string value stays `<!--`/`-->`, same as the existing `</script` handling).

Both the printer escaping and the tagged-template lowering are gated on `compat.InlineScript` being supported, consistent with the existing `</script` behavior.

```js
// in
console.log("<!-- <script>", "-->");
// out
console.log("\x3c!-- <script>", "--\x3e");
```

```js
// in
x = String.raw`<!--`;
// out
import { __template } from "<runtime>";
var _a;
x = String.raw(_a || (_a = __template(["\x3c!--"])));
```

`go build ./...` and `go test ./...` pass; `gofmt` and `go vet` clean on the touched files.

### DIFF
--- a/internal/js_parser/js_parser.go
+++ b/internal/js_parser/js_parser.go
@@ -13405,6 +13405,21 @@ func containsClosingScriptTag(text string) bool {
 	return false
 }
 
+// containsInlineScriptHazard reports whether text contains a character
+// sequence that the HTML parser's script data state machine treats specially
+// inside a <script> element: "</script" (closes the element), "<!--" (enters
+// the "script data escaped" state), or "-->" (exits an escaped state). When
+// such a sequence appears in a tagged template literal, the printer cannot
+// escape it because the raw template content is emitted verbatim, so the
+// template must be lowered into a regular call first (mirroring Terser's
+// inline_script option, which escapes these sequences in string output).
+func containsInlineScriptHazard(text string) bool {
+	if containsClosingScriptTag(text) {
+		return true
+	}
+	return strings.Contains(text, "<!--") || strings.Contains(text, "-->")
+}
+
 func (p *parser) isUnsupportedRegularExpression(loc logger.Loc, value string) (pattern string, flags string, isUnsupported bool) {
 	var what string
 	var r logger.Range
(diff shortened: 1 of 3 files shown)


MsfPablo:
Following up here. I realise escaping behaviour is a deliberate design area in esbuild, so if this isn't a change you want I'm happy to close it — mainly looking for a yes/no.

